PR #2391, opened 2020-06-15
hack/run-in-docker.sh: use node 12 which is now required

kdomanski:
```release-note
NONE
```


kgroschoff:
@kdomanski could you rebase your pr, please? The e2e tests should work afterwards :)

PR #2410, opened 2020-06-23
Set page title according to view

### LINKED ISSUE #2062: Page Title
currently we have only "Kubermatic" always and with several tabs that can be confusing

Suggestion:
add reverse breadcrumbs or the like, e. g. "[ClusterName]/[ProjectName] - Kubermatic"

kgroschoff:
Set page title according to view e.g. `members in project 'xyz'` and  `cluster 'frosty-pare' in project 'xyz'`. 
Updated version of #2377, which is from my old broken fork.

Fixes #2062

/

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/cluster-details/cluster-details.component.ts
+++ b/src/app/cluster/cluster-details/cluster-details.component.ts
@@ -335,7 +335,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
   }
 
   isSSHKeysEditEnabled(): boolean {
-    return MemberUtils.hasPermission(this._user, this._currentGroupConfig, 'sshKeys', Permission.Delete);
+    return MemberUtils.hasPermission(this._user, this._currentGroupConfig, 'sshkeys', Permission.Delete);
   }
 
   editSSHKeys(): void {
--- a/src/app/cluster/cluster-details/cluster-details.component.ts
+++ b/src/app/cluster/cluster-details/cluster-details.component.ts
@@ -16,6 +16,7 @@ import {
 import {SettingsService} from '../../core/services/settings/settings.service';
 import {Addon} from '../../shared/entity/addon';
 import {Cluster, ClusterType, getClusterProvider, MasterVersion} from '../../shared/entity/cluster';
+import {View} from '../../shared/entity/common';
 import {Datacenter} from '../../shared/entity/datacenter';
 import {Event} from '../../shared/entity/event';
 import {Health, HealthState} from '../../shared/entity/health';
@@ -335,7 +336,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
   }
 
   isSSHKeysEditEnabled(): boolean {
-    return MemberUtils.hasPermission(this._user, this._currentGroupConfig, 'sshkeys', Permission.Delete);
+    return MemberUtils.hasPermission(this._user, this._currentGroupConfig, View.SSHKeys, Permission.Delete);
   }
 
   editSSHKeys(): void {
--- a/src/app/core/services/page-title/page-title.service.ts
+++ b/src/app/core/services/page-title/page-title.service.ts
@@ -1,0 +1,127 @@
+import {Injectable} from '@angular/core';
+import {Title} from '@angular/platform-browser';
+import {Subject, of} from 'rxjs';
+import {takeUntil, switchMap, tap} from 'rxjs/operators';
+
+import {ProjectService, ClusterService, ApiService} from '../../services';
+import {ParamsService, PathParam} from '../../services/params/params.service';
+
+export enum PageView {
+  Clusters = 'clusters',
+  Projects = 'projects',
+  Members = 'members',
+  SSHKeys = 'sshkeys',
+  ServiceAccounts = 'serviceaccounts',
+  Wizard = 'wizard',
+  Account = 'account',
+  Settings = 'settings',
+  NodeDeployment = 'nd',
+}
+
+@Injectable()
+export class PageTitleService {
+  projectName: string;
+  clusterName: string;
+  ndName: string;
+  private _unsubscribe = new Subject<void>();
+
+  constructor(
+    private titleService: Title,
+    private readonly _params: ParamsService,
+    private readonly _projectService: ProjectService,
+    private readonly _clusterService: ClusterService,
+    private readonly _apiService: ApiService
+  ) {}
+
+  setTitle(url: string): void {
+    const viewName = this.getPageViewName(url.split('/'));
+    this.titleService.setTitle(viewName);
+
+    this._projectService.selectedProject
+      .pipe(tap(project => (this.projectName = project ? project.name : '')))
+      .pipe(
+        switchMap(() =>
+          this._params.get(PathParam.ProjectID) &&
+          this._params.get(PathParam.ClusterID) &&
+          this._params.get(PathParam.SeedDC)
+            ? this._clusterService.cluster(
+                this._params.get(PathParam.ProjectID),
+                this._params.get(PathParam.ClusterID),
+                this._params.get(PathParam.SeedDC)
+              )
+            : of(null)
+        )
+      )
+      .pipe(tap(cluster => (this.clusterName = cluster ? cluster.name : '')))
+      .pipe(
+        switchMap(() =>
+          this._params.get(PathParam.ProjectID) &&
+          this._params.get(PathParam.ClusterID) &&
+          this._params.get(PathParam.NodeDeploymentID) &&
+          this._params.get(PathParam.SeedDC)
+            ? this._apiService.getNodeDeployment(
+                this._params.get(PathParam.NodeDeploymentID),
+                this._params.get(PathParam.ClusterID),
+                this._params.get(PathParam.SeedDC),
+                this._params.get(PathParam.ProjectID)
+              )
+            : of(null)
+        )
+      )
+      .pipe(tap(nd => (this.ndName = nd ? nd.name : '')))
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(_ => {
+        if (this._params.get(PathParam.ProjectID)) {
+          if (this._params.get(PathParam.ClusterID)) {
+            if (this._params.get(PathParam.NodeDeploymentID)) {
+              this.titleService.setTitle(
+                "node deployment '" +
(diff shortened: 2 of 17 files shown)


src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+      case PageView.Wizard:
+        return 'wizard';
+      case PageView.Projects:
+        return 'projects';
+      case PageView.Account:
+        return 'user settings';
+      case PageView.Settings:
+        return 'admin panel';
+      default:
+        switch (urlArray[urlArray.length - 2]) {
```
maciaszczykm, line 117: Method does not check for array lenght.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
+import {Injectable} from '@angular/core';
+import {Title} from '@angular/platform-browser';
+import {Subject, of} from 'rxjs';
+import {takeUntil, switchMap, tap} from 'rxjs/operators';
+
+import {ProjectService, ClusterService, ApiService} from '../../services';
+import {ParamsService, PathParam} from '../../services/params/params.service';
+
+export enum PageView {
```
floreks, line 9: This is basically a copy of `View` enum from `auth.guard`. Export that to a common place i.e. `shared/entity`.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+        )
+      )
+      .pipe(tap(nd => (this.ndName = nd ? nd.name : '')))
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(_ => {
+        if (this._params.get(PathParam.ProjectID)) {
+          if (this._params.get(PathParam.ClusterID)) {
+            if (this._params.get(PathParam.NodeDeploymentID)) {
+              this.titleService.setTitle(
+                "node deployment '" +
```
floreks, line 78: Try to avoid hardcoded strings in code.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  setTitle(url: string): void {
+    const viewName = this.getPageViewName(url.split('/'));
+    this.titleService.setTitle(viewName);
+
+    this._projectService.selectedProject
+      .pipe(tap(project => (this.projectName = project ? project.name : '')))
+      .pipe(
+        switchMap(() =>
+          this._params.get(PathParam.ProjectID) &&
```
floreks, line 44: Export such long checks to a methods.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+
+@Injectable()
+export class PageTitleService {
+  projectName: string;
+  clusterName: string;
+  ndName: string;
+  private _unsubscribe = new Subject<void>();
+
+  constructor(
+    private titleService: Title,
```
floreks, line 29: Missing `readonly` and `_` as it is private.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+  Settings = 'settings',
+  NodeDeployment = 'nd',
+}
+
+@Injectable()
+export class PageTitleService {
+  projectName: string;
+  clusterName: string;
+  ndName: string;
+  private _unsubscribe = new Subject<void>();
```
floreks, line 26: Missing `readonly` property.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+                this._params.get(PathParam.SeedDC),
+                this._params.get(PathParam.ProjectID)
+              )
+            : of(null)
+        )
+      )
+      .pipe(tap(nd => (this.ndName = nd ? nd.name : '')))
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(_ => {
+        if (this._params.get(PathParam.ProjectID)) {
```
floreks, line 74: Try to refactor this to avoid such high cyclomatic complexity of the code.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+          } else {
+            this.titleService.setTitle(viewName + " in project '" + this.projectName + "'");
+          }
+        } else {
+          this.titleService.setTitle(viewName);
+        }
+      });
+  }
+
+  getPageViewName(urlArray: string[]): string {
```
floreks, line 98: Methods that are not supposed to be used outside of the service should be private for clarity.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    switch (urlArray[urlArray.length - 1]) {
```
floreks, line 99: Export page display names to an enum and refactor logic to i.e. 
1. Iterate over the enum values and try to find the name taken from url
2. Map found name to a display name.

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,127 @@
...
+                "node deployment '" +
+                  this.ndName +
+                  "' in cluster '" +
+                  this.clusterName +
+                  "' in project '" +
+                  this.projectName +
+                  "'"
+              );
+            } else {
+              this.titleService.setTitle("cluster '" + this.clusterName + "' in project '" + this.projectName + "'");
```
maciaszczykm, line 87: I think the title should always start with the capital letter.

floreks: IMHO we should actually use title case in the tab title.

maciaszczykm: Also, for the long strings you can use `` and ${} inside.

kgroschoff:
Thanks a lot. 😄 
Forgot the WIP label, as I also wanted to change a few things that you mentioned. Will also take care about the other points you made 😊

kgroschoff:
@maciaszczykm @floreks I hope I covered everything, though not entirely sure if everything is better now or even worse :trollface: PTAL 😃

src/app/shared/model/Config.ts
```diff
@@ -30,7 +30,7 @@ export interface Viewable {
 export class GroupConfig {
   projects?: Projects;
   members?: Members;
-  sshKeys?: SSHKeys;
+  sshkeys?: SSHKeys;
```
kgroschoff, line 33: This is a "not so nice" part. But as we already wanted to change the whole role management in the ui, this is probably the fastest "solution" as we use a shared "View" enum now for auth & the page title. All urls are lowerCase...

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,107 @@
...
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(_ => {
+        this._titleService.setTitle(this._generateTitle(viewName));
+      });
+  }
+
+  private _getViewName(urlArray: string[]): string {
+    if (!_.isEmpty(urlArray)) {
+      let viewName = Object.values(View).find(view => view === urlArray[urlArray.length - 1]);
+      if (_.isEmpty(viewName)) {
```
floreks, line 47: We don't really need `isEmpty` check for a simple string here. `!viewName` should be enough to check for empty/undefined string. Nesting `if` is also not needed here. This could be:

```typescript
let viewName: string;
// Here simple array.length > 0 should be enough since viewName is a result of split and split will always return at least empty array.
if(urlArray.length > 0) {
  viewName = Object.values(View).find(view => view === urlArray[urlArray.length - 1]);
}

// We need to check if array length is actually bigger than 1. isEmpty only checks if it's bigger than 0.
if(!viewName && urlArray.length > 1) {
  viewName = Object.values(View).find(view => view === urlArray[urlArray.length - 2]);
}

return viewName ? getViewDisplayName(viewName) : '';
```

src/app/core/services/page-title/page-title.service.ts
```diff
@@ -0,0 +1,107 @@
...
+      return getViewDisplayName(viewName);
+    }
+    return '';
+  }
+
+  private _generateTitle(viewName: string): string {
+    // titleString should at least display the viewName
+    let titleString: string = viewName;
+    const project = this._params.get(PathParam.ProjectID)
+      ? `${ViewDisplayName.Projects.slice(0, -1)} '${this.projectName}'`
```
floreks, line 59: I am not sure I follow. Why use `slice` here?

src/app/shared/model/Config.ts
floreks: I think that this actually fixes it. Auth guard does not use case insensitive comparison, so for ssh keys it should not work currently.

src/app/cluster/cluster-details/cluster-details.component.ts
```diff
@@ -335,7 +335,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
   }
 
   isSSHKeysEditEnabled(): boolean {
-    return MemberUtils.hasPermission(this._user, this._currentGroupConfig, 'sshKeys', Permission.Delete);
+    return MemberUtils.hasPermission(this._user, this._currentGroupConfig, 'sshkeys', Permission.Delete);
```
floreks, line 338: Since you are already renaming it can you use enum here instead of hardcoding it again?

src/app/core/services/page-title/page-title.service.ts
kgroschoff: Yes. Not entirely happy with this though...

src/app/shared/entity/common.ts
```diff
@@ -23,3 +23,52 @@ export enum ResourceType {
...
+  Members = 'Members',
+  SSHKeys = 'SSH Keys',
+  ServiceAccounts = 'Service Accounts',
+  Wizard = 'Wizard',
+  Account = 'User Settings',
+  Settings = 'Admin Panel',
+  NodeDeployment = 'Node Deployment',
+}
+
+export function getViewDisplayName(view: string): string {
```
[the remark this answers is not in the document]
kgroschoff: Absolutely! 👍

PR #2416, opened 2020-06-24
Rename Node Deployments to Machine Deployments

### LINKED ISSUE #2392: Machine Deployment vs Node Deployment
I've heard devs call the Node Deployment a Machine Deployment multiple times, it almost seems more straight forward to just change the naming in the UI, wdyt?

kgroschoff:
Rename Node Deployments to Machine Deployments

Fixes #2392

/

**Release note**:

```release-note
NONE
```


kgroschoff:
/hold
@floreks @maciaszczykm Right now only the visible parts are renamed. Thought about also changing everything (variables, components,...) inside the repo, to make it look more clean, but noticed that e.g. the endpoints use `nodedeployments`, so we will not be completely consistent anyways (don't know how many parts are using it in the backend, so maybe a never ending story). WDYT? Easier to keep it as it is and just rename the visible parts or deeply rename everything possible?

maciaszczykm:
If we want to merge it I would vote for changing the API too. It doesn't need to be done in the same moment but I think we should keep it aligned in the longer run. We can add the new endpoints and start the deprecation process of the current ones.

cc @zreigz @lsviben

zreigz:
The Kubermatic abstraction for the MachineDeployments is NodeDeployment. Everywhere in the API, we use Nodes and we have endpoints for this. It would be inconsistent to name those resources Machines having everywhere Nodes. We should first migrate our API Node object to Machine

kgroschoff:
/test test-e2e
--

kgroschoff:
@maciaszczykm I think (& hope) I covered everything in the UI. PTAL :)

maciaszczykm:
A lot of work done here :)

/lgtm
/hold

Please cancel the hold after you will create issues for changing the API and changing UI to the new API afterwards.

PR #2444, opened 2020-07-01
Fix default image for openstack for rhel and flatcar

kgroschoff:
Fix default image for openstack when RHEL or Flatcar is selected. 

/

**Release note**:

```release-note
NONE
```


kgroschoff:
FYI: Also added it to the "old" wizard as it isn't that much extra work. Though the existing code isn't very nice, I didn't spend any time to refactor it as the new wizard (hopefully) is available soon.

PR #2501, opened 2020-07-17
Improve e2e tests stability

floreks:
- Added new `TrafficMonitor` class to easily wait for request or expect a specific response from the API with the ability to retry.
- Updated `workerCount` in test to 100
- Refactored some magic strings to enums.
- Fixed some issues with properly waiting for project and selecting it. (full page refresh seems to help, but we might have some underlying issue here in the frontend)

**Release note**:

```release-note
NONE
```

PR #2454, opened 2020-07-03
Add license header to the sources

maciaszczykm:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

**Special notes for your reviewer**:

**Release note**:

```release-note
NONE
```

PR #2507, opened 2020-07-22
[v2.14] Improve e2e tests stability

floreks:
Manual cherry-pick of #2501 changes.

**Release note**:

```release-note
NONE
```


floreks:
It is more problematic than I expected. Tests against `v2.14` fail not because they are flaky, but because tests are using the latest API and due to changes in the API `v2.14` frontend will not work with the latest API.

It requires refactoring the way we run our tests to try and use proper API versions. This is a much bigger task, but I guess it is a high priority since we can't merge any fixes to `v2.14` now.

@kdomanski @kgroschoff

cypress/pages/members.po.ts
```diff
@@ -1,6 +1,19 @@
-import {Group} from '../utils/member';
-import {wait} from '../utils/wait';
+// Copyright 2020 The Kubermatic Kubernetes Platform contributors.
```
floreks, line 1: Should license headers be there in v2.14?

kgroschoff: Probably yes, but seems like the [related PR](https://github.com/kubermatic/dashboard/pull/2454) was only merged into master 🤔

floreks: We can backport it then separately

floreks:
Tests have passed 5 times in a row. I guess it is stable :slightly_smiling_face:

kgroschoff:
🎉 
/lgtm

floreks:
Added last small improvement to wait for project to be deleted before logging out.

PR #2520, opened 2020-07-28
Add EOL info component

### LINKED ISSUE #2497: Provide a warning when clusters are running EOL version of K8S
Many users will set up their cluster once and then neglect the need to upgrade its version. This results in running clusters which are at an EOL version containing vulnerabilities - a strong case for immediate upgrade. However we cannot always just automatically upgrade the clusters, because sometimes manual migrations of loads are required, e.g. due to API versions deprecated in v1.16. For that purpose we'd like to visually highlight clusters running EOL versions in the UI.

**Acceptance criteria*:
- [ ] clusters running versions `<1.16.0` are highlighted in the UI with a warning
  - [ ] in their respective detail views
  - [ ] on a project's cluster list
  - [ ] ~~in the project list, projects containing EOL clusters are highlighted similarly~~
- [ ] the warning is of a higher severity (e.g. in red) so that in the future we can use a lower severity visual warning to show clusters which are not yet EOL, but will soon be
- [ ] the specific version value `<1.16.0` is relatively easy to change in the future

floreks:
### Features
- [x] clusters running versions `<1.16.0` are highlighted in the UI with a warning
  - [x] in their respective detail views
  - [x] on a project's cluster list
  - [ ] ~in the project list, projects containing EOL clusters are highlighted similarly~
- [x] the warning is of a higher severity (e.g. in red) so that in the future we can use a lower severity visual warning to show clusters which are not yet EOL, but will soon be
- [x] the specific version value `<1.16.0` is relatively easy to change in the future

~### Design
@cschieder I could use some ideas here. We are already showing a warning icon for container linux clusters. I don't have any good ideas on how to nicely show more warnings/errors. We also need to differentiate between versions where end-of-life date has already passed (error) and not yet passed (warning).~

### # Update 2

Zrzut ekranu z 2020-08-14 12-02-58

Zrzut ekranu z 2020-08-14 12-03-12

Zrzut ekranu z 2020-08-14 12-04-30

Zrzut ekranu z 2020-08-14 12-04-41

Fixes #2497

**Release note**:

```release-note
Kubermatic Dashboard now supports end-of-life configuration for Kubernetes versions. All versions up to defined version will be marked as deprecated or soon-to-be deprecated in the UI. It can be defined in the config.json file as follows:
...
"end_of_life": {
  "1.17.9": "2020-07-01",
  "1.18.0": "2020-09-15"
}
```


floreks:
@kdomanski 

> in the project list, projects containing EOL clusters are highlighted similarly

What about a case where there are multiple clusters within the project and some of them are near EOL and some of them use already deprecated versions?

Should we show multiple warnings/errors? Show only error with generic information about status?

floreks:
@cschieder PTAL at new screenshots.

PR #2539, opened 2020-08-05
Update e2e docker image

floreks:
Seems that e2e now require Go version 1.13 or newer. I have updated it and other dependencies to newer versions.

Removed `e2e-kind` image files as it was moved to infra repo.
https://github.com/kubermatic/infra/tree/master/docker_images/e2e-kind

**Release note**:

```release-note
NONE
```


floreks:
Other PRs will have to be rebased after this gets merged.

PR #2541, opened 2020-08-05
Choose smallest/cheapest instance as the default one

floreks:
- Updated providers logic in the wizard to choose the smallest/cheapest instance by default
- Updated display name format to show all information after selecting an instance
- Fixed an issue with triggering refresh of sizes and related endpoints when datacenters are refreshed

**Release note**:

```release-note
Cheapest/smallest provider instance will be chosen by default in the wizard
```


floreks:
NPM has some issues and tests can't download dependencies. We'll have to wait.

Issue #2367, opened 2020-06-05
Design editor


maciaszczykm:
**For what do you want a design draft?**: Text editor that will be added in the https://github.com/kubermatic/dashboard/pull/2353. If the pull request is already merged it can be seed during creation/edit of dynamic datacenters. It would be good to establish some styling so we can reuse it i.e. in the object YAML editor.

**Requirements how it should look like / what should be in**: We are using the https://microsoft.github.io/monaco-editor/.

maciaszczykm:
<img width="805" alt="Zrzut ekranu 2020-06-19 o 09 17 23" src="https://user-images.githubusercontent.com/2823399/85107119-b50bf480-b20d-11ea-8030-f4ad35c12617.png">
<img width="791" alt="Zrzut ekranu 2020-06-19 o 09 17 01" src="https://user-images.githubusercontent.com/2823399/85107129-b806e500-b20d-11ea-9322-3b1548291593.png">

This is the current version.

cschieder:
![image](https://user-images.githubusercontent.com/49145827/90729770-5aebe600-e2c7-11ea-9333-2ec3c2cee2f9.png)

font: Inconsolata, size 14

PR #2566, opened 2020-08-18
Styling improvements

floreks:
#### Changes
- [x] Improve style files loading order
- [x] Get rid of all `::ng-deep` usages
- [x] Get rid of all/most `!important` usages
- [x] Move styling logic not related to theming to `main.scss` file
- [x] Clean up generic styles and move as much styling to component-level styles
- [x] Check views and make sure styling is same as before
- [x] Redesign the editor (fixes https://github.com/kubermatic/dashboard/issues/2367)

**Release note**:

```release-note
NONE
```


maciaszczykm:
I'll work on both settings views.

maciaszczykm:
/assign @kgroschoff 

Can you review it? If you want we can setup a meeting so I will show you what was changed here.

kgroschoff:
I just checked it out locally and found quite a few things that are looking different at the moment/aren't working as I would expect it. Sorry for all the text 😅 
Probably some could be fixed afterwards, but I will list everything that I noticed so far.
Code review itself will follow.

- Projects View:
  - [ ] Project switcher has no color on change. `All Projects` was blue previously.
- User Settings:
  - [ ] `None` is not selected even though I have no default project
- Admin Settings:
  - [ ] `Dynamic Datacenters` & `Administrators` seems to be too far to the right. Or maybe it's the content that it too far to the left?
- Wizard:
  - [ ] `Add SSH key` button is placed too far to the right, isn't it?
  - [ ] When closing the `Add SSH key` dialog without adding an ssh key, I get following error in the dev console `ERROR EmptyErrorImpl {message: "no elements in sequence", name: "EmptyError"}`
  - [ ] DO: `Provider Tags` field isn't at 100% width.
  - [ ] AWS/Alibaba/Azure/GCP/Openstack views are too wide. (classic case:) Dev Console on the right is enough to get frustrated, because you need to scroll to click on the buttons. I know that this also isn't fixed completely on dev right now, but it wasn't a problem in all steps so far.
  - [ ] AWS/Azure: checkbox for `Assign public IP` is centered. Is this intended?
  - [ ] GCP: `Preemptible` see above.
  - [ ] Openstack: `Allocate Floating IP` see above.
  - [ ] VSphere: `Use custom VSphere Cloud Credentials` see above.
  - [ ] The delete icon in dropdowns is placed above the text. Only noticed it on Packets Sizes, but probably in theory the same for all others with long input.
  - [ ] Summary: The distance between single rows seems to be too big. e.g. between `x Monitoring` and `x IPv` on Digitalocean.
- Cluster Details burger menu on the right:
  - [ ] Is it just me or is the font size for the menu items smaller than before?
  - [ ] Edit SSH Keys: The `Add SSH key` row is not clickable on full width - only when you directly click on the text.
- Cluster Delete Dialog:
  - [ ] The `!` icon with it's text isn't centered vertically. Also it seems like the distance from icon to text is a bit too big?
- Service Account View, specifically the Token View:
  - [ ] If you have a brand new Service Account without a Token, you will only see `No tokens available` (fine so far), but the table has black borders on the right and on the left. The table get nice borders when a token was added.
- Update views in generally is not 100% working. Not sure if I found everything (so maybe double check needed), but most prominent are the following: 
  - [ ] Noticed it when I tried to delete project members. I had to manually go to the page again to get the reduced list. 
  - [ ] Same as above for projects? Seems like they are in deletion forever.
  - [ ] Also when I added a member as owner to a project and go back to the project view, the owners in this project only got updated after I reload the page. (maybe I was too fast)
  - [ ] Wasn't too sure about the cluster details view update. Felt like it only changed when I re-entered the page, but maybe this was my bad timing...
  - [ ] Delete Clusters in the cluster list directly felt like it was deleting forever. Re-entered the page and it was gone. So eather it took long or also this reloading thingy. 
  - [ ] Adding a Token to a Service Account doesn't refresh the table. Only if you close & open the accordion again. Same on deletion.

maciaszczykm:
> Update views in generally is not 100% working.

@floreks put an easter egg before his vacation for us 😃 

```
refreshTimeBase: 60000, // Unit: ms
```

maciaszczykm:
I think I've fixed most of the stuff mentioned here. I've opened two issues for things that exist also on master.

PTAL @kgroschoff

maciaszczykm:
@kgroschoff The tests are fixed, please take a look :)

src/app/dynamic/enterprise/theming/picker/style.scss
```diff
@@ -12,27 +12,20 @@
 @import 'variables';
 @import 'mixins';
 
-.km-account-settings-entry-label {
+.entry-label {
```
kgroschoff, line 15: More as a general question: 
Is it intended that there is no `km` prefix? 
And if yes: How are we deciding if we prefix it or not? Already saw the not-prefixed-version in some places, where it is in a `km-[...]` class, which makes sense.

kgroschoff (commented):
great work 👍 
Just a small question and because of it I will wait with the LGTM ;)

src/app/dynamic/enterprise/theming/picker/style.scss
maciaszczykm: Yes, it is intended. There is no need to prefix values that are part of component-level stylesheets as they are encapsulated (with exception if we change it manually in the component config). However, when it comes to the global styles it is better to keep the prefix, especially the part that points to the component or view, i.e. `*-account-settings-*`. `km-*` in general is up to us, but at the moment it is used mostly in global rules. It can be changed later though once we decide if we want to keep it or not but it may be better to have it to avoid issues that may be hard to find.

maciaszczykm: By the way, `*theme.scss` files that have theme mixins and are on component-level should be also considered global as they are referenced from global files, not from components themselfs.

maciaszczykm: We've added them to keep the rules next to each other, thanks to this it's easier to find and modify the component look.

kgroschoff: Alright, sounds good 👍 Thanks 😄

PR #2612, opened 2020-09-10
[WIP] Guided Tour

### LINKED ISSUE #2518: Guided Tour
For the sake of separating this topic from #1934 

Prototype (outdated)
https://www.figma.com/proto/dJQKgUPzIDND36o2Nq8dct/Loodse-UI?node-id=3547%3A18325&viewport=969%2C-282%2C0.338676780462265&scaling=min-zoom

kgroschoff:
This PR adds the Guided Tour Feature.

[now a bunch of text is following, but since there are a lot of changes it might make sense to explain a few things]

I thought it would be best if we don't have to adjust the existing code base too much, so all you need to do to add the `kmGuidedTour` directive to the part where the step should pop up. To this directive you have to add an id, e.g. `kmGuidedTour="gt-start-tour"`. It is also possible to additionally add `(next)`, `(prev)` or `(done)` and define what should happen when you click next, prev or done button. (e.g. `src/app/shared/components/add-project-dialog/add-project-dialog.component.html`)
But in most cases you don't have to define next, prev or done. Also in most cases the directive is enough and no further changes in the .ts files needs to be done. However sometimes there are some exceptions. ¯\_(ツ)_/¯

Detailed information about the content of this step are currently defined inside the `GuidedTourItemsService`, because I don't want to have a lot of text in the original tag. Maybe here we'll find something better, but for now I like the idea that everything is in one file and I can easily add or adjust the step content(s).
Each item could (or should) have following information:
| Field | Description |
| ----- | ----- |
| id: string | This one has to be the same that you choose for the directive. |
| title?: string | Title of the item. |
| text?: string | Text of the item. |
| nextText?: string | In case you don't want to display `next` on each button, but maybe `create` or `got it` or whatever. |
| stepPosition?: string | Where this step should be placed. Default is already set via the "global" options, but e.g. the step for the project dropdown should be placed at the bottom of it, while the step for the menu items should be placed at the right. |
| route?: string | The route to this step, e.g. `/projects`. This one makes it fairly easy to jump between all our different views. |

The Guided Tour itself could be started via the new Help Panel on the top right (at the moment... maybe there could also be a button on the projects view if you don't have a project). Codewise you only need to call `guidedTourService.startTour()`. Right now I directly defined all options inside this function, since we probably only have this one guided tour, but it would also be pretty easy to just move them to the `startTour()` call.
Following options are possible:
| Field | Description |
| ----- | ----- |
| steps: string[] | All id's of steps that should be displayed. This also is the order in which the steps are getting displayed. |
| startWith?: string | Which step should be the first (thought this could be useful if we want to have one tour with different starting points? could also be removed pretty easily) |
| stepDefaultPosition?: string | Default position where all steps should be placed. top, bottom, left, right, center |
| showPrevButton?: boolean | If the prev button should be displayed. Isn't in the current mockups, but stepping back might be useful? |
| waitingTime? | In a few moments I noticed that it is useful to define a waiting time, even though it's just a milisecond... Decided to put it, but easy to remove. |

For all the fake data I also created some mocks in the `GuidedTourItemsService`. I am not entirely happy with this, but since I already have a lot of info in here, it has been good enough for the start. Atm to use this data and to not change too much in a component itself, I just added it to the services with all endpoints, e.g. for getting a dedicated project:
```
private _getProject(projectID: string): Observable<Project> {
  if (this._guidedTourService.isTourInProgress()) {
    return of(this._guidedTourItemsService.guidedTourProject());
  }

  if (!projectID) {
    return EMPTY;
  }

  const url = `${this._restRoot}/projects/${projectID}`;
  return this._http.get<Project>(url).pipe(catchError(() => of<Project>()));
}
```
If the Guided Tour is in progress, I just return the faked data and don't call the api endpoint. Other ideas are welcome. However, this is a pretty simple and fast solution without that many changes.

I think this are the most "important" things so far. If not feel free to ask ;) 

Also it isn't perfectly working/looking/coded yet and a few things are missing (maybe more, we'll figure it out):
- [ ] "Hide" projects or clusters before creation and only display them if we "created" them via the tour
- [ ] Machine Deployments view
- [ ] Styling isn't perfect yet
- [ ] Placement of the step sometimes is a bit "broken", e.g. Wizard -> Step Node Settings or Summary, where the whole dialog is too high and sometimes is hidden on "smaller" windows.
- [ ] New conventions for e.g. names needs to be applied to all new files (regarding to the styling improvements PR)
- [ ] A lot of tests need to get fixed 😅 
- [ ] Add e2e tests
- [ ] Circular dependency needs to be fixed: 
```
Circular dependency detected:
src\app\core\services\guided-tour\guided-tour-step.service.ts -> src\app\shared\components\guided-tour\step\step.component.ts -> src\app\core\services\guided-tour\guided-tour-step.service.ts
Circular dependency detected:
src\app\shared\components\guided-tour\step\step.component.ts -> src\app\core\services\guided-tour\guided-tour-step.service.ts -> src\app\shared\components\guided-tour\step\step.component.ts
```

Fixes #2518

/

**Release note**:

```release-note
Adding a help panel with a guided tour to the ui
```


kgroschoff:
@maciaszczykm / @floreks I know it is a lot and not 100% done yet, but I think this is the perfect time to get some feedback from your side. (since I'm on vacation next week and we are going to have the "summerparty" afterwards, just have a look whenever it fits you)

maciaszczykm:
We need e2e test for this as risk of breaking stuff here is high but let's wait for review and fixes.

src/app/cluster/cluster-details/cluster-details.component.ts
```diff
@@ -106,6 +109,10 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
       .getCurrentUserGroup(this.projectID)
       .subscribe(userGroup => (this._currentGroupConfig = this._userService.getCurrentUserGroupConfig(userGroup)));
 
+    if (this._guidedTourService.isTourInProgress()) {
```
floreks, line 112: The biggest issue here is that directive usage should be independent and not require logic changes inside the components that use it. This `if` is reused 65 times across this change. We can't have such hard dependency. It should be minimized to some common service/interceptor etc. so that we won't have to remember about adding it manually when using it.

src/app/core/services/guided-tour/guided-tour-step.service.ts
```diff
@@ -0,0 +1,329 @@
...
+  }
+
+  private _scrollWhenTargetOrStepAreHiddenTop(): void {
+    const totalTargetTop = this._getMaxTargetAndStepTopPosition();
+    if (totalTargetTop < this._winTopPosition) {
+      window.scrollBy(0, totalTargetTop - this._winTopPosition);
+    }
+  }
+
+  private _getMaxTargetAndStepBottomPosition(): number {
```
maciaszczykm, line 214: It looks like something that could be done from stylesheets. Can you explain this?

src/app/core/services/guided-tour/guided-tour.service.ts
```diff
@@ -0,0 +1,87 @@
...
+    }
+  }
+
+  isTourInProgress(): boolean {
+    return this._tourInProgress;
+  }
+
+  getGuidedTourOrder(): string[] {
+    return [
+      'km-gt-start-tour',
```
maciaszczykm, line 60: Please change to `enum` and use it in the templates.

src/app/shared/components/guided-tour/arrow/arrow.component.scss
```diff
@@ -0,0 +1,34 @@
...
+.guided-tour-arrow-right {
+  border-bottom: 11px solid transparent;
+  border-left: 11px solid #fff;
+  border-top: 11px solid transparent;
+}
+
+.guided-tour-arrow-left {
+  border-bottom: 11px solid transparent;
+  border-right: 11px solid #fff;
+  border-top: 11px solid transparent;
```
maciaszczykm, line 33: This has to be moved to theme file with all the color-related styles.

src/app/core/services/guided-tour/guided-tour-steps-container.service.ts
```diff
@@ -0,0 +1,112 @@
...
+import {GuidedTourStep, Step, StepActionType} from '../../../shared/entity/guided-tour';
+import {GuidedTourOptionsService} from './guided-tour-options.service';
+import {GuidedTourItemsService} from './guided-tour-items.service';
+
+@Injectable()
+export class GuidedTourStepsContainerService {
+  private _steps: Step[];
+  private _tempSteps: GuidedTourStep[] = [];
+  // eslint-disable-next-line @typescript-eslint/no-magic-numbers
+  private _currentStepIndex = -2;
```
maciaszczykm, line 24: Why -2?

src/app/core/services/guided-tour/guided-tour-items.service.ts
```diff
@@ -0,0 +1,392 @@
...
+      },
+      status: {
+        url: 'https://4k6txp5sq.europe-west3-c.dev.kubermatic.io:30002',
+        version: '1.18.8',
+      },
+      type: ClusterType.Kubernetes,
+    };
+  }
+
+  guidedTourDOMachineDeployment(): MachineDeployment {
```
maciaszczykm, line 249: This would need pretty high maintenance effort as we would have to keep it with sync all the time. I'd rather not keep data like this in a service.

maciaszczykm: I am not sure if it wouldn't be better to operate on real data instead of mocks.

maciaszczykm: What would happen if client doesn't have DigitalOcean enabled?

src/app/core/services/guided-tour/guided-tour-items.service.ts
```diff
@@ -0,0 +1,392 @@
...
+import {Project} from '../../../shared/entity/project';
+import {Cluster, ClusterType, MasterVersion} from '../../../shared/entity/cluster';
+import {Node} from '../../../shared/entity/node';
+import {MachineDeployment} from '../../../shared/entity/machine-deployment';
+import {DigitaloceanSizes} from '../../../shared/entity/provider/digitalocean';
+import {Health} from '../../../shared/entity/health';
+
+@Injectable()
+export class GuidedTourItemsService {
+  getGuidedTourItems(): GuidedTourItem[] {
```
floreks, line 23: Having all the data inside a single service will be a very high maintenance effort. Those items should be separate component-like objects, similar to dialogs, where we can define custom logic and view per item.

src/app/core/services/guided-tour/guided-tour-items.service.ts
floreks: We should definitely operate on real data and only guide user instead of doing actions for him. Inside the wizard, different customers might use different configurations and we should have a generic way to guide them through it.

src/app/core/services/guided-tour/guided-tour-step.service.ts
floreks: We should rely on styles wherever possible. Trying to manipulate DOM and CSS from the code should be the last thing we should try if everything else fails.

src/app/core/services/guided-tour/guided-tour-steps-container.service.ts
kgroschoff: Was a leftover 😅

kgroschoff:
closing this one for the new discussed approach and smaller steps

Issue #2615, opened 2020-09-11
UI/UX Bugs/Improvements/Questions


cschieder:
clicked around on dev a bit and came up with some issues

### Bugs

- [x] Sometimes in project dropdown it scrolls/jumps back down to the project you're currently on, I assume this is related to refresh? tried it when having external projects turned on, you have to wait for it sometimes
- [x] Admin Settings - Cannot write anything into a new custom link, instantly deletes
- [x] Admin Settings - Info icons are missing? Tooltip on hover where they should be is still there

### Fixes

- [x] Cluster Page - when hovering over the kubernetes version can we remove one of the borders in the middle? or even both? see what looks better
- [x] Links hover effect missing (see cluster -> share, visit our documentation) hover: #4DC3EA
- [x] Blue button hover effect missing -> #4DC3EA
- [x] Dropdown x should be same size as the down arrow optically and horizontally aligned with it, it keeps being wrong one way or another, the circle on hover is also not center aligned with the arrow, sadly I'm not really sure what numbers/changes to give you, would be great if we can make it work though :)
- [x] Cluster Page -> Revoke Token - Spacing everywhere between checkbox and text should be 15px, then increase spacing between "Revoke Admin Token" and Revoke Viewer Token" so it's clearer

**Cluster Create**

- [x] Initial Nodes - Name Hint change to ---> Leave this field blank to use automatic name generation.
- [x] Initial Nodes - Provider Tags headline too little spacing on top

**Admin Settings page**
Idea: Dynamic Datacenters - in Provider dropdown use provider logo?

### Questions
Shouldn't bring your own be KubeADM in Provider Dropdown on Admin Settings? It's confusing
Can the user not set items per page on the page that he's viewing itself, only in the settings?

maciaszczykm:
> Shouldn't bring your own be KubeADM in Provider Dropdown on Admin Settings? It's confusing

KubeAdm is actually bare-metal. Bring your own is placeholder for any further types it seems but I am not sure why it was added. @kgroschoff do you know this?

maciaszczykm:
> Can the user not set items per page on the page that he's viewing itself, only in the settings?

Yes, that's the case at the moment.

kgroschoff:
Looooong time ago KubeAdm was called Bring your own. But as far as I know we only changed the frontend logo... so in kubermatic kubeAdm === byo. Maybe we also should refactor it

maciaszczykm:
[...]

PR #2622, opened 2020-09-14
Fix custom links and missing info icon

floreks:
Fixed issues:
- [x] Admin Settings - Cannot write anything into a new custom link, instantly deletes
- [x] Admin Settings - Info icons are missing? Tooltip on hover where they should be is still there

Fixes 2 issues mentioned in https://github.com/kubermatic/dashboard/issues/2615

**Release note**:

```release-note
NONE
```


maciaszczykm:
The links are not initialized when I enter from the different view, i.e. project view:

Zrzut ekranu 2020-09-15 o 09 31 54

They load after refresh.

PR #2640, opened 2020-09-18
Fix text overlapping clear button in our custom combobox

### LINKED ISSUE #2601: The delete icon in dropdowns overlaps long texts
Only noticed it on Packets Sizes, but probably in theory the same for all others with long input.

floreks:
Fixes #2601

Zrzut ekranu z 2020-09-18 14-06-49

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/shared/components/combobox/style.scss
+++ b/src/app/shared/components/combobox/style.scss
@@ -16,3 +16,18 @@ input {
   outline: none;
   padding: 16px 0 16px 16px;
 }
+
+mat-select {
+  &.mat-select {
+    mat-select-trigger {
+      div {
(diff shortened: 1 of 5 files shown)


src/app/shared/components/combobox/style.scss
```diff
@@ -16,3 +16,18 @@ input {
   outline: none;
   padding: 16px 0 16px 16px;
 }
+
+mat-select {
```
maciaszczykm, line 20: Doesn't it happen anywhere else that we have close icon inside mat-select? I'm asking as this one will fix it only for combobox component.

floreks: Good point. I'll add a global class.

PR #2644, opened 2020-09-22
Add option to enable/disable external cluster import from admin settings

### LINKED ISSUE #2642: Add option to disable external clusters feature from admin settings

floreks:
API error with information about the disabled external cluster import feature is silenced and `Connect Cluster` button is hidden after disabling the feature. We could think about improving the error behavior of handling the GET of a single cluster in case someone uses a direct link? Not sure what should be done in this case though.

Fixes #2642

**Release note**:

```release-note
Add option to enable/disable external cluster import feature from admin settings.
```

### DIFF
--- a/src/app/core/interceptors/error-notifications/error-notifications.service.ts
+++ b/src/app/core/interceptors/error-notifications/error-notifications.service.ts
@@ -19,7 +19,7 @@ import {NotificationService} from '../../services';
 export class ErrorNotificationsInterceptor implements HttpInterceptor {
   private readonly _notificationService: NotificationService;
   // Array of partial error messages that should be silenced in the UI.
-  private readonly _silenceErrArr = ['external cluster'];
+  private readonly _silenceErrArr = ['external cluster functionality'];
 
   constructor(private readonly _inj: Injector) {
     this._notificationService = this._inj.get(NotificationService);
--- a/src/app/core/interceptors/error-notifications/error-notifications.service.ts
+++ b/src/app/core/interceptors/error-notifications/error-notifications.service.ts
@@ -19,7 +19,7 @@ import {NotificationService} from '../../services';
 export class ErrorNotificationsInterceptor implements HttpInterceptor {
   private readonly _notificationService: NotificationService;
   // Array of partial error messages that should be silenced in the UI.
-  private readonly _silenceErrArr = [];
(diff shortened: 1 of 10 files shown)


src/app/core/interceptors/error-notifications/error-notifications.service.ts
```diff
@@ -34,20 +34,32 @@ export class ErrorNotificationsInterceptor implements HttpInterceptor {
...
-              `Error ${errorInstance.status}: ${
-                errorInstance.error.error.message || errorInstance.message || errorInstance.statusText
-              }`
-            );
-          } else if (
-            errorInstance.message &&
-            this._silenceErrArr.every(partial => !errorInstance.message.includes(partial))
-          ) {
-            this._notificationService.error(`${errorInstance.message}`);
+          if (this._shouldSilenceError(errorInstance)) {
```
maciaszczykm, line 37: Can we silence the error only if the feature is disabled? Otherwise, we might silence something that should be visible.

src/app/settings/admin/admin-settings.component.html
```diff
@@ -170,6 +170,19 @@
         <km-settings-status [isSaved]="isEqual(settings.enableOIDCKubeconfig, apiSettings.enableOIDCKubeconfig)"></km-settings-status>
       </div>
 
+      <div fxLayout="row">
+        <div fxFlex="16%"
+             fxLayoutAlign=" center"
+             class="entry-label">
+          <span>Enable External Cluster Import</span>
```
maciaszczykm, line 177: Instead of `Import` I'd use `Connection` like we do in other views.

maciaszczykm: Or just `Enable External Clusters`.

src/app/core/interceptors/error-notifications/error-notifications.service.ts
floreks: I'd rather make partial messages more detailed. Everything is handled now in a generic way and it would be hard to keep it that way if we have to check for an exact single error and additionally asynchronously get admin settings to find out if some feature is enabled/disabled.

maciaszczykm: Ok, sure.

Issue #2564, opened 2020-08-17
UI/UX Fixes & Discuss


cschieder:
**Fixes**
- [x] Creating new project, big delay until project shows up in table/as a card 
- [ ] User with viewer role sees ssh key management as not greyed out in ui 

**To Discuss**
- Delete Project name copy function missing, argubly a hurdle for users that we might want to keep
- Evaluate keyboard usability, enter on modals to confirm dialogue

floreks:
> Misstyping vcenter credentials and then moving focus away from the credential inputs totally breaks the ui, it will then repeatedly try to use these credentials to fetch data (eg available networks etc), spam error toasts, ramp the browser cpu usage up

I can't reproduce that. After providing invalid credentials it shows an error, but everything still works. I can remove old credentials and provide the new one and everything works. Do you have better steps to reproduce?

cschieder:
@erdii received this feedback from customer, can you give us more info, Joshi?

floreks:
Anything related to the wizard that comes from customer is invalid for us as we are using a new wizard now and they are still using the old one.

cschieder:
Ok I'm deleting it

erdii:
> Anything related to the wizard that comes from customer is invalid for us as we are using a new wizard now and they are still using the old one. It would have to be checked and fixed separately for v2.14.

@floreks can customers already try the new wizard? we just had the credential problem multiple times in a row

maciaszczykm:
@erdii The new wizard was introduced in https://github.com/kubermatic/dashboard/pull/2481 as the default one on master. It should be fully functional at the moment.

floreks:
It is available on dev, but not for the clients. It will be a part of v2.15 release.

PR #2645, opened 2020-09-22
Fix call order in the SSH key view

maciaszczykm:

Fixes https://github.com/kubermatic/dashboard/issues/2564

**Release note**:

```release-note
Fix loading of the access rights in the SSH keys view.
```

### DIFF
--- a/src/app/sshkey/sshkey.component.ts
+++ b/src/app/sshkey/sshkey.component.ts
@@ -15,7 +15,7 @@ import {MatPaginator} from '@angular/material/paginator';
 import {MatSort} from '@angular/material/sort';
 import {MatTableDataSource} from '@angular/material/table';
 import {Subject, timer} from 'rxjs';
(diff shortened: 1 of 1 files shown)


kgroschoff:
I'm still able to see the ssh keys, but not to add a new one. Is this meant by `User with viewer role sees ssh key management as not greyed out in ui`? 
Asking because when the whole user management was added, this was the requested behaviour: `"sshkeys": {"view": true, "edit": false, "create": false, "delete": false}`. We obviously could change that. Just wondering about the issue 😅 

ssh

Apart from that, code looks good

maciaszczykm:
> Is this meant by User with viewer role sees ssh key management as not greyed out in ui?

Yes, I think so. The view rights are here and backend also supports them. Only the add, edit and delete should be blocked and it wasn't working because of wrong initialization.

Issue #2592, opened 2020-08-31
Use shorter file names


maciaszczykm:
**What would you like to cleanup?**: At the moment we are following `*.component.*` pattern in most of the places when it comes to naming files. It would be better to use shorter names, i.e. `style.scss` or `template.html` as the directory names already contain component names.

maciaszczykm:
This is already done in the new wizard.

maciaszczykm:
We should wait until the stylesheet refactoring will be done as doing it now will cause a lot of conflicts.

PR #2691, opened 2020-10-06
Refactor entry components and API types

floreks:
- Added `unsafe-perm` to the install task. Previously `ngcc` was not able to run due to privileges issue.
- Added import path shortcuts (i.e. IntelliJ now marks some imports with suggestions to use shorter version):
  - `@shared` - `src/app/shared`
  - `@core` - `src/app/core`
  - `@app` - `src/app`
  - `@assets` - `src/assets`
  - `@environments` - `src/environments`
- Refactored `km-machine-networks` component
- Removed `index.ts` that gathered shared services as it was more likely to introduce cyclic dependencies this way
- Updated most imports to use new, shorter versions
- Renamed some component files to use new naming style. Ref: #2592

**Release note**:

```release-note
NONE
```


jest.config.js
```diff
@@ -13,4 +13,11 @@ module.exports = {
       tsConfig: 'src/tsconfig.spec.json',
     },
   },
+  moduleNameMapper: {
```
maciaszczykm, line 16: 👍🏻

PR #2747, opened 2020-10-27
Allow to add ssh keys for kubeadm provider

kgroschoff:
Allow to add ssh keys for kubeadm provider

Fixes https://github.com/kubermatic/kubermatic/issues/6086

/

**Release note**:

```release-note
Support User SSH Keys in KubeAdm cloud provider
```


kgroschoff:
@maciaszczykm Could you have a look please? :) 
FYI: I found a bug related to the assignment of ssh keys: https://github.com/kubermatic/dashboard/issues/2748. If you directly add ssh keys in your first cluster all other clusters afterwards will also have this ssh keys assigned, even though you didn't specified them. ;) I would like to fix it in a new pr, as this also might be a problem in older releases (haven't checked yet)

floreks:
How about moving ssh keys component to the previous step as it is not related to the provider, but rather to the cluster itself?

We wouldn't have to worry about adding special logic to distinguish providers.

The bug you found should be very easy to fix. Probably ssh keys stored in the service (cache) are not cleared after closing the wizard.

kgroschoff:
Thanks @floreks, I think SSH keys fits very well into the cluster section 👍 
PTAL :) 

cluster-wizard

As the SSH keys are now moved, I also moved them to the cluster section in the summary view:
summary-wizard

kubeadm-details

floreks:
Could you reset `_sshKeys` array in the `ClusterService` used by the wizard to fix the issue with cached keys being reused?

/hold

PR #2782, opened 2020-11-12
Provide a better version info

xrstf:
As we tag releases on the release/XXX branches, a `git describe` on the master branch does not yield useful information. Currently it says something like `v2.14.0+255.g24fa837d`, because we more or less accidentally tagged that version on the exact moment we split the release branch away.

This PR changes the behaviour to treat master as 2.16.0. It was easier to hardcode the version into the Makefile than trying to do arithmetic and trying to calculate it based on the latest release branch.

I couldn't see where all the fields from the VersionInfo type are actually used, so maybe we can now skip the git-describe dependency alltogether?

**Release note**:
```release-note
NONE
```

### DIFF
--- a/version.js
+++ b/version.js
@@ -18,7 +18,7 @@ const gitInfo = gitDescribeSync({
 gitInfo.edition = getEditionDisplayName();
 
 // Re-use the version logic from our Makefile
-gitInfo.ui = execSync("make version").toString().trim();
+gitInfo.ui = execSync("make version --no-print-directory").toString().trim();
 
 const versionInfoJson = JSON.stringify(gitInfo, null, 2);
 
--- a/version.js
+++ b/version.js
@@ -1,6 +1,7 @@
 const {gitDescribeSync} = require('git-describe');
 const {resolve} = require('path');
 const {writeFileSync} = require('fs');
+const {execSync} = require('child_process');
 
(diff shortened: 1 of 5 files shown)


version.js
```diff
@@ -16,6 +17,9 @@ const gitInfo = gitDescribeSync({
 // Append edition information
 gitInfo.edition = getEditionDisplayName();
 
+// Re-use the version logic from our Makefile
+gitInfo.ui = execSync("make version --no-print-directory").toString().trim();
```
floreks, line 21: nit: I'd maybe change `ui` to something like `humanReadable`, `humanFriendly`, `readable`.

@kgroschoff could you check if this doesn't conflict with your workflow? As you are working on windows I am kind of worried it could break something for you.

kgroschoff: Does not break something, but the version isn't entirely correct for me ;) `"ui": "v2.14.0-257-g15c8a66b"`
I personally don't have a big problem with it, as it is only local 🤷

xrstf: FYI, the "wrong" version will also be generatedon linux if you're not on the master branch. The Makefile logic is setup to work primarily for CI (i.e. where it matters). To make it work in local release branches, more logic would be necessary and I think none of us want to make the Makefile even more complicated ^^

Issue #2828, opened 2020-12-04
Closing delete member dialog without confirmation produces console error


maciaszczykm:
**What happened?**: 
`EmptyErrorImpl {message: "no elements in sequence", name: "EmptyError"}` is shown in the console.

**What should happen?**:
Observable should be handled properly. Possibly this is related to usage of `first()` method. `take(1)` is safer.

**How to reproduce?**:

**Environment**
- UI Version:
- API Version:
- Domain:
  - [ ] run
  - [ ] run-2
  - [ ] dev
  - [x] locally
- Others:

PR #2840, opened 2020-12-07
Replace first() with take(1)

maciaszczykm:
**What this PR does / why we need it**: `EmptyErrorImpl {message: "no elements in sequence", name: "EmptyError"}` was shown after closing some dialogs etc. It was caused by filtering observable before it was passed through `first()` pipe. `first()` always require at least one object in the observable and if the requirement is not met then error will be produced. `take(1)` doesn't have this requirement and that is why I think we should use it in most of the cases.

Fixes https://github.com/kubermatic/dashboard/issues/2828

**Release note**:

```release-note
NONE
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,6 +157,12 @@
       "rxjs/no-nested-subscribe": "error",
       "rxjs/no-unbound-methods": "error",
       "rxjs/no-unsafe-takeuntil": "error",
+      "rxjs/ban-operators": [
+        "error",
+        {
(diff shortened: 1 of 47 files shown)


maciaszczykm:
/hold

I will check if I can add lint check for that.

maciaszczykm:
I've also added check:

Zrzut ekranu 2020-12-7 o 12 41 36

cc @kgroschoff

Issue #2871, opened 2020-12-11
Notification popups fail to escape HTML


xrstf:
**What happened?**:
I created an SSH key named `<u>test</u>`. The popup notification I got about its successful creation showed that the HTML was not escaped properly.

![2020-12-11 13_39_35-](https://user-images.githubusercontent.com/127499/101904605-6042a880-3bb6-11eb-8e41-ffec63b1d9c9.png)

**What should happen?**:
HTML should have been escaped.

**How to reproduce?**:
See above.

**Environment**
- UI Version: latest
- API Version: latest
- Domain: other
- Others: none

maciaszczykm:
The designs do not include bold fonts etc. I'll remove the possibility to use HTML inside at all.

PR #2885, opened 2020-12-21
Do not allow HTML inside notification messages

maciaszczykm:

Fixes https://github.com/kubermatic/dashboard/issues/2871

**Release note**:

```release-note
NONE
```


maciaszczykm:
/assign @floreks @kgroschoff 

Can one of you review it?

xrstf:
I find it a bit sad that because of fixing the XSS vulnerability, this PR also reduces the readability by removing the <strong> tags in the popups. The bold font was pretty nice IMHO.

floreks:
> I find it a bit sad that because of fixing the XSS vulnerability, this PR also reduces the readability by removing the <strong> tags in the popups. The bold font was pretty nice IMHO.

The XSS vulnerability was actually never there as by default Angular applies `sanitize(SecurityContext.HTML, htmlString)` when `innerHTML` attribute is used and the result string is always guaranteed to be secure. Only basic HTML was interpreted there. On the other hand it was possible to provide clickable links or embedded images that could allow other attacks, i.e. phishing.

maciaszczykm:
Yeah, exactly. The previous form would need to be changed. We would need to escape all the names and messages coming from the API and then apply our styling elements like `<strong>`. I have removed them completely because in the designs there is no additional styling. We can go back to it though if that's important. @cschieder, what do you think? Should we use bold resource names in the notification messages?

floreks (approved):
/lgtm

Issue #2881, opened 2020-12-17
UI/UX Fixes


cschieder:
### Text Changes

- [x] Rbac rename column "Name" to "User/Group"
- [x] Rename column "Name" from cluster events table to "Resource ID"

- [x] Add machine deployment button opens the dialog but the dialog says "Add Node Deployment" please fix, we want to use "Machine Deployment" consistently, same with the Edit dialogue

- [x] On Cluster Create Dialoge make final button say "Create Cluster"

- [x] Edit Cluster Dialog rename button to "Save Changes"

### Design/UX Fixes

- [x] Darken serach: border #CAD1D9 / text color #7E868F

- [x] Lighten card shadows (we have 2 shadows stacked, remove the one with 0.25 opacity), reduce corner radius from 5 to 3

- [x] What tooltip text size do we use? Looks pretty small, let's try increasing it by 2 pt/px

**Notifications**

- [x] put number first and remove brackets, like this 2 Notifications

- [x] When only 1 Notification remove the s -> not 1 Notifications but 1 Notification

- [x] Filter icon not vertically centered, can we push it down just a little? 1-2px maybe?

**Edit Node Deployment Dialog**

- [x] alert: "You have changed node specification. All nodes will be recreated." Not an accurate message, as it also appears when only replica number changes, which only adds a replica, rather than recreating the already existing ones? Can we change it to not appear then?

- [x] ^ We might want to change it to info icon rather than alert icon in general or make it yellow instead of red, wdyt?

- [x] Button Text change to "Save Changes"

**Addon**

- [ ] Change styling and behavior to a mini modal (like shown)
![image](https://user-images.githubusercontent.com/49145827/103633194-b00e5880-4f45-11eb-93b1-0056473d34d4.png)

**Connect Cluster**

- [x] sometimes when clicking on connect cluster (most often when it's the first time I try it out in the session) the kubeconfig text field shows up with a noticeable delay, can that be fixed?

- [x] expand Kubeconfig field upon entering data that is longer than the field, only capped by the max height of the whole modal that we set (I think it was around 800-850px) or make the field/dialogue max-height by default

- [x] spacing between kubeconfig field and connect cluster button missing

- [x] remove top margin of "external clusters can only be viewed, not edited." note

kgroschoff:
> Rbac rename column "Name" to "User Email"

In the dialog you can eather specify a email or a group. Not sure if a group has to be an email?!

---
[...]

Issue #2496, opened 2020-07-16
Recreation warning in edit deployment is inconsistent


floreks:
There are some scenarios where it is not triggered or it is triggered and does not disappear once you change options back to default ones.

We have to either rethink the logic or adapt logic to more scenarios. Currently, we are doing a deep object comparison which is quite hard because objects returned by the API do not always provide all properties that are marked as `omitempty`.

PR #2886, opened 2020-12-21
UI/UX fixes

kgroschoff:
Fixes everything possible that was mentioned in the issue. Other points need some clarification/discussion.

Related to https://github.com/kubermatic/dashboard/issues/2881

/

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/node-data/dialog/component.ts
+++ b/src/app/node-data/dialog/component.ts
@@ -163,6 +163,7 @@ export class NodeDataDialogComponent extends BaseFormValidator implements OnInit
   }
 
   private _isRecreationWarningVisible(): boolean {
-    return this.mode === Mode.Edit && !_.isEqual(objectDiff(this._data.initialNodeData, this._output.nodeData), {});
+    const diff = objectDiff(this._data.initialNodeData, this._output.nodeData);
+    return this.mode === Mode.Edit && !_.isEqual(diff, {}) && !(Object.keys(diff).length === 1 && !!diff.count);
   }
 }
--- a/src/app/node-data/dialog/component.ts
+++ b/src/app/node-data/dialog/component.ts
@@ -138,6 +138,17 @@ export class NodeDataDialogComponent extends BaseFormValidator implements OnInit
     this._dialogRef.close(this._output);
   }
 
+  getConfirmButtonText(): string {
+    switch (this.mode) {
+      case Mode.Add:
+        return 'Add Machine Deployment';
+      case Mode.Edit:
+        return 'Save Changes';
+      default:
+        return 'Add Machine Deployment';
+    }
+  }
+
   private _updateNodeData(): void {
     this._output.nodeData = this._nodeDataService.nodeData;
     this.isRecreationWarningVisible = this._isRecreationWarningVisible();
--- a/src/app/node-data/dialog/component.ts
+++ b/src/app/node-data/dialog/component.ts
@@ -174,6 +174,8 @@ export class NodeDataDialogComponent extends BaseFormValidator implements OnInit
   }
 
   private _isRecreationWarningVisible(): boolean {
+    // the icon should not be displayed if only the node replica has changed,
+    // but of course it should be displayed if something else (also) has changed
     const diff = objectDiff(this._data.initialNodeData, this._output.nodeData);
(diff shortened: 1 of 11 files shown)


kgroschoff:
@maciaszczykm / @floreks PTAL

src/app/node-data/dialog/component.ts
```diff
@@ -163,6 +163,7 @@ export class NodeDataDialogComponent extends BaseFormValidator implements OnInit
   }
 
   private _isRecreationWarningVisible(): boolean {
-    return this.mode === Mode.Edit && !_.isEqual(objectDiff(this._data.initialNodeData, this._output.nodeData), {});
+    const diff = objectDiff(this._data.initialNodeData, this._output.nodeData);
+    return this.mode === Mode.Edit && !_.isEqual(diff, {}) && !(Object.keys(diff).length === 1 && !!diff.count);
```
maciaszczykm, line 178: This looks complex. What is the expected result here? If we want to show the warning only if the replica number has changed then why we cannot use `!==` instead of `objectDiff()`?

maciaszczykm: Let's also keep in mind that this was requested in some issue and at the moment we have an issue to fix bug in it: https://github.com/kubermatic/dashboard/issues/2496.

src/assets/css/material/_main.scss
```diff
@@ -508,7 +508,7 @@ mat-chip-list {
 
 // Tooltips.
 .mat-tooltip {
-  font-size: $font-size-caption;
+  font-size: $font-size-caption !important;
```
maciaszczykm, line 511: Can we avoid using `!important` by using more specific selector? I mean something like:

```css
.a {
  .b {
    font-size: 10px;
  }
}
```

src/app/node-data/dialog/component.ts
kgroschoff: we **don't** want to show this warning when replica number has changed. But for all other changes we want to display it

src/app/shared/components/external-cluster-data-dialog/style.scss
```diff
@@ -0,0 +1,18 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+.mat-dialog-content {
+  margin-bottom: 16px;
+}
+
+.km-dialog-warning {
+  margin-top: 0 !important;
```
maciaszczykm, line 17: Can we use specific selector to avoid using `!important`?

maciaszczykm:
Checked it, looks good :) Let's simplify the check. I'd like to avoid using `!important` too.

src/app/node-data/dialog/component.ts
maciaszczykm: Ah, okay. You're right. I'd leave some comment as it's not really easy to understand what is going on here.

kgroschoff: Sure, makes sense. Added it.

maciaszczykm: I also have some reservations about using `!!` on a number property. If the `this._data.initialNodeData.count` would be 0 (it seems to be defaulted at the moment and it works fine) then we may get invalid result. I'd rather check against `undefined` or even if the `Object.keys` contains the `count` property. What do you think?

kgroschoff: You mean something like this?

```
const diff = objectDiff(this._data.initialNodeData, this._output.nodeData);
return this.mode === Mode.Edit && !_.isEqual(diff, {}) && 
       !(Object.keys(diff).length === 1 && !!diff.hasOwnProperty('count'));
```

maciaszczykm: Exactly, feels a bit safer :) BTW, we don't need `!!` on `hasOwnProperty` as it returns boolean anyways.

kgroschoff: Yes, you're right.
Commited it and only changed it a bit because of `Do not access Object.prototype method 'hasOwnProperty' from target object  - no-prototype-builtins` error ;)

maciaszczykm: Good, let's merge.

Issue #2875, opened 2020-12-14
Save status icon does not disappear


maciaszczykm:
**What happened?**:
<img width="718" alt="Zrzut ekranu 2020-12-14 o 08 36 57" src="https://user-images.githubusercontent.com/2823399/102053180-974bd080-3de7-11eb-995d-91800e9b33e0.png">

**What should happen?**:
It should disappear after some time. It should be centered.

**How to reproduce?**:
Change theme in the user settings.

**Environment**
- UI Version: master
- API Version: master
- Domain: localhost
- Others: Chrome on macOS

PR #2889, opened 2020-12-22
Fix theme save status indicator

maciaszczykm:

Fixes https://github.com/kubermatic/dashboard/issues/2875

**Release note**:

```release-note
NONE
```


maciaszczykm:
@floreks It works after change and there are no errors in the console. It seems that `detectChanges()` methods solve the issue and we don't need to change the change detection strategy (that makes the statuses stay visible until something else changes).

PR #2890, opened 2020-12-22
Added keydown.enter to forms

### LINKED ISSUE #2879: Allow keyboard navigation in forms
When I try to create a new project, I intuitively enter the name into the popup:

screenshot-2020-12-15-180741

... and then i press **[ENTER]** on my keyboard and _nothing happens_. :-(

Some oldschool bois like me are used to web forms treating [ENTER] as "submit the form now". Would be cool if KKP dashboard could treat it the same.

kgroschoff:
Wherever possible, added keydown.enter to dialogs to submit the form.

Related to #2879

/

**Release note**:

```release-note
None
```

### DIFF
--- a/src/app/cluster/cluster-details/cluster-delete-confirmation/template.html
+++ b/src/app/cluster/cluster-details/cluster-delete-confirmation/template.html
@@ -12,7 +12,7 @@ limitations under the License.
 -->
 <form [formGroup]="deleteForm"
       id="km-delete-cluster-dialog"
-      (keydown.enter)="deleteCluster()">
+      (submit)="deleteCluster()">
   <km-dialog-title>Delete Cluster</km-dialog-title>
   <mat-dialog-content>
 
@@ -52,7 +52,7 @@ limitations under the License.
   </mat-dialog-content>
   <mat-dialog-actions>
     <button id="km-delete-cluster-dialog-delete-btn"
-            (click)="deleteCluster()"
+            type="submit"
             mat-dialog-close
             mat-flat-button
             [disabled]="!inputNameMatches()">
--- a/src/app/cluster/cluster-details/cluster-delete-confirmation/template.html
+++ b/src/app/cluster/cluster-details/cluster-delete-confirmation/template.html
@@ -10,12 +10,12 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 -->
-<form [formGroup]="deleteForm"
-      id="km-delete-cluster-dialog"
-      (submit)="deleteCluster()">
-  <km-dialog-title>Delete Cluster</km-dialog-title>
-  <mat-dialog-content>
-
+<km-dialog-title>Delete Cluster</km-dialog-title>
+<mat-dialog-content>
+  <form [formGroup]="deleteForm"
+        id="km-delete-cluster-dialog"
+        (submit)="deleteCluster()"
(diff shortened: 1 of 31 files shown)


kgroschoff:
PTAL @maciaszczykm

maciaszczykm:
Works well but there is one more approach that we should consider:

```html
<form [formGroup]="form"
      (submit)="onSubmit()">...</form>
```
```html
<button mat-flat-button
        type="submit"
        [disabled]="!form.valid">Save</button>
```

I think it would be a bit better as we would reference the `onSubmit()` function only once in the template. Doesn't matter if the user clicks the button or press enter key the same function would be called. What do you think?

kgroschoff:
Good idea, will change it

maciaszczykm:
Also, I've found https://github.com/kubermatic/dashboard/issues/2891 issue that is related but not caused by this change. It would be nice to solve it in the near future if we want to advertise allowing keyboard navigation as this is closely related issue.

kgroschoff:
Regarding the issue we could add `onkeypress="return event.keyCode != 13;"` to the input fields from the label form. Not sure if this is the best option, but I know that it is working 😅

maciaszczykm:
@kgroschoff Could be a temporary solution until we will find something better. It's probably something with labels form.

kgroschoff:
@maciaszczykm I added it to this pr for now. Should we nevertheless let the issue open with a comment?

maciaszczykm:
@kgroschoff Yes. I've removed the comment because I am not sure about this but let's keep it. Issue should stay open because it needs to be improved.

/lgtm

maciaszczykm:
By the way, I think this may be a good point for a changelog. It's quite good improvement of UX.

maciaszczykm:
The edit project dialog is broken after the last change:

Zrzut ekranu 2021-01-4 o 12 35 14

Please verify all other modified forms.

maciaszczykm:
@kgroschoff I've fixed the issue with labels in https://github.com/kubermatic/dashboard/pull/2903.

maciaszczykm:
@kgroschoff To associate buttons with forms you can use `form` attribute of the `button` element. Check this out: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button

kgroschoff:
@maciaszczykm finally... PTAL
problem for failing tests: `mat-dialog-close` doen't work in combination with `type="submit"` :ok_woman:

maciaszczykm:
Good to know :) I think we should close the dialog from the handler method if we are using submit button.

kgroschoff:
Yes, we're already closing the dialog within `deleteCluster()` (problem appeared in cluster delete dialog) so this was already kind of duplicated.

maciaszczykm:
I'll review tomorrow.

maciaszczykm:
There are some of the fields that are not covered:
- datacenter dialogs in the admin panel - probably skipped because of the multi-line editor at the end
- connect cluster dialog - as above
- delete service account dialog - on pressing enter the dialog closes, would be good to change this and other confirmation dialogs too
- wizard - it seems complex enough to be additional pull request but perhaps we should do it before we will state that key navigation is enabled in the app

maciaszczykm:
/lgtm

Let's merge and adjust other places from the next pull request.

kgroschoff:
Thanks @maciaszczykm I will have a look and take care in a new pr. Removed the release note for now

maciaszczykm:
If the wizard or any other place will require too much effort then I'd leave it as it is. If there will be any incoming requests about this then we can rethink.

Issue #2902, opened 2021-01-04
Use endpoint to list AWS security groups


zreigz:
Related issue: https://github.com/kubermatic/kubermatic/pull/6331

PR #2909, opened 2021-01-05
Use an endpoint to get AWS Security Group IDs in the wizard for the cluster creation

maciaszczykm:

Fixes https://github.com/kubermatic/dashboard/issues/2902

**Release note**:

```release-note
Use an endpoint to get AWS Security Group IDs in the wizard for the cluster creation.
```


maciaszczykm:
The number of elements is really huge:

Zrzut ekranu 2021-01-8 o 12 59 54

Combobox does not perform well with this big amount of data. Downloading the data takes about 9-10 seconds with local dashboard running with dev cluster.

maciaszczykm:
/hold

To investigate the possible solutions.

cc @zreigz

maciaszczykm:
I've tried to follow the combobox behavior here but what do you think about making the autocompletion load more transparent? I mean what if we would make input always enabled and remove all the 'Loading...` hints, then the user would be able to use the input like before but after a few seconds autocompletions would be loaded.

/cc @kgroschoff @floreks 

Please review it and tell me what do you think about the functionality.

kgroschoff:
Agree, the combobox is super slow with this big amount of data. 
I like the idea of letting the user specify his input even though there is some loading in the background. But for this cases we maybe might need to think about some "validation"? e.g. for some fields we already allow input (I think openstack, because users could have the necessary data input but doesn't have the rights to list all available). Or was this already implemented somewhere (user enteres input that doesn't match any of the options); I don't remember? On the other hand that is the current situation... No "real" validation for inputs 🤷

maciaszczykm:
For the security groups we have validation pattern, so it's quite good. Besides that, cluster creation will fail on the summary page if the security group that was provided doesn't exist.

maciaszczykm:
I've checked with @floreks and he agrees too. I'll adjust it.

maciaszczykm:
@kgroschoff @floreks PTAL

PR #2929, opened 2021-01-12
Pod Node Selector Config

### LINKED ISSUE #2884: Add admission plugins configuration
Add new input (the same as for the label) if PodNodeSelector is enabled during cluster creation or update.
The cluster structure is extended for the new field: `"podNodeSelectorAdmissionPluginConfig":{"clusterDefaultNodeSelector":"env=development"}` in cluster spec. The user should be able to add /edit this map where key is a namespace, value is a label selector.

kgroschoff:
Added option to specify a Pod Node Selector Configuration during cluster creation or cluster edit. It will only be visible, when Pod Node Selector is enabled as Admission Plugin.

Fixes #2884

podNodeSelectorConfig1
podNodeSelectorConfig2
podNodeSelectorConfig3

**Release note**:

```release-note
Add option to specify Pod Node Selector Configuration
```

### DIFF
--- a/src/app/shared/services/cluster.service.ts
+++ b/src/app/shared/services/cluster.service.ts
@@ -94,7 +94,6 @@ export class ClusterService {
   }
 
   set podNodeSelectorAdmissionPluginConfig(config: object) {
-    delete this._cluster.spec.podNodeSelectorAdmissionPluginConfig;
     this._cluster.spec.podNodeSelectorAdmissionPluginConfig = config;
   }
 
--- a/src/app/shared/services/cluster.service.ts
+++ b/src/app/shared/services/cluster.service.ts
@@ -93,6 +93,11 @@ export class ClusterService {
     this._cluster.labels = labels;
   }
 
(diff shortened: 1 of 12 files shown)


kgroschoff:
@maciaszczykm @floreks PTAL
also maybe @zreigz it would be good if you could have a brief look, if this is looking as expected

zreigz:
I will take a look, thanks

zreigz:
looks and works fine :+1:

src/app/shared/services/cluster.service.ts
```diff
@@ -93,6 +93,11 @@ export class ClusterService {
     this._cluster.labels = labels;
   }
 
+  set podNodeSelectorAdmissionPluginConfig(config: object) {
+    delete this._cluster.spec.podNodeSelectorAdmissionPluginConfig;
```
maciaszczykm, line 97: Is this line required? In the next one we are overwriting the same property.

kgroschoff: you're right. removed it.

PR #2936, opened 2021-01-13
Use endpoints to get Azure resource groups, security groups and route tables

maciaszczykm:

Fixes https://github.com/kubermatic/dashboard/issues/2921
Fixes https://github.com/kubermatic/dashboard/issues/2920
Fixes https://github.com/kubermatic/dashboard/issues/2918

**Release note**:

```release-note
Azure resource groups, security groups and route tables will be now loaded from the API to provide autocompletion.
```

### DIFF
--- a/src/app/core/services/wizard/provider/azure.ts
+++ b/src/app/core/services/wizard/provider/azure.ts
@@ -15,6 +15,7 @@ import {NodeProvider} from '../../../../shared/model/NodeProviderConstants';
 import {Provider} from './provider';
 import {
   AzureResourceGroups,
+  AzureRouteTables,
   AzureSecurityGroups,
   AzureSizes,
   AzureZones,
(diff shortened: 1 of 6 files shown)


maciaszczykm:
@kgroschoff @floreks PTAL

maciaszczykm:
/assign @kgroschoff @floreks 

I've also implemented https://github.com/kubermatic/dashboard/issues/2920 here.

kgroschoff (commented):
Working quite good, only noticed that fields aren't cleared correctly. e.g.:
- selected Resource Group `kubernetes-4lbqb2bbz2`
- selected Route Table `kubernetes-4lbqb2bbz2`
- selected Security Group `kubernetes-4lbqb2bbz2`
- changed selection of Resource Group e.g. to `cloud-shell-storage-westeurope`
-> fields aren't cleared.
This somehow works if I only specify the Route Table, but not a Security Group.

Also I'm wondering if it might make sense to display something like "no security groups available" (and so on) if there aren't any (though no idea where. dropdown? hint?)? Because if you haven't selected a Resource Group, you will get a hint, that it has to be specified first. Once you entered one it is gone, which is correct, but I was waiting for the dropdown to appear some time and was wondering if it is a bug 🤔 However if we want to do it, I think this could also be a nice new issue for this dropdown improvements.

maciaszczykm:
>Also I'm wondering if it might make sense to display something like "no security groups available" (and so on) if there aren't any (though no idea where. dropdown? hint?)? Because if you haven't selected a Resource Group, you will get a hint, that it has to be specified first. Once you entered one it is gone, which is correct, but I was waiting for the dropdown to appear some time and was wondering if it is a bug 🤔 However if we want to do it, I think this could also be a nice new issue for this dropdown improvements.

I'd put this in the existing issue as I'm really not sure how it should look like.

maciaszczykm:
@kgroschoff I could reproduce it only when I was typing new names quite fast. Can you please try it again as I have decreased debounce time that I've initially added to avoid too many API requests? In the longer run we will need to take care of it because it is not considered for other endpoints and we receive multiple error notification because of the requests happening when user is still typing.

kgroschoff:
It is clearing the fields correctly now. But I see what happens, when typing too slow/entering something wrong 🤔

maciaszczykm:
@kgroschoff This is the common problem at the moment. Open Hetzner for example and start typing the token manually. You will see multiple error notifications as the requests happen too quickly.

maciaszczykm:
I think this is yet another thing that we should work on rather soon.

kgroschoff:
Yes 👍

maciaszczykm:
Is it good to be merged already or do we want any other changes before?

kgroschoff:
fine for me, let's wait for @floreks

maciaszczykm:
Added https://github.com/kubermatic/dashboard/issues/2987.

PR #2956, opened 2021-01-14
UI/UX Fixes

### LINKED ISSUE #2881: UI/UX Fixes

kgroschoff:
- Rbac column: renamed "Name" to "User/Group"
- Adjusted box-shadows and reduced border-radius from 5 to 3
- Fixed node list, as some missing checks on azure let the list crash & browser console explode for all other providers
 
Related to #2881

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/cluster-details/node-list/template.html
+++ b/src/app/cluster/cluster-details/node-list/template.html
@@ -317,19 +317,19 @@ limitations under the License.
               <div value>{{element.spec.cloud.anexia.vlanID}}</div>
             </km-property>
 
-            <km-property>
+            <km-property *ngIf="element.spec.cloud.azure?.size">
               <div key>Node Size</div>
               <div value>{{element.spec.cloud.azure.size}}</div>
             </km-property>
-            <km-property *ngIf="element.spec.cloud.azure.zones">
+            <km-property *ngIf="element.spec.cloud.azure?.zones">
               <div key>Zones</div>
               <div value>{{element.spec.cloud.azure.zones}}</div>
             </km-property>
-            <km-property *ngIf="element.spec.cloud.azure.imageID">
+            <km-property *ngIf="element.spec.cloud.azure?.imageID">
               <div key>Image ID</div>
               <div value>{{element.spec.cloud.azure.imageID}}</div>
             </km-property>
-            <km-property>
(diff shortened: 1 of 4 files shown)


kgroschoff:
PTAL @maciaszczykm / @floreks

Issue #2969, opened 2021-01-18
Implement OPA table for Constraints in Cluster Details View


kgroschoff:
simple table with headline `Constraints` and columns `Constraint Name`, `Constraint Template`, `Targets` & `Violations`
use endpoint `GET /api/v2/projects/{project_id}/clusters/{cluster_id}/constraints`

only display when `"opaIntegration": {"enabled": true}` (in cluster spec)

PR #2999, opened 2021-01-21
OPA - Add actions for Constraint Templates in Admin Settings

### LINKED ISSUE #2963: Implement 'Add' functionality for OPA to Admin Settings
Dialog to add a new Constraint Template. Should include `Template Name` & `Spec`, which is YAML.
use endpoint `POST /api​/v2​/constrainttemplates`

### LINKED ISSUE #2964: Implement 'Edit' functionality for OPA to Admin Settings
Dialog to edit a Constraint Template. Should include `Template Name` & `Spec`, which is YAML.
use endpoint `PATCH /api​/v2​/constrainttemplates​/{ct_name}`
maybe needed `GET /api​/v2​/constrainttemplates​/{ct_name}`

### LINKED ISSUE #2965: Implement 'Delete' functionality for OPA to Admin Settings
Dialog to delete a Constraint Template. Needs warning, that deleting a Constraint Template will cause all Constraints related to it, to be deleted as well.
use endpoint `DELETE /api​/v2​/constrainttemplates​/{ct_name}`

kgroschoff:
OPA - Add actions (add/edit/delete) for Constraint Templates in Admin Settings

Fixes #2963
Fixes #2964
Fixes #2965

Done 3 issues within this pr, as they are similar & related

**Release note**:

```release-note
NONE
```


src/app/settings/admin/constraint-templates/component.ts
```diff
@@ -0,0 +1,154 @@
...
+import * as _ from 'lodash';
+import {Subject} from 'rxjs';
+import {takeUntil, filter, take, switchMap} from 'rxjs/operators';
+import {ConstraintTemplatesDataDialogComponent} from './constraint-templates-data-dialog/component';
+
+@Component({
+  selector: 'km-constraint-templates',
+  templateUrl: './template.html',
+})
+export class ConstraintTemplatesComponent implements OnInit, OnDestroy {
```
kgroschoff, line 31: Decided to rename the component, as now I think it is telling more about the content.

kgroschoff:
@maciaszczykm / @floreks PTAL

src/app/core/services/api/service.ts
```diff
@@ -344,4 +344,19 @@ export class ApiService {
     const url = `${this._newRestRoot}/constrainttemplates`;
     return this._http.get<ConstraintTemplate[]>(url);
   }
+
+  createConstraintTemplate(template: ConstraintTemplate): Observable<ConstraintTemplate> {
```
floreks, line 348: Can you extract it to a dedicated service? We've started in the past process of refactoring this big API service but it was never finished.

src/app/settings/admin/constraint-templates/component.ts
```diff
@@ -0,0 +1,174 @@
...
+import * as _ from 'lodash';
+import {Subject, timer} from 'rxjs';
+import {takeUntil, filter, take, switchMap} from 'rxjs/operators';
+import {ConstraintTemplatesDataDialogComponent} from './constraint-templates-data-dialog/component';
+
+@Component({
+  selector: 'km-constraint-templates',
+  templateUrl: './template.html',
+})
+export class ConstraintTemplatesComponent implements OnInit, OnChanges, OnDestroy {
```
floreks, line 32: Since this is a list/table I'd rather call it `km-constraint-template-list`.

src/app/settings/admin/constraint-templates/constraint-templates-data-dialog/component.ts
```diff
@@ -0,0 +1,85 @@
...
+  TemplateName = 'name',
+  Spec = 'spec',
+}
+
+@Component({
+  selector: 'km-add-constraint-template-dialog',
+  templateUrl: './template.html',
+  styleUrls: ['./style.scss'],
+})
+export class ConstraintTemplatesDataDialogComponent implements OnInit, OnDestroy {
```
floreks, line 39: I'd follow the selector naming `AddConstraintTemplateDialogComponent`. I guess we could actually drop the `Component` suffix for the dialogs since the dialog is like a special component already.

src/app/settings/admin/constraint-templates/component.ts
```diff
@@ -0,0 +1,174 @@
...
+        this._notificationService.success(`The constraint template ${constraintTemplate.name} was updated`);
+      });
+  }
+
+  delete(constraintTemplate: ConstraintTemplate): void {
+    const dialogConfig: MatDialogConfig = {
+      data: {
+        title: 'Delete Constraint Template',
+        message: `Are you sure you want to delete the constraint template ${constraintTemplate.name}?
+                  <p class="km-confirmation-dialog-delete-warning"> <i class="km-icon-warning"></i> Deleting this constraint template will cause all constraints related to it to be deleted as well.</p>`,
```
floreks, line 159: Please add a specific component with the template for this. Having HTML code in typescript is quite confusing and harder to maintain. Even if it seems simple, it still requires adding styling inside the main style file as there is no way to use a dedicated style file here.

src/app/settings/admin/constraint-templates/component.ts
```diff
@@ -0,0 +1,174 @@
...
+  templateUrl: './template.html',
+})
+export class ConstraintTemplatesComponent implements OnInit, OnChanges, OnDestroy {
+  constraintTemplates: ConstraintTemplate[] = [];
+  dataSource = new MatTableDataSource<ConstraintTemplate>();
+  displayedColumns: string[] = ['templateName', 'actions'];
+  @ViewChild(MatSort, {static: true}) sort: MatSort;
+  @ViewChild(MatPaginator, {static: true}) paginator: MatPaginator;
+  private readonly _refreshTime = 10; // in seconds
+  private _unsubscribe = new Subject<void>();
```
floreks, line 39: Missing `readonly`.

src/app/settings/admin/constraint-templates/constraint-templates-data-dialog/component.ts
kgroschoff: Should I also name it `AddConstraintTemplateDialog` if it is used for Add and Edit? 🤔

src/app/settings/admin/constraint-templates/component.ts
```diff
@@ -0,0 +1,174 @@
...
+  ) {}
+
+  ngOnInit() {
+    this.dataSource.data = this.constraintTemplates;
+    this.dataSource.sort = this.sort;
+    this.dataSource.paginator = this.paginator;
+    this.sort.active = 'templateName';
+    this.sort.direction = 'asc';
+
+    timer(0, this._refreshTime * this._appConfig.getRefreshTimeBase())
```
floreks, line 56: This part introduces a significant memory leak to the application. It will create a new subscription every X seconds without getting rid of the old ones.

src/app/settings/admin/constraint-templates/constraint-templates-data-dialog/component.ts
floreks: Then this is totally confusing since the selector is named `km-add-constraint-template-dialog`. If there are multiple modes handled by this dialog then simply drop Add/Edit prefix from the name.

kgroschoff: oh you're right. I wasn't aware of the `add` in the selector. not sure what happened there 🤔 `km-constraint-template-dialog` sounds good to me?!

src/app/settings/admin/constraint-templates/component.ts
kgroschoff: i will have a look

kgroschoff:
@floreks PTAL

src/app/settings/admin/constraint-templates/component.ts
floreks: Is this solved?

kgroschoff: I think yes, I was trying to implement it within the latest commit. So it depends on your judgment 😅

floreks: It's not. It still creates subscription in subscription every 10 seconds without closing the previous one.

kgroschoff: I've removed this part in https://github.com/kubermatic/dashboard/pull/2999/commits/fef7383923ae5a760bba51435cf876320ef57d9e 🤔 Or am I getting something wrong?

floreks: oh... github didn't refresh for me and I was still seeing the old code. It's partially better. I'd improve further though. Now it would refresh only in case user triggered some action. There is no autorefresh similar to what we do for most resources. You could do it similar to datacenters and add refresh + caching inside the service directly.

kgroschoff: like https://github.com/kubermatic/dashboard/pull/2999/commits/cd4c6044c962747ab59bc867065b855246bc3479 ?

src/app/settings/admin/constraint-templates/component.ts
```diff
@@ -0,0 +1,156 @@
...
+        confirmLabel: 'Add',
+      },
+    };
+
+    this._matDialog
+      .open(ConstraintTemplateDialog, dialogConfig)
+      .afterClosed()
+      .pipe(filter(constraintTemplate => !!constraintTemplate))
+      .pipe(take(1))
+      .subscribe((result: ConstraintTemplate) => {
```
floreks, line 97: nit: For 1 line arrow functions there is no need to add additional brackets. It can be inlined.

floreks: You can use `bind` method here also instead of calling private add method inside arrow function.

floreks: Shouldn't actually the dialog take care of creating/editing here instead of only acting as a proxy for data? @maciaszczykm WDYT?

maciaszczykm: It's hard to say at a first glance. I remember when I was implementing addons I had to take the logic outside of the addons component because it will act different inside cluster page and inside wizard (wizard part is not implemented yet). So, I'd say there might be some cases where we can take the logic outside and it makes sense.

However, I'm not sure if we have any pattern how to do it in general. I mean keeping logic in the components vs in the dialogs. I think in most of the cases we keep it in the dialogs so that might be a good point to stick to that.

src/app/settings/admin/constraint-templates/delete-constraint-template-dialog/component.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import {Component, Input} from '@angular/core';
+import {ConstraintTemplate} from '@shared/entity/opa';
+
+@Component({
+  selector: 'km-delete-constraint-template-dialog',
+  templateUrl: './template.html',
+  styleUrls: ['./style.scss'],
+})
+export class DeleteConstraintTemplateDialog {
```
maciaszczykm, line 20: Do we need specific component for that? Perhaps existing confirmation dialog would be sufficient.

kgroschoff: Since it needs some styling, probably yes.
Note https://github.com/kubermatic/dashboard/pull/2999#discussion_r562639871

src/app/settings/admin/constraint-templates/component.ts
kgroschoff: Alright, changed it in https://github.com/kubermatic/dashboard/pull/2999/commits/ad556c25a3ffe1aa3f96addc434711da4df036fe

src/app/settings/admin/constraint-templates/delete-constraint-template-dialog/component.ts
kgroschoff: you're right. Somehow I didn't saw the warning options

kgroschoff: done

kgroschoff:
https://github.com/kubermatic/dashboard/issues/2969

@maciaszczykm / @floreks PTAL

src/app/settings/admin/constraint-templates/component.ts
```diff
@@ -0,0 +1,139 @@
...
+
+    this._matDialog
+      .open(ConfirmationDialogComponent, dialogConfig)
+      .afterClosed()
+      .pipe(filter(isConfirmed => isConfirmed))
+      .pipe(switchMap(_ => this._opaService.deleteConstraintTemplate(constraintTemplate.name)))
+      .pipe(take(1))
+      .subscribe(_ => {
+        this._notificationService.success(`The constraint template ${constraintTemplate.name} was deleted`);
+        setTimeout(() => this._opaService.refreshConstraintTemplates(), this._defaultTimeout);
```
floreks, line 136: I'd remove this timeout. There is now a good 10s forced refresh so even if API doesn't remove it in time and this refresh won't get it then it will be refreshed after few seconds.

floreks:
Last minor comment and LGTM. I'll shortly test it now.

src/app/settings/admin/constraint-templates/component.ts
kgroschoff: good catch 👍 removed.

floreks:
Tested and it works good.

/lgtm

floreks:
PS. It would be nice to have a hint under template name in create dialog to inform user that name has to be the same as lowercase kind name. It's confusing that I do provide random name and then get an error. It would actually be even better if we could disable this name field and simply read name from the spec kind ourselves.

kgroschoff:
Yes that is a good idea. I will take care of it in one of the next PRs 👍

PR #3008, opened 2021-01-26
OPA - Add Constraints table to cluster details view

### LINKED ISSUE #2966: Implement 'Edit' functionality for OPA to Cluster Details View
Dialog to edit a Constraint. Should include `Constraint Name`, `Constraint Template` (dropdown) & `Spec`, which is YAML.
use endpoint `PATCH /api​/v2​/projects/{project_id}/clusters/{cluster_id}/constraints/{constraint_name}`
Constraint Templates could be loaded via `GET /api​/v2​/constrainttemplates​`

### LINKED ISSUE #2967: Implement 'Delete' functionality for OPA to Cluster Details View
Dialog to delete a Constraint.
use endpoint `DELETE /api​/v2​/projects/{project_id}/clusters/{cluster_id}/constraints/{constraint_name}`

### LINKED ISSUE #2969: Implement OPA table for Constraints in Cluster Details View

### LINKED ISSUE #2989: Implement 'Add' functionality for OPA to Cluster Details View
Dialog to add a new Constraint. Should include `Constraint Name`, `Constraint Template` (dropdown) & `Spec`, which is YAML.
use endpoint `POST /api​/v2​/projects/{project_id}/clusters/{cluster_id}/constraints`
Constraint Templates could be loaded via `GET /api​/v2​/constrainttemplates​`

kgroschoff:
OPA - Add Constraints table to cluster details view

Fixes #2969
Fixes #2989
Fixes #2966
Fixes #2967

/

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/cluster-details/component.ts
+++ b/src/app/cluster/cluster-details/component.ts
@@ -16,6 +16,7 @@ import {AppConfigService} from '@app/config.service';
 import {ApiService} from '@core/services/api/service';
 import {ClusterService} from '@core/services/cluster/service';
 import {DatacenterService} from '@core/services/datacenter/service';
+import {OPAService} from '@core/services/opa/service';
 import {NotificationService} from '@core/services/notification/service';
 import {PathParam} from '@core/services/params/service';
 import {SettingsService} from '@core/services/settings/service';
@@ -29,6 +30,7 @@ import {Health, HealthState} from '@shared/entity/health';
 import {MachineDeployment} from '@shared/entity/machine-deployment';
 import {Member} from '@shared/entity/member';
 import {ClusterMetrics} from '@shared/entity/metrics';
+import {Constraint} from '@shared/entity/opa';
 import {SSHKey} from '@shared/entity/ssh-key';
 import {Config, GroupConfig} from '@shared/model/Config';
 import {NodeProvider} from '@shared/model/NodeProviderConstants';
@@ -67,6 +69,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
   events: Event[] = [];
   addons: Addon[] = [];
   upgrades: MasterVersion[] = [];
+  constraints: Constraint[] = [];
   private _unsubscribe: Subject<any> = new Subject();
   private _user: Member;
   private _currentGroupConfig: GroupConfig;
@@ -82,6 +85,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
     private readonly _userService: UserService,
     private readonly _api: ApiService,
     private readonly _notificationService: NotificationService,
+    private readonly _opaService: OPAService,
     readonly settings: SettingsService
   ) {}
 
@@ -136,8 +140,9 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
                     this._clusterService.nodes(this.projectID, this.cluster.id),
                     this._api.getMachineDeployments(this.cluster.id, this.projectID),
                     this._clusterService.metrics(this.projectID, this.cluster.id),
+                    this._opaService.constraints(this.projectID, this.cluster.id),
                   ]
-                : [of([]), of([]), of([]), of([])]
+                : [of([]), of([]), of([]), of([]), of([])]
             );
 
           return combineLatest(reload$);
@@ -145,18 +150,20 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
       )
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(
-        ([upgrades, addons, nodes, machineDeployments, metrics]: [
+        ([upgrades, addons, nodes, machineDeployments, metrics, constraints]: [
           MasterVersion[],
           Addon[],
           Node[],
           MachineDeployment[],
-          ClusterMetrics
+          ClusterMetrics,
+          Constraint[]
         ]) => {
           this.addons = addons;
           this.nodes = nodes;
           this.machineDeployments = machineDeployments;
           this.metrics = metrics;
           this.upgrades = _.isEmpty(upgrades) ? [] : upgrades;
+          this.constraints = constraints;
         },
         error => {
           const errorCodeNotFound = 404;
@@ -344,7 +351,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
   }
 
   isOPAEnabled(): boolean {
-    return !!this.cluster.spec.opaIntegration && !!this.cluster.spec.opaIntegration.enabled;
+    return this.cluster.spec.opaIntegration.enabled;
   }
 
   ngOnDestroy(): void {
--- a/src/app/cluster/cluster-details/component.ts
+++ b/src/app/cluster/cluster-details/component.ts
@@ -351,7 +351,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
   }
 
   isOPAEnabled(): boolean {
-    return this.cluster.spec.opaIntegration.enabled;
+    return !!this.cluster.spec.opaIntegration && this.cluster.spec.opaIntegration.enabled;
   }
 
   ngOnDestroy(): void {
--- a/src/app/cluster/cluster-details/component.ts
+++ b/src/app/cluster/cluster-details/component.ts
@@ -343,6 +343,10 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
     return MemberUtils.hasPermission(this._user, this._currentGroupConfig, 'rbac', Permission.View);
   }
 
+  isOPAEnabled(): boolean {
+    return !!this.cluster.spec.opaIntegration && !!this.cluster.spec.opaIntegration.enabled;
+  }
+
   ngOnDestroy(): void {
     this._unsubscribe.next();
     this._unsubscribe.complete();
--- a/src/app/cluster/cluster-details/constraints/component.ts
+++ b/src/app/cluster/cluster-details/constraints/component.ts
@@ -9,12 +9,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Component, Input, OnDestroy, OnInit} from '@angular/core';
+import {ChangeDetectorRef, Component, Input, OnDestroy, OnInit, ViewChild} from '@angular/core';
+import {MatPaginator} from '@angular/material/paginator';
+import {MatSort} from '@angular/material/sort';
 import {MatTableDataSource} from '@angular/material/table';
+import {OPAService} from '@core/services/opa/service';
+import {UserService} from '@core/services/user/service';
 import {Cluster} from '@shared/entity/cluster';
 import {Constraint} from '@shared/entity/opa';
+import {UserSettings} from '@shared/entity/settings';
 import * as _ from 'lodash';
 import {Subject} from 'rxjs';
+import {takeUntil} from 'rxjs/operators';
 
 @Component({
   selector: 'km-constraints-list',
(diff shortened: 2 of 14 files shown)


kgroschoff:
@maciaszczykm / @floreks I think this one is ready for a review. PTAL

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,183 @@
...
+import {Cluster} from '@shared/entity/cluster';
+import {Constraint, Kind, Violation} from '@shared/entity/opa';
+import {UserSettings} from '@shared/entity/settings';
+import * as _ from 'lodash';
+import {Subject} from 'rxjs';
+import {take, takeUntil, filter, switchMap} from 'rxjs/operators';
+import {Mode, ConstraintDialog} from './constraint-dialog/component';
+
+@Component({
+  selector: 'km-constraints-list',
```
floreks, line 30: nit: s/km-constraints-list/km-constraint-list

src/app/cluster/cluster-details/component.ts
```diff
@@ -343,6 +343,10 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
     return MemberUtils.hasPermission(this._user, this._currentGroupConfig, 'rbac', Permission.View);
   }
 
+  isOPAEnabled(): boolean {
+    return !!this.cluster.spec.opaIntegration && !!this.cluster.spec.opaIntegration.enabled;
```
floreks, line 347: No need for double negation of a non-optional boolean: `this.cluster.spec.opaIntegration.enabled`

src/app/core/services/opa/service.ts
```diff
@@ -61,4 +63,36 @@ export class OPAService {
     const url = `${this._newRestRoot}/constrainttemplates/${ctName}`;
     return this._http.delete(url);
   }
+
+  constraints(projectId: string, clusterId: string): Observable<Constraint[]> {
+    this._constraints$ = merge(this._constraintsRefresh$, this._refreshTimer$)
```
maciaszczykm, line 68: `this._constraints$` is assigned each time the function is invoked, why?

maciaszczykm: If this should have the role of a cache then it should look like admin settings getter for example. If not, then `this._constraints$` is not required and you can return directly.

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,183 @@
...
+      data: {
+        title: 'Add Constraint',
+        projectId: this.projectID,
+        clusterId: this.cluster.id,
+        mode: Mode.Add,
+        confirmLabel: 'Add',
+      },
+    };
+
+    this._matDialog
```
floreks, line 133: This can be actually simplified to just `this._matDialog.open(ConstraintDialog, dialogConfig) since we do not do anything based on the dialog result. I am not sure why it is so common in our code if we are not using the subscribe at all.

src/app/core/services/opa/service.ts
floreks: I'd add cache here similar to what I've done in https://github.com/kubermatic/dashboard/pull/3033/files. This will make sure that we do not create multiple observables for the same projectID/clusterID pair.

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,183 @@
...
+        mode: Mode.Edit,
+        confirmLabel: 'Edit',
+      },
+    };
+
+    this._matDialog
+      .open(ConstraintDialog, dialogConfig)
+      .afterClosed()
+      .pipe(take(1))
+      .subscribe(_ => {});
```
maciaszczykm, line 157: It looks like this can be removed.

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
```diff
@@ -0,0 +1,137 @@
...
+  private _initConstraintSpecEditor(): void {
+    if (this.data.mode === Mode.Edit) {
+      const spec = this.data.constraint.spec;
+      if (!_.isEmpty(spec)) {
+        this.spec = dump(spec);
+      }
+    }
+  }
+
+  private _getSpec(): any {
```
floreks, line 132: This method is quite confusing. I'd add here at least some basic types to better identify the loaded object. We should be able to expect some kind of object schema loaded from the spec and cast it to a typed object instead of any.

Also, what is even more confusing here is that the object is loaded, then there is no check but something is assigned to a field of this `any` object and then it is being checked if it is empty. Is there an actual reason for this?

src/app/cluster/cluster-details/constraints/violation-details/component.ts
```diff
@@ -0,0 +1,39 @@
...
+  styleUrls: ['./style.scss'],
+})
+export class ViolationDetailsComponent implements OnInit {
+  @Input() violations: Violation[];
+  displayedColumns: string[] = ['name', 'enforcement', 'kind', 'message'];
+  dataSource = new MatTableDataSource<Violation>();
+  @ViewChild(MatSort, {static: true}) sort: MatSort;
+
+  ngOnInit(): void {
+    this.dataSource.data = this.violations;
```
floreks, line 29: There is no check here if violations are defined, but there is in `getDataSource` why?

Easiest way to default value here is to do `this.violations = this.violations || []`;

src/app/core/services/opa/service.ts
kgroschoff: Good catch. Wanted to cache the data, but forgot it half of the way. 🙈

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
kgroschoff: Yeah I also don't like it. I tried a bit to improve it in https://github.com/kubermatic/dashboard/pull/3008/commits/60bfb67a8f077edad78a0cb4ce3df7f804f39073, but I'm not sure if we can do it better.... Any input is welcome.

Also I'm wondering if we should use the `Constraint Template` from the form that we have or from the spec (if specified). I guess I think it is saver to use the form, as this field only displays existing templates. 🤔

src/app/cluster/cluster-details/component.ts
```diff
@@ -344,7 +351,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
   }
 
   isOPAEnabled(): boolean {
-    return !!this.cluster.spec.opaIntegration && !!this.cluster.spec.opaIntegration.enabled;
+    return this.cluster.spec.opaIntegration.enabled;
```
maciaszczykm, line 354: This actually can produce null pointer errors if the `opaIntegration` will be undefined.

maciaszczykm: I think @floreks wanted to make it look like this: `!!this.cluster.spec.opaIntegration && this.cluster.spec.opaIntegration.enabled`.

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
kgroschoff: @floreks ping ;)

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,171 @@
...
+  set matPaginator(mp: MatPaginator) {
+    this.paginator = mp;
+    if (this.paginator && this.settings) {
+      this.paginator.pageSize = this.settings.itemsPerPage;
+    }
+    this.dataSource.paginator = this.paginator;
+    this._cdr.detectChanges();
+  }
+
+  private _unsubscribe = new Subject<void>();
```
floreks, line 56: Should be also `readonly`

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,171 @@
...
+  ) {}
+
+  ngOnInit(): void {
+    this.dataSource.data = this.constraints;
+    this.dataSource.sort = this.sort;
+    this.dataSource.paginator = this.paginator;
+    this.sort.active = 'constraintName';
+    this.sort.direction = 'asc';
+
+    this._userService.currentUserSettings.pipe(takeUntil(this._unsubscribe)).subscribe(settings => {
```
floreks, line 73: `{}` brackets not needed here.

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,171 @@
...
+    return _.isEmpty(data) && this.isClusterRunning;
+  }
+
+  toggleDetails(element: Constraint): void {
+    this.isShowDetails[element.name] = !this.isShowDetails[element.name];
+  }
+
+  listMatch(element: Kind[]): string {
+    const matches = [];
+    element.forEach(x => matches.push(x.kinds));
```
floreks, line 110: This is basically a `map` array method.

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,171 @@
...
+    this.sort.active = 'constraintName';
+    this.sort.direction = 'asc';
+
+    this._userService.currentUserSettings.pipe(takeUntil(this._unsubscribe)).subscribe(settings => {
+      this.settings = settings;
+    });
+  }
+
+  ngOnChanges(): void {
+    this.dataSource.data = this.constraints;
```
floreks, line 79: I'd actually add a check here if constraints have changed instead of always assigning them. `ngOnChanges` has an argument that provides what has changed.

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,171 @@
...
+
+  hasNoData(data: Constraint[]): boolean {
+    return _.isEmpty(data) && this.isClusterRunning;
+  }
+
+  toggleDetails(element: Constraint): void {
+    this.isShowDetails[element.name] = !this.isShowDetails[element.name];
+  }
+
+  listMatch(element: Kind[]): string {
```
floreks, line 108: It seems more like `toKindDisplayName`, `listMatch` is not very informative.

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
```diff
@@ -0,0 +1,141 @@
...
+})
+export class ConstraintDialog implements OnInit, OnDestroy {
+  readonly Controls = Controls;
+  form: FormGroup;
+  spec = '';
+  constraintTemplates: ConstraintTemplate[] = [];
+  private readonly _unsubscribe = new Subject<void>();
+
+  constructor(
+    public _matDialogRef: MatDialogRef<ConstraintDialog>,
```
floreks, line 58: The name suggests it is private, but the modifier says otherwise? :confused:

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
```diff
@@ -0,0 +1,141 @@
...
+    private readonly _opaService: OPAService,
+    private readonly _notificationService: NotificationService,
+    @Inject(MAT_DIALOG_DATA) public data: ConstraintDialogConfig
```
floreks, line 61: If it's not used in the template directly, it can be `private readonly`

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
```diff
@@ -0,0 +1,141 @@
...
+  ) {}
+
+  ngOnInit(): void {
+    this.form = new FormGroup({
```
floreks, line 65: Use `FormBuilder` please.

kgroschoff:
@floreks done

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,182 @@
...
+    return _.isEmpty(data) && this.isClusterRunning;
+  }
+
+  toggleDetails(element: Constraint): void {
+    this.isShowDetails[element.name] = !this.isShowDetails[element.name];
+  }
+
+  displayKindNames(element: Kind[]): string {
+    const kinds = [];
+    element.map(x => kinds.push(x.kinds));
```
floreks, line 121: `map` is maps values. It does not work the same as `forEach`.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
floreks: What about this one?

floreks: I see it is used in the template. No need to change then.

src/app/cluster/cluster-details/constraints/constraint-dialog/component.ts
```diff
@@ -0,0 +1,142 @@
...
+    }
+  }
+
+  private _getSpec(): ConstraintSpec {
+    let spec = new ConstraintSpec();
+    const raw = load(this.spec) as ConstraintSpec;
+    if (!_.isEmpty(raw)) {
+      spec = raw;
+    }
+    spec['constraintType'] = this.form.get(Controls.ConstraintTemplate).value;
```
floreks, line 139: Looks better. Since `spec` is now of `ConstraintSpec` type there is no need to use bracket notation to get the value. Take it from the object instead.

src/app/cluster/cluster-details/constraints/constraint-dialog/template.html
```diff
@@ -0,0 +1,52 @@
...
+        Template Name is <strong>required</strong>.
+      </mat-error>
+    </mat-form-field>
+
+    <mat-form-field fxFlex>
+      <mat-label>Constraint Template</mat-label>
+      <mat-select [formControlName]="Controls.ConstraintTemplate"
+                  disableOptionCentering>
+        <mat-option *ngFor="let constraintTemplate of constraintTemplates"
+                    [value]="constraintTemplate.spec.crd.spec.names.kind">
```
floreks, line 35: I'd add here `?` to be sure that it won't throw undefined error.

kgroschoff:
@floreks done

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -0,0 +1,181 @@
...
+  hasNoData(data: Constraint[]): boolean {
+    return _.isEmpty(data) && this.isClusterRunning;
+  }
+
+  toggleDetails(element: Constraint): void {
+    this.isShowDetails[element.name] = !this.isShowDetails[element.name];
+  }
+
+  displayKindNames(element: Kind[]): string {
+    const kinds = element.map(x => x.kinds);
```
floreks, line 120: One last thing left.
`return element.map(x => x.kinds).join(', ');

kgroschoff: 👍

PR #3053, opened 2021-02-08
Filter external networks in openstack wizard

phiphi282:
External networks cannot be used as cluster networks and will result in
a failure during cluster creation when selected.
Filter external networks in the cluster creation wizard for openstack to
prevent users from selecting those options.

Fixes #

**Release note**:

```release-note
Filter external openstack networks during cluster creation
```


floreks:
You can never use an external network for the cluster creation or is it only true for some installations?

phiphi282:
I am pretty sure that we can never use external networks.
During cluster creation openstack looks specifically for internal networks and fail with an error message similar to:
```
400 invalid cluster... failed to get network ext-net with external=false
```
where ext-net is our external network

kgroschoff:
@floreks I remember that we used to have this filter in the old wizard. I guess it has been forgotten during wizard-relaunch.

kgroschoff (commented):
/approve

PR #3063, opened 2021-02-10
Validate name during cluster creation with the same pattern as on edit

stoewer:
This PR changes the pattern used to validate cluster names during cluster creation to the [same one](https://github.com/kubermatic/dashboard/blob/b062244f8645f14f9cdb015fa63d236141fde013/src/app/cluster/cluster-details/edit-cluster/component.ts#L67) that is used when editing the cluster.

**Which issue(s) this PR fixes** 
This fixes a bug where it is possible to create a cluster that can't be edited later.

```release-note
NONE
```


xrstf:
This will presumably fix #3066.

kgroschoff:
Hello @stoewer , thanks for your PR. 
It is looking good so far, but lets wait for the tests first.

/test all

kgroschoff (commented):
/lgtm
/approve

PR #3100, opened 2021-02-22
Remove OpenShift support

maciaszczykm:
**What this PR does / why we need it**: We need to remove OpenShift support from the UI before https://github.com/kubermatic/kubermatic/pull/6539 can be merged.

Fixes #

**Release note**:

```release-note
Removed OpenShift support.
```


maciaszczykm:
@kubermatic/sig-ui @cschieder What should be done with cluster types that are displayed in the app? From now on the only available type is Kubernetes since OpenShift is removed. What do you think?

floreks:
I guess it should work the same as when you only select Kubernetes in the admin settings. No distribution information as Kubernetes is the default and only one.

maciaszczykm:
@floreks Yeah, that's what I did here. I was wondering about the lists and other places where we show Kubernetes logo.

floreks:
I wouldn't worry about other places right now. It looks fine IMO.

maciaszczykm:
/assign @kgroschoff @floreks 

Please take a look and do some manual checks if possible.

kgroschoff:
Not sure, if we already talked to customers about removing openshift (probably?), but I'm wondering if we might need some hint at the openshift cluster, that the specified type isn't available anymore? This of course only is needed if openshift ever was used somewhere 🤔 
Otherwise I couldn't find anything unexpected so far

os
os2

floreks:
> Not sure, if we already talked to customers about removing openshift (probably?), but I'm wondering if we might need some hint at the openshift cluster, that the specified type isn't available anymore? This of course only is needed if openshift ever was used somewhere 
> Apart from that I couldn't find anything unexpected so far
> 
> os
> os2

@shaase-ctrl could you provide some information here? How should we proceed?

maciaszczykm:
I guess they will be removed as the OpenShift cluster type was removed from the backend in https://github.com/kubermatic/kubermatic/pull/6539.

shaase-ctrl:
We will communicate this with the release. Afaik there is no production grade environment affected.

PR #3158, opened 2021-03-12
Added network to Hetzner

### LINKED ISSUE #3124: Allow configuring the network for Hetzner clusters
Starting with KKP 2.17 (at latest), we want to have much better Hetzner support. For this it's necessary to configure the Hetzner _Network_ for each usercluster. 

The dashboard should allow users to change the network during cluster creation. The network is defaulted from the datacenter, so it's perfectly fine for users to leave it empty and just use the defaults.

https://github.com/kubermatic/kubermatic/pull/6588 already adds basic support for a new `network` field to the API, which is optional (again, as it's defaulted to the datacenter config anyway).

**Important**: While machines can be in multiple networks, the Hetzner CCM requires a single network. This is why the PR mentioned above adds support for exactly one network per cluster. When translating cluster information to MachineDeployments, _then_ we turn the single network into a 1-element array (see https://github.com/kubermatic/kubermatic/pull/6588/files#diff-54336f4c05a19427373cedb64f1d1d5a6672c95946c5a256b91448487201bfd0 and https://github.com/kubermatic/kubermatic/pull/6588/files#diff-6a146064ec9accb1750f2b79473e00b79ed77fb11df11e94c050895cdd864ed9)

**Important 2**: This might all be backported to 2.16, so any changes to the dashboard should keep that in mind.

kgroschoff:
Added network to Hetzner

Fixes #3124

FYI: Networks should only be available during cluster creation, but not during cluster edit. The machine deployment needs the exact same network as the cluster, therefor it is copied over.

**Release note**:

```release-note
Added network to Hetzner
```

### DIFF
--- a/src/app/node-data/basic/provider/hetzner/component.ts
+++ b/src/app/node-data/basic/provider/hetzner/component.ts
@@ -87,9 +87,9 @@ export class HetznerBasicNodeDataComponent extends BaseFormValidator implements 
 
     this._nodeDataService.nodeData = this._getNodeData();
 
-    this._clusterService.clusterChanges.pipe(takeUntil(this._unsubscribe)).subscribe(_ => {
-      this._nodeDataService.nodeData = this._getNodeData();
-    });
+    this._clusterService.clusterChanges
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(_ => (this._nodeDataService.nodeData = this._getNodeData()));
   }
 
   ngOnDestroy(): void {
--- a/src/app/node-data/basic/provider/hetzner/component.ts
+++ b/src/app/node-data/basic/provider/hetzner/component.ts
@@ -144,6 +144,7 @@ export class HetznerBasicNodeDataComponent extends BaseFormValidator implements 
       spec: {
         cloud: {
           hetzner: {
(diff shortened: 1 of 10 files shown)


kgroschoff:
@maciaszczykm / @floreks PTAL

src/app/node-data/basic/provider/hetzner/component.ts
```diff
@@ -80,6 +84,12 @@ export class HetznerBasicNodeDataComponent extends BaseFormValidator implements
     });
 
     this._typesObservable.pipe(takeUntil(this._unsubscribe)).subscribe(this._setDefaultType.bind(this));
+
+    this._nodeDataService.nodeData = this._getNodeData();
+
+    this._clusterService.clusterChanges.pipe(takeUntil(this._unsubscribe)).subscribe(_ => {
+      this._nodeDataService.nodeData = this._getNodeData();
```
maciaszczykm, line 91: nit: Let's change it to one-liner.

maciaszczykm (commented):
I've 2 minor comments. Besides that looks good.

kgroschoff:
@maciaszczykm done :)

maciaszczykm:
Good :)

/lgtm

PR #3160, opened 2021-03-12
OPA- fix typo

kgroschoff:
@lsviben noticed a typo in constraint dialog. Instead of `Template Name` it should be `Constraint Name`. 👍 

**Release note**:

```release-note
NONE
```


kgroschoff:
Just a super small one, PTAL @maciaszczykm

PR #3170, opened 2021-03-16
Improve dropdown styling

### LINKED ISSUE #3129: Minor Dropdown changes and fixes
### Changes

- Increase size of arrow icon to 14px
- Add hover animation like already present on the x to the arrow as well (only if hovered over the arrow/arrow area specifically)
image
https://material-ui.com/components/autocomplete/ (behavior example)
- Decrease size of hover circle of both arrow and x to 30px 

### Bugs/Broken
Admin Settings -> Presets (for some reason way larger x icon)
image

kgroschoff:
Changes and fixes related to dropdowns, such as hovers and sizes

Fixes #3129

/

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/settings/admin/presets/template.html
+++ b/src/app/settings/admin/presets/template.html
@@ -35,7 +35,7 @@ limitations under the License.
                     mat-icon-button
                     class="km-clear-input-btn"
                     aria-label="Clear">
-              <i class="km-icon-close"></i>
+              <i class="km-icon-mask km-icon-remove"></i>
             </button>
           </div>
         </mat-select-trigger>
@@ -64,7 +64,7 @@ limitations under the License.
                     mat-icon-button
                     class="km-clear-input-btn"
                     aria-label="Clear">
-              <i class="km-icon-close"></i>
+              <i class="km-icon-mask km-icon-remove"></i>
             </button>
           </div>
         </mat-select-trigger>
--- a/src/app/settings/admin/presets/template.html
+++ b/src/app/settings/admin/presets/template.html
@@ -17,7 +17,7 @@ limitations under the License.
        fxLayoutGap="16px">
     <div class="filter-title">Filter by</div>
     <mat-form-field fxFlex
-                    class="km-dropdown-with-suffix km-filter-with-suffix">
+                    class="km-dropdown-with-suffix">
       <mat-label>Provider</mat-label>
       <mat-select disableOptionCentering
                   class="km-select-ellipsis"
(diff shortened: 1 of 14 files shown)


kgroschoff:
@maciaszczykm / @floreks PTAL

src/app/settings/admin/presets/template.html
```diff
@@ -35,7 +35,7 @@
                     mat-icon-button
                     class="km-clear-input-btn"
                     aria-label="Clear">
-              <i class="km-icon-close"></i>
```
floreks, line 38: This might be a good chance to improve overall styling and use `matSuffix` instead of absolute positioning.

https://material.angular.io/components/form-field/overview#form-field-appearance-variants

kgroschoff: Probably spend a bit too much time to figure out some css rules to let everything nearly look the same as before...
This causes a new class `km-dropdown-with-suffix` if a dropdown should have a suffix and additionally if it is a filter, which has reduced with `km-filter-with-suffix` class.

Finally noticed following part of the documentation 😅 
`We do not recommend using text prefix and suffixes in the fill and outline variants because the label and input do not have the same alignment. It is therefore impossible to align the prefix or suffix in a way that looks good when compared with both the label and input text.`

However, I commited my changes, would be great if you could have a look. (I'm also fine with reverting it if needed ;) )

floreks:
One general remark:
- https://material-ui.com/components/autocomplete/ - This is a custom implementation of material design guidelines from a third party company. I wouldn't make it our base to follow as we are using the official angular material framework from Google and i.e. their behavior for arrow icon for select component is different than in our framework: https://material.angular.io/components/select/examples

I'll actually push a small update so you can take a look.

floreks:
PTAL

cschieder:
Hi, my two cents, Google actually puts a hover on their arrow in their dropdowns too
image

And it makes sense to me, if the x has a hover the arrow should have one too. I would rather do as they do and do what sounds logical rather than strictly following what the components would do by default.

floreks:
Google sites are actually not the best example as depending on the page behavior of components varies highly. Google calendar for example does not display any indicator for their people search autocomplete input.

More important thing is that the example you have provided shows a different feature. The dropdown icon in there opens a completely new component with unique options. When you type autocomplete is triggered but when you click on the icon you get this
image

It makes sense then highlight it as a unique button that offers a different feature.

cschieder:
I guess you have a point.. now that I think about it the hover would have to be triggered by the whole field and that might be awkward as well, let's stick with no hover on the arrow then

kgroschoff (commented):
👍 

@floreks okay let's merge it. Though not sure if I can lgtm my own PR. However, lets try 🤔 
/lgtm
/retest

floreks:
Actually, please test it again as I wasn't focusing on the spacing etc. but rather on the behavior. Make sure that positioning, sizes, spacing fits the issue description.

kgroschoff:
I already did ;)

PR #3174, opened 2021-03-17
OPA - Improvements

kgroschoff:
Added OPA prefix to all components
Fixed position of checkbox in cluster step in wizard
Added paginators

/

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/cluster-details/constraints/violation-details/component.ts
+++ b/src/app/cluster/cluster-details/constraints/violation-details/component.ts
@@ -10,9 +10,11 @@
 // limitations under the License.
 
 import {Component, Input, OnInit, ViewChild} from '@angular/core';
+import {MatPaginator} from '@angular/material/paginator';
 import {MatSort} from '@angular/material/sort';
 import {MatTableDataSource} from '@angular/material/table';
 import {Violation} from '@shared/entity/opa';
+import {UserSettings} from '@shared/entity/settings';
 
 @Component({
   selector: 'km-violation-details-list',
@@ -20,15 +22,28 @@ import {Violation} from '@shared/entity/opa';
   styleUrls: ['./style.scss'],
 })
 export class ViolationDetailsComponent implements OnInit {
-  @Input() violations: Violation[];
+  @Input() violations: Violation[] = [];
+  @Input() settings: UserSettings;
   displayedColumns: string[] = ['name', 'enforcement', 'kind', 'message'];
   dataSource = new MatTableDataSource<Violation>();
   @ViewChild(MatSort, {static: true}) sort: MatSort;
+  @ViewChild(MatPaginator, {static: true}) paginator: MatPaginator;
 
   ngOnInit(): void {
     this.dataSource.data = this.violations || [];
     this.dataSource.sort = this.sort;
+    this.dataSource.paginator = this.paginator;
+    this.paginator.pageSize = this.settings.itemsPerPage;
     this.sort.active = 'name';
     this.sort.direction = 'asc';
+  }
+
+  isPaginatorVisible(): boolean {
+    return (
+      this.violations &&
+      this.violations.length > 0 &&
+      this.paginator &&
+      this.violations.length > this.paginator.pageSize
+    );
   }
 }
--- a/src/app/cluster/cluster-details/constraints/violation-details/component.ts
+++ b/src/app/cluster/cluster-details/constraints/violation-details/component.ts
@@ -10,9 +10,10 @@
 // limitations under the License.
 
 import {Component, Input, OnInit, ViewChild} from '@angular/core';
-import {MatPaginator} from '@angular/material/paginator';
+import {MatPaginator, PageEvent} from '@angular/material/paginator';
 import {MatSort} from '@angular/material/sort';
 import {MatTableDataSource} from '@angular/material/table';
+import {OPAService} from '@core/services/opa/service';
 import {Violation} from '@shared/entity/opa';
 import {UserSettings} from '@shared/entity/settings';
 
@@ -29,11 +30,14 @@ export class ViolationDetailsComponent implements OnInit {
   @ViewChild(MatSort, {static: true}) sort: MatSort;
   @ViewChild(MatPaginator, {static: true}) paginator: MatPaginator;
 
+  constructor(private readonly _opaService: OPAService) {}
+
   ngOnInit(): void {
     this.dataSource.data = this.violations || [];
     this.dataSource.sort = this.sort;
     this.dataSource.paginator = this.paginator;
     this.paginator.pageSize = this.settings.itemsPerPage;
+    this.paginator.pageIndex = this._opaService.violationPageIndex;
     this.sort.active = 'name';
     this.sort.direction = 'asc';
   }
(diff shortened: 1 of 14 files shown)


src/app/cluster/cluster-details/constraints/violation-details/component.ts
```diff
@@ -10,25 +10,40 @@
...
+  @Input() settings: UserSettings;
   displayedColumns: string[] = ['name', 'enforcement', 'kind', 'message'];
   dataSource = new MatTableDataSource<Violation>();
   @ViewChild(MatSort, {static: true}) sort: MatSort;
+  @ViewChild(MatPaginator, {static: true}) paginator: MatPaginator;
 
   ngOnInit(): void {
     this.dataSource.data = this.violations || [];
     this.dataSource.sort = this.sort;
+    this.dataSource.paginator = this.paginator;
```
kgroschoff, line 35: Paginator for Violations is a bit harder. It is there but the page will be resetted every few minutes due to "refresh" (e.g. you are on page 2, but it refreshes to 1 fairly quick). I've tried some other implementations, but nothing seems to work in the way it should be, so I just committed "simple" way for now. 
Anyhow, do you have an idea how to solve this page/refresh issue?

maciaszczykm: I'd say that if the number of pages increases then there is no problem and we stay on the same page. If the number is decreased and the page that we were on will no longer exist then we should go to the last possible page. Does it make sense? Can it be implemented in that way?

kgroschoff: Sorry I somehow missed your comment. 
I'm not entirely sure if we're thinking the same 🤔 

Screenshot is a bit small but it should be enough to see the problem. The toggled area is refreshed every few seconds and therefor the paginator is set to 0 again and again. 
pag

I spend some additional time on it and tried a few different implementations and maybe I have done something wrong, but I didn't manage to get everything up and running. :/

kgroschoff: @maciaszczykm @floreks

maciaszczykm:
There seems to be some problem with adding constraint without YAML. After doing it I got the `The constraint X was created` message but there is nothing in the table even after refresh.

src/app/cluster/cluster-details/constraints/violation-details/component.ts
maciaszczykm: The current issue is that the `ViolationDetailsComponent` gets initialized each 5-10 seconds (add some log in the `ngOnInit` to see this). I'll check if I have any idea how to fix that.

maciaszczykm: Sorry, I didn't have enough time yesterday. The workaround might be moving the pagination init somewhere else however I am not sure if it will work. Could you try that?

maciaszczykm: You could also keep somewhere page index and set it during validation.

maciaszczykm: Sorry, I didn't have enough time yesterday. The workaround might be moving the pagination init somewhere else or keeping the page index somewhere else and setting it during the init. I'm not sure if any of these will work though, could you try that?

kgroschoff: saving the index seems to work. PTAL @maciaszczykm

maciaszczykm: Great, I'll review today :)

maciaszczykm:
> There seems to be some problem with adding constraint without YAML. After doing it I got the The constraint X was created message but there is nothing in the table even after refresh.

cc @lsviben

maciaszczykm:
@kgroschoff You have to save the page index into a map because having global page index is impossible when you can open more than one violations tables. Check the example below:

Zrzut ekranu 2021-04-6 o 11 59 12

Zrzut ekranu 2021-04-6 o 11 58 41

I didn't click anything in the table above. The index of the second page was loaded from the service.

lsviben:
> > There seems to be some problem with adding constraint without YAML. After doing it I got the The constraint X was created message but there is nothing in the table even after refresh.
> 
> cc @lsviben

This is because we are creating the Kubermatic Constraints in the Seed cluster through the API, and they are synced as Gatekeeper Constraints to the User cluster with a controller, and sometimes that doesnt work out if the Constraint is not valid.

Also, the GET and LIST endpoints were getting the Constraints directly from the user cluster and transforming them to the API version because we wanted to get the Status fields which are filled out by Gatekeeper on the User cluster Gatekeeper Constraints. But, this was changed in a recent PR (https://github.com/kubermatic/kubermatic/pull/6800) and now the GET works fine, even if the Constraints were not synced properly.

And they are not synced usually because there are some errors in the constraints, but that fails silently in the constraint controller. There is a ticket to add validation to the Kubermatic Constraints(https://github.com/kubermatic/kubermatic/issues/6767) that Ill be working on soon, so we should have less of such situations. But in the meantime, at least users will be able to GET their Constraints and update them if they see something is wrong.

maciaszczykm:
@lsviben Okay, thanks for letting me know :)

kgroschoff:
@maciaszczykm True, that was stupid. Changed it 👍

src/app/core/services/opa/service.ts
```diff
@@ -144,11 +143,13 @@ export class OPAService {
...
-    return this._violationPageIndex;
+  getViolationPageIndex(projectId: string, clusterId: string, violationIndex: number): number {
+    const id = `${projectId}-${clusterId}-${violationIndex}`;
+    return this._violationPageIndex.get(id);
   }
 
-  saveViolationPageIndex(index: number): void {
-    this._violationPageIndex = index;
+  saveViolationPageIndex(projectId: string, clusterId: string, violationIndex: number, pageIndex: number): void {
+    const id = `${projectId}-${clusterId}-${violationIndex}`;
```
maciaszczykm, line 152: Are we sure that the `violationIndex` will not change if the number of violations will change? I think it would be better to use constraint name (is it unique?) and the violation name instead and keep the project and cluster IDs.

kgroschoff: It isn't connected to the violation count, but you're right, I think using the constraint name is better. Checked and I can't create a constraint with the same name in this cluster. We cannot really use a violation name, because which one do you would choose out of the list (we need the paginator for this list)? I think project-cluster-constraint is as explicit as we could be. Will change it :)

kgroschoff: @maciaszczykm done

maciaszczykm: Yeah, not violations but constraint count. Violation name is also out of scope.

maciaszczykm:
/lgtm

Quick fixes 👍🏻

kgroschoff:
Sorry @maciaszczykm I forgot a unit test. Can you please lgtm again?

maciaszczykm:
/hold

@kgroschoff you have to rebase as the PR that might improve the situation was just merged.

kgroschoff:
@maciaszczykm also in here: done

PR #3194, opened 2021-03-23
can configure autoscaling options

furkhat:

It adds autoscaling options in machine deploymens, so users can set min/max replicas

Partially Fixes https://github.com/kubermatic/kubermatic/issues/6507
Related API changes https://github.com/kubermatic/kubermatic/pull/6596

Right now I display autoscaler controls to any cluster. Should we add a condition for displaying them? Like if it certain type of cluster or check if autoscaler is deployed. What approach would you suggest?

**Release note**:

```release-note
* Add autoscaler options in machine deployment
```

### DIFF
--- a/src/app/node-data/component.ts
+++ b/src/app/node-data/component.ts
@@ -46,6 +46,9 @@ enum Controls {
   ProviderBasic = 'providerBasic',
   ProviderExtended = 'providerExtended',
   Kubelet = 'kubelet',
+  UseAutoscaling = 'useAutoscaling',
+  MaxReplicas = 'maxReplicas',
+  MinReplicas = 'minReplicas',
 }
 
 @Component({
@@ -114,6 +117,13 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       [Controls.RhsmOfflineToken]: this._builder.control(''),
       [Controls.ProviderBasic]: this._builder.control(''),
       [Controls.ProviderExtended]: this._builder.control(''),
+      [Controls.UseAutoscaling]: this._builder.control(this._nodeDataService.nodeData.maxReplicas),
+      [Controls.MinReplicas]: this._builder.control(this._nodeDataService.nodeData.minReplicas || 0, [
+        Validators.min(0),
+      ]),
+      [Controls.MaxReplicas]: this._builder.control(this._nodeDataService.nodeData.maxReplicas || 1, [
+        Validators.min(1),
+      ]),
     });
 
     if (this.isDialogView()) {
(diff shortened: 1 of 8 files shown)


src/app/node-data/template.html
```diff
@@ -47,21 +47,55 @@
       </mat-error>
     </mat-form-field>
 
-    <mat-form-field>
+    <mat-checkbox [formControlName]="Controls.UseAutoscaling"
+                  class="km-use-autoscaling"
+                  fxFlex>
+      Use horizontal node autoscaling
```
maciaszczykm, line 53: We are using title case for checkboxes, please change.

src/app/node-data/template.html
```diff
@@ -47,21 +47,55 @@
...
       <mat-label>Replicas</mat-label>
-      <input [formControlName]="Controls.Count"
-             id="km-node-count-input"
-             matInput
-             type="number"
-             min="0"
-             autocomplete="off"
-             required>
+      <input [formControlName]="
+                    Controls.Count"
```
maciaszczykm, line 59: This line break should not happen.

src/app/node-data/template.html
```diff
@@ -47,21 +47,55 @@
...
+                    *ngIf="form.get(Controls.UseAutoscaling).value">
+      <mat-label>Max Replicas*</mat-label>
+      <input [formControlName]="Controls.MaxReplicas"
+             id="km-node-max-replicas-input"
+             matInput
+             type="number"
+             min="0"
+             min="1"
+             autocomplete="off">
+      <mat-error *ngIf="form.get(Controls.MaxReplicas).hasError('min')">Must be greater or equal 1</mat-error>
```
maciaszczykm, line 93: Add `.` at the end.

src/app/node-data/template.html
```diff
@@ -47,21 +47,55 @@
...
+             id="km-node-min-replicas-input"
+             matInput
+             type="number"
+             min="0"
+             autocomplete="off">
+    </mat-form-field>
+
+    <mat-form-field fxFlex
+                    *ngIf="form.get(Controls.UseAutoscaling).value">
+      <mat-label>Max Replicas*</mat-label>
```
maciaszczykm, line 85: `*` should be coming from `required` directive. Is it intended, if yes then why?

src/app/node-data/style.scss
```diff
@@ -17,6 +17,10 @@ div.km-os-options {
   margin: 0 0 10px;
 }
 
+mat-checkbox.km-use-autoscaling {
+  margin-bottom: 2.5 * $baseline-grid;
```
maciaszczykm, line 21: Please merge the latest master. We are no longer using baseline grid.

src/app/node-data/component.ts
```diff
@@ -156,9 +162,44 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
...
+        const minR = this.form.get(Controls.MinReplicas).value || 0;
+        if (minR < 0) {
+          this.form.get(Controls.MinReplicas).setErrors({min: true});
+        } else {
+          this.form.get(Controls.MinReplicas).setErrors(null);
+        }
+
+        const maxR = this.form.get(Controls.MaxReplicas).value;
+        if (maxR < minR) {
+          this.form.get(Controls.MaxReplicas).setErrors({lessThanMin: true});
```
maciaszczykm, line 189: It would be great to create a validator for that instead of doing it from this subscription.

src/app/node-data/component.ts
```diff
@@ -156,9 +162,44 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
...
+        const useAutoscaling = !!this.form.get(Controls.UseAutoscaling).value;
+        if (!useAutoscaling) {
+          this.form.get(Controls.MinReplicas).setErrors(null);
+          this.form.get(Controls.MaxReplicas).setErrors(null);
+          return;
+        }
+
+        const minR = this.form.get(Controls.MinReplicas).value || 0;
+        if (minR < 0) {
+          this.form.get(Controls.MinReplicas).setErrors({min: true});
```
maciaszczykm, line 182: It's in the template. You can also use ` Validators.min(0)`.

maciaszczykm: In general I think you can set validators for both fields and then keep them defaulted like now. If user will not select the autoscaling checkbox then the form will still be valid. You can also make them disabled/enabled depending on if the checkbox is marked or not.

maciaszczykm:
@furkhat I've added some additional comments. The tests are not passing for some time now. Please take a look on them. In case on any questions ping me here or on Slack.

src/app/node-data/component.ts
furkhat: I need current values of min replicas and max replicas here. How do I write a validator that can access both values?

furkhat:
@maciaszczykm updated.

This implementation always displays "enable node autoscaling" checkbox even if cluster has no autoscaler deployed. I think we should may it conditional, how would you suggest check that cluster has autoscaler deployed?

src/app/node-data/style.scss
```diff
@@ -17,6 +17,10 @@ div.km-os-options {
   margin: 0 0 10px;
 }
 
+mat-checkbox.km-use-autoscaling {
```
floreks, line 20: s/km-use-autoscaling/use-autoscaling

We tend to use `km` prefix only for global classes.

floreks:
> @maciaszczykm updated.
> 
> This implementation always displays "enable node autoscaling" checkbox even if cluster has no autoscaler deployed. I think we should may it conditional, how would you suggest check that cluster has autoscaler deployed?

@zreigz does API expose any information if the cluster can use autoscaler option?

src/app/node-data/component.ts
maciaszczykm: Ok, then let's not make it too complex and leave it as it is.

ahmedwaleedmalik:
Closing this PR since the API part never made it through and this PR needs heavy refactoring.

PR #3198, opened 2021-03-28
Improve CI, update Docker images

xrstf:
**What this PR does / why we need it**:
This PR started out as an update to the Docker images, but then slowly evolved into the "test why CI is so slow" PR. From there we

* replaced port-forwardings with NodePort services in the hopes to make them more stable (still thinking the ECONNRESET errors during the test were an actual issue, but it turns out, they are totally normal in low amounts)
* upload Cypress artifacts to our Minio, so we can see the videos and screenshots

**Release note**:
```release-note
NONE
```

PR #3212, opened 2021-04-01
Fix default operating system selection

### LINKED ISSUE #3199: Default operating system is not selected for Anexia
**What happened?**: There is no default operating system for Anexia.

**What should happen?**: The only one available option should be selected.

**How to reproduce?**:
Zrzut ekranu 2021-03-29 o 08 22 27

**Environment**
- UI Version: master
- API Version: master
- Domain: dev
- Others:

kgroschoff:
Fix default operating system selection

Fixes #3199

/

**Release note**:

```release-note
NONE
```

### DIFF
--- a/src/app/node-data/component.ts
+++ b/src/app/node-data/component.ts
@@ -322,20 +322,20 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       }
     }
 
-    return OperatingSystem.Ubuntu;
+    return null;
   }
 
   private _getDefaultOS(): OperatingSystem {
-    if (this.isOperatingSystemSupported(this._nodeDataService.operatingSystem)) {
-      return this._nodeDataService.operatingSystem;
+    if (this.isProvider(NodeProvider.ANEXIA)) {
+      return OperatingSystem.Flatcar;
     }
 
-    if (this._datacenterSpec) {
+    if (this._datacenterSpec && !!this._getDefaultSystemTemplate(this.provider)) {
       return this._getDefaultSystemTemplate(this.provider);
     }
 
-    if (this.isProvider(NodeProvider.ANEXIA)) {
-      return OperatingSystem.Flatcar;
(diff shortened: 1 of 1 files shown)


floreks:
Fix for this is already bundled with the e2e updates: https://github.com/kubermatic/dashboard/pull/3198/files#diff-63e050f0feb764530b887203b5248c3441e4f5d9aebd3716616df2ad4526f83e

kgroschoff:
I see. 😅  But actually the default system template also is broken
```
if (this._datacenterSpec) {
    return this._getDefaultSystemTemplate(this.provider);
}
```
since this part always returns Ubuntu first
```
if (this.isOperatingSystemSupported(this._nodeDataService.operatingSystem)) {
      return this._nodeDataService.operatingSystem;
}
```

Should I then just remove the anexia part or do you want to take over the change to your PR? WDYT?

floreks:
This part is actually only used when editing since `this._nodeDataService.operatingSystem` is empty when using wizard. It is filled with initial data when editing node.
```typescript
if (this.isOperatingSystemSupported(this._nodeDataService.operatingSystem)) {
      return this._nodeDataService.operatingSystem;
}
```

The actual code that runs inside the wizard would be then
```typescript
    if (this._datacenterSpec) {
      return this._getDefaultSystemTemplate(this.provider);
    }

    if (this.isProvider(NodeProvider.ANEXIA)) {
      return OperatingSystem.Flatcar;
    }

    return OperatingSystem.Ubuntu;
```

Getting system from datacenterspec should be the last step as it only supports vsphere right now and even though other datacenters can have this spec defined, they do not provide OS template that we could map. That's why I have simply switched those 2.

kgroschoff:
Okay. But hasn't the anexia part then come first? Because `_getDefaultSystemTemplate` currently always is returning Ubuntu too, but Anexia only has Flatcar 🤔

floreks:
It does come first in my fix. I have pasted the current code with the issue present.

kgroschoff:
sorry then I just close here

PR #3232, opened 2021-04-09
Add pre-release/release procedures

floreks:
With @maciaszczykm we have prepared a pre-release/release procedures for the Kubarmatic Dashboard repository. It consists of new issue/pull request templates as well as the document that in general describes both parts.

Closes to #3231

**Release note**:

```release-note
NONE
```


docs/manuals/release.md
```diff
@@ -0,0 +1,104 @@
...
+  - Machine deployment deletion
+  - Addon creation & deletion
+  - RBAC creation & deletion
+  - SSH Key management from cluster details 
+- External cluster connection
+
+## Release Procedure
+List of steps to follow after performing manual tests:
+
+Create a new issue based on the [release issue template]().
```
floreks, line 95: Will be updated after merge when the link to the issue template will become available.

.github/ISSUE_TEMPLATE/pre-release.md
```diff
@@ -0,0 +1,98 @@
+---
+name: Pre-Release
+title: [vX.XX] Pre-Release Procedure
+assignees: @floreks @maciaszczykm @kgroschoff
+about: The list of steps that should be executed prior to the release step list 
+labels: sig/ui sig/release
+---
+
+## Pre-Release Procedure
```
maciaszczykm, line 9: I'd make some note here that it should be kept in sync with the document below.

.github/ISSUE_TEMPLATE/pre-release.md
```diff
@@ -0,0 +1,98 @@
...
+List of steps to execute before proceeding with the [release procedure](docs/manuals/release.md#release-procedure). It mainly consists of
+manual tests and visual checks of the Kubermatic Dashboard.
```
kgroschoff, line 11: Should we also test project, service accounts & members (add/edit/delete) + display all projects as an admin?
Or is this something we rely on the e2e tests?

floreks: > I mean we can't test the provider stuff without a project, but just to complete the list...?

We've dropped projects from manual testing for exactly this reason. There is no need to remember about that as it will always be tested anyway. As for the SA & Members, we've got e2e tests for them that already check basically everything that is possible to check there. There is no need to do this manually then. Wizard & clusters are much more complex and e2e do not always catch everything there.

kgroschoff: Okay understand. 👍 
But still - maybe a quick check if the `display all projects` function is working? I think we don't test it somewhere else, do we? 🤔

floreks: Good idea with the `display all projects` switch. We do not have a test for that. I'll add a check.

floreks: Done

docs/manuals/release.md
```diff
@@ -0,0 +1,5 @@
+## Release Procedure
+1. Create a new issue based on the [pre-release issue template](). It mainly consists of manual tests and visual checks of the Kubermatic Dashboard.
```
floreks, line 2: Links will be updated after merge.

PR #3245, opened 2021-04-13
Added Alibaba VSwitch

kgroschoff:
Added option to select Alibaba VSwitch from list

### Release note

```release-note
NONE
```


kgroschoff:
PTAL @maciaszczykm / @floreks

floreks:
After discussing with @zreigz and @maciaszczykm it would be safer to use input with autocomplete rather than combobox. As we do not know exactly how alibaba works it might be better to give users the option to manually provide vSwitch value if needed.

Other than that it lgtm.

kgroschoff:
@floreks done, PTAL

Issue #3207, opened 2021-03-30
UI: vSphere VM template not filled if only one OS templage is specified


toschneck:
**What happened?**:
If at the seed configuration is only one OS template image is specified, the UI doesn't show up the image template value
![image.png](https://images.zenhubusercontent.com/5b6454d031c5fe7463837d8b/939cb271-a626-4be2-ad82-43aa05a7ce52)

`seed.yaml`:
```
apiVersion: kubermatic.k8s.io/v1
kind: Seed
metadata:
  name: tobi-seed-run-2
  namespace: kubermatic
spec:
  # Optional: Country of the seed as ISO-3166 two-letter code, e.g. DE or UK.
  # For informational purposes in the Kubermatic dashboard only.
  country: DE
  # Datacenters contains a map of the possible datacenters (DCs) in this seed.
  # Each DC must have a globally unique identifier (i.e. names must be unique
  # across all seeds).
  datacenters:
########### Vsphere
    tobi-run-2-vsphere:
      # Optional: Country of the seed as ISO-3166 two-letter code, e.g. DE or UK.
      # For informational purposes in the Kubermatic dashboard only.
      country: DE
      location: Hamburg (onsite-seed)
      spec:
        vsphere:
          allow_insecure: false
          cluster: cl-1
          datacenter: dc-1
          datastore: exsi-nas
          endpoint: https://vcenter.loodse.io
          root_path: /dc-1/vm/kubermatic
          templates:
            flatcar: flatcar_2021-02.ova
```

**What should happen?**:
`flatcar_2021-02.ova` should be filled out

**How to reproduce?**:
Create/Configure a vSphere seed datacenter with only one OS template specified, like in the above example + create a user cluster on this seed.

**Environment**
- UI Version: `weekly-2021-12`
- API Version: `weekly-2021-12`
- Domain: `run-2`
- Others:

PR #3268, opened 2021-04-20
Set default template based on the available operating systems

maciaszczykm:
Previously Ubuntu was used as default even if it wasn't configured for the chosen datacenter. Right now, the datacenter configuration is taken under consideration.

Fixes https://github.com/kubermatic/dashboard/issues/3207

### Release note

```release-note
Set default VM template based on the available operating systems.
```


src/app/node-data/basic/provider/vsphere/component.ts
```diff
@@ -121,7 +121,8 @@ export class VSphereBasicNodeDataComponent extends BaseFormValidator implements
     }
   }
 
-  private _setDefaultTemplate(os: OperatingSystem): void {
+  private _setDefaultTemplate(os: OperatingSystem = undefined): void {
```
floreks, line 124: If this parameter is no longer used, I'd remove it.

maciaszczykm: It's used in line 108.

PR #3281, opened 2021-04-23
Fix cachin' too stronk

### LINKED ISSUE #3137: Caches too stronk
**What happened?**:
I upgraded my KKP from 2.15.6 to 2.16.3.

**What should happen?**:
The footer in the dashboard should say "v2.16.3". It didn't say anything anymore.

After some investigation it turns out that my browser was really, really keen on caching the version.json, which seems to have changed formats between 2.15 and 2.16.

Cachebusting would be great.

**How to reproduce?**:
- Install 2.15.6.
- Upgrade
- Refresh browser window

**Environment**
- UI Version: 2.16.3
- API Version: 2.16.3
- Domain: Captain
- Others: Firefox

floreks:
I have added `config` and `assets/config` directories to the no-cache zone to make sure all files there will be always taken from the server.

Fixes #3137

### Release note

```release-note
NONE
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -70,10 +70,10 @@ func main() {
 }
 
 func handler(w http.ResponseWriter, r *http.Request) {
-	// disable caching for the main HTML page
-	if r.URL.Path == "/" || r.URL.Path == "/index.html" {
+	// disable caching for the root(index.html) and all config files
+	if r.URL.Path == "/" || isCacheDisabled(r.URL.Path) {
 		// Forces caches to submit the request to the origin server for validation before releasing a cached copy.
-		w.Header().Add("Cache-Control", "no-cache")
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -70,10 +70,10 @@ func main() {
 }
 
 func handler(w http.ResponseWriter, r *http.Request) {
-	// disable caching for the main HTML page
-	if r.URL.Path == "/" || r.URL.Path == "/index.html" {
+	// disable caching for the root(index.html) and all config files
+	if r.URL.Path == "/" || isCacheDisabled(r.URL.Path) {
```
maciaszczykm, line 74: Can we add "/" to the method too?

Issue #3290, opened 2021-04-26
[v2.17] Release Procedure


floreks:
## Release Procedure
List of steps to follow after finishing the pre-release procedure.

- [x] Version bump in the following files:
  - [x] `package.json`
  - [x] `package-lock.json`
  - [x] `Makefile`
- [x] Update changelog file (`src/assets/config/changelog.json`) with the latest release notes

PR #3291, opened 2021-04-26
Bump version to v2.17 and update changelog

floreks:
Together with @maciaszczykm we have prepared an initial Dashboard changelog based on the https://github.com/kubermatic/kubermatic/pull/6922.

Additionally, changelog categories were updated as the previous ones were not usable.

#### Current changelog (no worries it is scrollable :slightly_smiling_face:)
screencapture-localhost-8000-projects-2021-04-26-15_46_30

/cherry-pick release/v2.17

Closes https://github.com/kubermatic/dashboard/issues/3290

Please take a look and see if we should mention any other changes. We only scanned the PR names and mentioned what we thought sounded important.

@cschieder as we have updated the design a bit, could you also take a look?

### Release note

```release-note
NONE
```


src/assets/config/changelog.json
```diff
@@ -1 +1,93 @@
-{}
+{
+  "externalChangelogURL": "https://github.com/kubermatic/kubermatic/blob/release/v2.17/CHANGELOG.md#kubermatic-217",
```
floreks, line 2: Is it safe to use this link upfront as a link to our full changelog?

src/assets/config/changelog.json
```diff
@@ -1 +1,93 @@
...
+  "entries": [
+    {
+      "category": "added",
+      "description": "Added addon for Multus CNI",
+      "links": [
+        {
+          "url": "https://docs.kubermatic.com/kubermatic/master/guides/addons/multus/",
```
floreks, line 9: Remember to update `master` to `2.17` before merge.

floreks:
Until all comments are resolved let's hold this one.
/hold

cschieder:
1. I think I would maybe change the colors of the points, like this in the respective order, I made `removed` the yellow color because I think it's a rather rare point or one that warrants attention? like removed support of os. If that impression is wrong please tell me.
```
#00A3D9
#00CCA7
#1B658F
#F7B500
```

2. maybe we can add like 5-10 px space below links? so that it becomes clearer that they belong to the point above and theres less potential confusion that they could belong to the point below

3. last idea would be to add an external link icon to the full changelog button (as far as I know we link to github here) and move the button to the right, grouped with the got it one

(old content, don't mind it)
image

floreks:
@cschieder FYI new categories palette as there will be now 7 categories. Feel free to update as needed. 
```scss
  changelog-action-required: #f63535,
  changelog-added: #1dc67f,
  changelog-removed: #00a3d9,
  changelog-fixed: #ff6701,
  changelog-changed: #008c99,
  changelog-deprecated: #6b979b,
  changelog-security: #fea82f,
```

Maybe some minor comments from my side:
- I'd keep `action-required`, `security` and `fixed` (as the bug fix) in the red/orange/yellow palette as they are quite important and should induce the feeling of something potentially dangerous
- `deprecated` is usually gray'ish
- `added`, `removed` and `changed` can be basically anything. I'd just keep those 3 in a similar color palette.

floreks:
> last idea would be to add an external link icon to the full changelog button (as far as I know we link to github here) and move the button to the right, grouped with the got it one

Good idea with the external icon. I'd not push it to the right as clicking on `Got it` sends some data to the server and there is a spinner while waiting for the response from the server. If they would both be on the right then the spinner would push the changelog link to the left.

floreks:
@xrstf please override as this is a flake. The other e2e passed.

cschieder:
Thanks for giving me the full list :) I don't like the idea of the user feeling danger when they look at our changelog so I tried to tone down the warning colors while still grabbing attention, additionally I would prefer exclusively using red for errors, or things that need immediate attention before they explode. 

(above: current state, below: my suggested colors) 
image


Actually, to ensure the attention without the red, we can work more with the order of the points, maybe like this? 
1.	Action Required
2.	Security
3.	Fixed
4.	Added
5.	Changed
6.	Deprecated
7.	Removed

For the spinner, would it be possible (and not too much effort) to make it push the full changelog button to the side with a mini animation? That way we could have them both grouped to the right side. My reasoning for wanting it on the right side is that we usually keep "negative" actions to the left side, i. e. deleting or going back, and "positive" actions that lead you further on the right side, so I would like to continue that pattern. 

And last but not least, we don't have an x to close the dialogue, I think we should add one, behavior should either be the same as clicking "got it" or just close the dialoge, but make it come back the next times until the user clicks "got it"

Issue #3271, opened 2021-04-22
Card-View sometimes broken in KKP-ee v2.15


cedi:
**What happened?**:
Sometimes the card-view shows some UI glitches like this one:
![2021-04-22-123542_701x248_scrot](https://user-images.githubusercontent.com/1952599/115701044-a3ccd080-a367-11eb-8e42-e16648d08460.png)

**What should happen?**:
Carda are either resized or the cluster name is truncated so it does not line-break anymore

**Environment**
- UI Version: 2.15.11

maciaszczykm:
We have to fix the method that is handling longer project names. It's causing the issue. I'd like to avoid resizing cards because all of them should have the same size.

PR #3297, opened 2021-04-28
Fix project card headers

maciaszczykm:
Previously longer names with `-` in it were automatically wrapped and this was causing some visual glitches. I've set `nowrap` and switched to CSS method as it is more stable (not all letters have the same width and this sets limit on div size not letter count).

Zrzut ekranu 2021-04-28 o 08 55 39

Fixes https://github.com/kubermatic/dashboard/issues/3271.

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/project/template.html
+++ b/src/app/project/template.html
@@ -209,7 +209,10 @@ limitations under the License.
            class="km-health-state km-project-cards-status"
            fxFlexAlign=" center"></i>
         <mat-card-title [matTooltip]="getProjectTooltip(project.name)"
-                        [attr.id]="'km-project-name-' + project.name">{{getName(project.name)}}</mat-card-title>
+                        [attr.id]="'km-project-name-' + project.name"
+                        class="card-header">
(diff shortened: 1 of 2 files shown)


kgroschoff (commented):
Great!
/lgtm

src/app/project/template.html
```diff
@@ -209,7 +209,10 @@
            class="km-health-state km-project-cards-status"
            fxFlexAlign=" center"></i>
         <mat-card-title [matTooltip]="getProjectTooltip(project.name)"
-                        [attr.id]="'km-project-name-' + project.name">{{getName(project.name)}}</mat-card-title>
+                        [attr.id]="'km-project-name-' + project.name"
+                        class="card-header">
```
floreks, line 213: I do not see this `card-header` class definition. Where is it?

maciaszczykm: It's a leftover. PTAL

PR #3300, opened 2021-04-28
Shorten file names part II

### LINKED ISSUE #2592: Use shorter file names

kgroschoff:
Shorten file names

Fixes #2592

### Release note

```release-note
NONE
```


kgroschoff:
@maciaszczykm now I think I really covered everything. PTAL

We though have a few special cases, where we have a few files that have a longer name, e.g. `google-analytics.service.ts`. They don't have their own folders, so we eather let them stay as they are _or_ we create folders (e.g. every service in core/service has its own folder) _or_ getting rid of the `.service` endings in at least the `/services` folders. WDYT? I don't have a strong opinion on that tbh.
- app (2)
- cluster/services (2)
- shared/services (2)
- testing/components (3)
- testing/services (14)
- testing/fake-data (23, ending with .fake)

Also I will open a new issue to reduce the length of an import. We have a few places that could use the shorter `@[...]` form.

maciaszczykm:
I'm not sure. For me it's okay to have names like `cluster.service.ts` and `eol.service.ts` if we group these in the `service` directory. What I would like in that case is to keep them organized in directories and not shattered everywhere. Having directories is also okay, but having directory for 1 file seems a bit too much. No strong opinion here.

cc @floreks WDYT?

floreks:
What we do inside the k8s dashboard seems reasonable here. For services, interceptors, etc. we have dedicated folders and then we simply use very short names based on the service name, i.e.
Service -> `AuthenticationService`
File -> `authentication.ts`
Path -> `common/services/authentication.ts`

It is then clear that this is service since it is placed inside `services` directory and it relates to authentication. I'd avoid creating a directory per service. In case we need a separate module to gather some specific services we can then create a separate directory. For example, we could have a module for all the global services (`services/global/...`).

kgroschoff:
Okay, sounds good.

I guess since we have the most services placed in core/services, we will also move the two from cluster/services and the two from shared/services in there? 
We only might to take a look how to seperate/combine them nicely, because e.g. in shared/services we're having a ClusterService, that mostly handles wizard stuff, but we also have a ClusterService in core/services that is mostly used in cluster-details :dizzy_face:

Also I guess we then get rid of most (or all? 🤔) of the folders in core/services?
core-services

maciaszczykm:
> Also I guess we then get rid of most (or all? 🤔) of the folders in core/services?

+1

> We only might to take a look how to seperate/combine them nicely, because e.g. in shared/services we're having a ClusterService, that mostly handles wizard stuff, but we also have a ClusterService in core/services that is mostly used in cluster-details 😵

I'd like to have only one cluster service anyways, it's confusing to have two of them. If keeping them makes sense then at least one should be renamed IMO.

kgroschoff:
Following are left.
I'm not entirely sure if I should keep the folders for `auth` & `changelog` and if yes do we rename them also to e.g. `auth/auth.ts` (would be at least consistent to the other files without folders)? Also I'm wondering if the `global/module.ts` is stored correctly in `core/services` 🤔 WDYT? @maciaszczykm / @floreks 
cs

maciaszczykm:
`listener.ts` contains `ChangelogManagerService` and `service.ts` contains `ChangelogService`. We can place them in the same directory as other services using their names, i.e. `changelog-manager.ts` and `changelog.ts`.

maciaszczykm:
@kgroschoff Remember to remove the [WIP] if it can be reviewed already.

kgroschoff:
@maciaszczykm I guess we can merge if everything is okay

/test pre-dashboard-test-e2e

maciaszczykm:
@kgroschoff You need to resolve conflict and then we can merge. I'd rather avoid waiting here because it will cause more conflicts. Also, does this close the issue?

kgroschoff:
@maciaszczykm I haven't seen the conflict befor, but I rebased now 👍 
Should be done afterwards, so I guess yes.

PR #3353, opened 2021-05-11
Add e2e tests for admin settings related to wizard

### LINKED ISSUE #3303: [E2E] Admin Settings Without Cluster Creation
Includes:
- OPA Integration
- Machine Deployment (replicas)
- Resource Quota

floreks:
- Updated cypress to `v7.3.0`
- Removed deprecated `cy.server` and `cy.route` usage in favor of `cy.intercept`
- Removed v2.17 changelog from master
- Added e2e tests for admin settings:
  - opa integration
  - machine deployment replicas
  - resource quota

Closes #3303

### Release note

```release-note
NONE
```


src/app/wizard/step/cluster/component.ts
```diff
@@ -186,7 +186,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
 
   private _enforce(control: Controls, isEnforced: boolean): void {
     if (isEnforced) {
-      this.form.get(control).setValue(true);
+      this.form.get(control).setValue(this._settings.opaOptions.enabled);
```
kgroschoff, line 189: You might not be done yet, but just wanted to mention (because it came into my mind):
The `_enforce` function is also used in line 124 `this._enforce(Controls.AuditLogging, dc.spec.enforceAuditLogging);`.
AuditLogging is specified via the dc and should definitely be an enabled + disabled.

Also please change the OPA Enforced behavior in the edit cluster dialog.

floreks: In such a case (enforced by dc), all such settings should follow this behavior. I'll then leave it as is and adjust the test itself. We should update other settings to follow this logic.

floreks: I have created a follow up issue #3355

Issue #3355, opened 2021-05-12
Enforce settings should disable and check related checkboxes


floreks:
Related to https://github.com/kubermatic/dashboard/pull/3353#discussion_r630258487

Since it is desired that dc enforce settings mark and disable those checkboxes we should be consistent and apply this behavior everywhere.

PR #3392, opened 2021-05-31
Align enforce setting behavior

maciaszczykm:
It changes the current behavior of enforce OPA checkbox. Right now it will enforce setting that is selected in the first checkbox (enable by default). I've changed "enforce" label to "enforce defaults" to make it more clear.

Fixes https://github.com/kubermatic/dashboard/issues/3355.

### Release note
```release-note
NONE
```


maciaszczykm:
/assign @kgroschoff 

Please check if the applied logic is fine.

kgroschoff:
So just for clarification:
Admin Settings - OPA enabled + enforced = wizard checkbox is checked and disabled
Admin Settings - OPA not enabled + enforced = wizard checkbox is only disabled, but not checked
Right?

I think the original idea of your issue was to check + disable every time when `enforce` is true, because we have some `enforceXYZ` things on dc level. But tbh I think the `Enforce Default` is a great idea and makes it more clear, that it only works in combination with the `Enable` option.

Code looks good.

floreks:
I think that on the DC level we do not actually have cleanup nor OPA settings. If that is true the idea was to make `enforce` manipulate enable/disable status of the field and `enable by default` would check/uncheck it.

maciaszczykm:
> So just for clarification:
Admin Settings - OPA enabled + enforced = wizard checkbox is checked and disabled
Admin Settings - OPA not enabled + enforced = wizard checkbox is only disabled, but not checked
Right?

True.

maciaszczykm:
I was checking it yesterday and there is nothing OPA or cluster-cleanup related in the DC settings.

If that's okay then PTAL @kgroschoff.

maciaszczykm:
I'll have to adjust tests.

kgroschoff:
Yes there is nothing for OPA on dc level.

I'm waiting for the tests then ;)

maciaszczykm:
/hold

To take a look on tests.

maciaszczykm:
@kgroschoff @floreks Can one of you recheck?

PR #3393, opened 2021-05-31
Migrate to angular 12

### LINKED ISSUE #3364: Migrate to angular 12
As new angular version was recently released we could start the migration. It could be also a good opportunity to start using ES modules instead of the CJS that we use right now.

floreks:
- using angular 12 now  (couple of new features described [here](https://blog.angular.io/angular-v12-is-now-available-32ed51fbfd49) including new shortened operator)
- moved `npm test` from pre-commit to pre-push event
- added npm strict engine check (supported node versions 14-15)
- updated most of our dependencies including RxJS
- most of our files now use ESM file format (basic info [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules))

Closes #3364

### Release note

```release-note
NONE
```


floreks:
/hold

There are still some issue with e2e

PR #3402, opened 2021-06-03
Fix replica count

kgroschoff:
Fix replica count, as it was automatically reset to default after a few seconds. 

### Release note

```release-note
Fixed replica count, as it was automatically reset to default after a few seconds.
```

### DIFF
--- a/src/app/node-data/component.ts
+++ b/src/app/node-data/component.ts
@@ -175,12 +175,13 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(_ => (this._nodeDataService.operatingSystemSpec = this._getOperatingSystemSpec()));
 
-    this._settingsService.adminSettings.pipe(takeUntil(this._unsubscribe)).subscribe(settings => {
-      if (this.form.get(Controls.Count).pristine) {
+    this._settingsService.adminSettings
+      .pipe(take(1))
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(settings => {
         const replicas = this.dialogEditMode ? this._nodeDataService.nodeData.count : settings.defaultNodeCount;
         this.form.get(Controls.Count).setValue(replicas);
-      }
-    });
+      });
   }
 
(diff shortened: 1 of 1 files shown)


src/app/node-data/component.ts
```diff
@@ -176,8 +176,10 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       .subscribe(_ => (this._nodeDataService.operatingSystemSpec = this._getOperatingSystemSpec()));
 
     this._settingsService.adminSettings.pipe(takeUntil(this._unsubscribe)).subscribe(settings => {
```
maciaszczykm, line 178: If we would change this pipe to `take(1)` then the bug should be gone. The service is responding very quickly as admin settings are cached but the problem is that there is refresh happening sometimes.

maciaszczykm: The initialization should happen only once anyways so there is no point of keeping this subscription open.

kgroschoff: ah you're right. Always forget `take(1)` :see_no_evil:
Changed 👍

src/app/node-data/component.ts
```diff
@@ -175,10 +175,13 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(_ => (this._nodeDataService.operatingSystemSpec = this._getOperatingSystemSpec()));
 
-    this._settingsService.adminSettings.pipe(takeUntil(this._unsubscribe)).subscribe(settings => {
-      const replicas = this.dialogEditMode ? this._nodeDataService.nodeData.count : settings.defaultNodeCount;
-      this.form.get(Controls.Count).setValue(replicas);
-    });
+    this._settingsService.adminSettings
+      .pipe(take(1))
+      .pipe(takeUntil(this._unsubscribe))
```
maciaszczykm, line 180: Last thing would be to remove this line as usually we don't keep both together. There is almost no chance to fire unsubscribe before admin settings cache would be returned.

kgroschoff: done :)

PR #3429, opened 2021-06-11
Clarify PV cleanup during cluster delete

wozniakjan:
https://github.com/kubermatic/kubermatic/pull/7189
https://kubernetes.io/docs/concepts/storage/persistent-volumes/#retain

Our cluster cleanup algorithm can still leave some residual volumes in the cloud provider. Deleting dynamically provisioned `PVs` bypasses kubernetes triggers for remote volumes cleanup. Mostly we leverage dynamically provisioned `PVs`, at which point we can rely on the CSI plugins and cloud providers to do their job. For statically provisioned `PVs` that have `retain` as their `ReclaimPolicy`, we can't do much and need to leave it to the user to cleanup the volumes in the cloud provider manually.

I didn't want to overload the message here by adding all the context explained above, but I think it might be worth adding a better explanation than what I am proposing, I am open to suggestions.

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/cluster-details/cluster-delete-confirmation/template.html
+++ b/src/app/cluster/cluster-details/cluster-delete-confirmation/template.html
@@ -40,7 +40,7 @@ limitations under the License.
     <div class="km-delete-cluster-checkbox"
          [matTooltip]="getCheckboxTooltip()">
       <mat-checkbox formControlName="clusterVolumeCleanupCheckbox"
-                    id="km-delete-cluster-volume-cleanup">Cleanup connected volumes (PVs and PVCs)</mat-checkbox>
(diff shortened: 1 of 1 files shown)


maciaszczykm:
Sounds good to me. @floreks, please take a look.

wozniakjan:
thanks @maciaszczykm, btw do you think this can benefit from potentially on hover tooltip where we can provide more context? A link to https://kubernetes.io/docs/concepts/storage/persistent-volumes/#retain might be helpful. Or digest the information in our own words.

floreks:
I'd try to avoid linking to external pages as it might not work in more strict environments with firewalls.

/lgtm

PR #3441, opened 2021-06-15
Add condition to opa components call

kgroschoff:
Added condition to opa components call.

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/cluster-details/component.ts
+++ b/src/app/cluster/cluster-details/component.ts
@@ -161,7 +161,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
                 : of([])
             )
             .concat(
-              this._canReloadNodes() && this.isOPAEnabled() && this.isHavingMD()
+              this._canReloadNodes() && this.isOPAEnabled() && this.hasMachineDeployments()
                 ? [
                     this._opaService.constraints(this.projectID, this.cluster.id),
                     this._opaService.gatekeeperConfig(this.projectID, this.cluster.id),
@@ -403,7 +403,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
     );
   }
 
-  isHavingMD(): boolean {
+  hasMachineDeployments(): boolean {
     return !_.isEmpty(this.machineDeployments);
   }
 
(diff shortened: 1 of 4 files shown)


kgroschoff:
Maybe it is just a coincidence, but today I had problems to create machine deployments, especially when opa is enabled. It seems like the endpoints for constraints & gatekeeper config are called too early, which then breaks the machine deployment creation. 

PTAL @maciaszczykm / @floreks

src/app/cluster/cluster-details/component.ts
```diff
@@ -403,6 +403,10 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
     );
   }
 
+  isHavingMD(): boolean {
```
maciaszczykm, line 406: s/isHavingMD/hasMachineDeployments

src/app/cluster/cluster-details/constraints/component.ts
```diff
@@ -97,11 +98,11 @@ export class ConstraintsComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   isLoadingData(data: Constraint[]): boolean {
-    return _.isEmpty(data) && !this.isClusterRunning;
+    return _.isEmpty(data) && !this.isClusterRunning && !this.hasMachineDeployments;
   }
 
   hasNoData(data: Constraint[]): boolean {
-    return _.isEmpty(data) && this.isClusterRunning;
+    return _.isEmpty(data) && this.isClusterRunning && this.hasMachineDeployments;
```
maciaszczykm, line 105: What will happen if there is no data, cluster is running but has no machine deployments? I think due to the new checks you have added not all scenarios are covered. `hasNoData` will return false and the zero state message will not be displayed.

kgroschoff: Sorry, I've missed your comment.

Yes you are right. I thought about all cases again, and I think the machine deployments check isn't necessary at all for this views. data + cluster running is the important part and it doesn't change anything if there are machine deployments or not 👍

maciaszczykm:
Looks good now, thanks.

/lgtm

kgroschoff:
No problem, you're hint was right :)

PR #3461, opened 2021-06-28
Redesign admin panel

### LINKED ISSUE #3434: Redesign Admin Panel
Based on @cschieder design

https://www.figma.com/file/dJQKgUPzIDND36o2Nq8dct/Loodse-UI?node-id=5163%3A319

image

floreks:

- Added new lazy-loaded views
- Redesigned admin panel handling
- Updated e2e to match new admin panel design

**Note** Due to how we handle sidenav open/close feature it is hard to extend and support it on the new admin panel page (Ref: https://github.com/kubermatic/dashboard/issues/3468).

image
image
image
image
image
image

Closes #3434

### Release note

```release-note
Redesigned Admin Panel
```


src/app/settings/admin/defaults/style.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  mat-checkbox {
+    margin-right: 48px;
+  }
+
+  .km-pointer,
+  .km-icon-error {
+    margin-left: 4px;
+    min-height: 16px;
+    min-width: 16px;
```
maciaszczykm, line 35: You can use size mixin here.

maciaszczykm:
The code looks good, I'll test how it works.

maciaszczykm:
@cschieder The "Go Back" is redirecting to the previous view that was accessed. Since we have more than one view for admin settings, if you switch the view inside it will not redirect you back outside the admin settings anymore. Should we adjust it somehow? I mean user goes to the admin panel and switches to dynamic datacenters for example, clicking on the "Go Back" button will move him back to the "Administrators" view (first one) not to the projects/cluster (whatever was used before admin panel).

@floreks One more thing is that in the Slack voting we decided to go for "1" menu order from https://github.com/kubermatic/dashboard/issues/3434.

cschieder:
@maciaszczykm Yes, ideally we adjust that, the behavior you described seems a bit awkward. Good catch 👍

floreks:
@cschieder adjust how? Remove the button?

cschieder:
@floreks can we go back to the last non-admin-settings page the user was on? i. e. if you were on a specific cluster before and then go to admin setting, then when you click back you get sent back to that cluster page again.

floreks:
We can altough it will require some hacky logic from our side. I'll check.

PR #3484, opened 2021-07-02
Bump dependencies and fix scss warnings

floreks:
- Bump all our dependencies at once
- Fix scss warnings issue related to using `/`

### Release note

```release-note
NONE
```


src/app/core/components/navigation/project/style.scss
```diff
@@ -12,7 +12,7 @@
 @import 'variables';
 
 .km-project-selector {
-  height: $toolbar-height - 8px / 4;
+  height: $toolbar-height - 8px * .25;
```
maciaszczykm, line 15: What is the order of operations here? Can we replace it with `2px`?

floreks: we definitely should :smile:

PR #3528, opened 2021-07-09
Fix Openstack credentials support

### LINKED ISSUE #3521: [Openstack] Username/password does not work for non-admin OpenStack users
Introducing application credentials caused a regression with the username/password handling for some of the OpenStack users.

> When using KKP 2.17.2 and trying to create a new user cluster using the classic username/password auth I get the error message "Error 500 couldn't get identity endpoint: No suitable endpoint could be found in the service catalog.". Using an administrator account works however. Application credentials work for the regular account though. This is a regression as this worked previously. 

**What needs to be fixed**
- When username/password is used it should always provide tenant information to other endpoints for both admin/non-admin users
- When application credentials fields are used it should not provide tenant information to other endpoints as it causes errors

floreks:
- Added credentials type selector to switch between the user/application credentials
- Updated credentials handling and additional data resolving

screencapture-localhost-8000-projects-x8tg4jx2mv-wizard-2021-07-09-12_52_59

screencapture-localhost-8000-projects-x8tg4jx2mv-wizard-2021-07-09-12_52_46

Fixes #3521

### Release note

```release-note
Redesign Openstack provider settings step to better support different type of credentials
```

### DIFF
--- a/src/app/core/services/cluster-spec.ts
+++ b/src/app/core/services/cluster-spec.ts
@@ -10,23 +10,27 @@
 // limitations under the License.
 
 import {EventEmitter, Injectable} from '@angular/core';
-import * as _ from 'lodash';
 import {CloudSpec, Cluster, ClusterType} from '@shared/entity/cluster';
 import {SSHKey} from '@shared/entity/ssh-key';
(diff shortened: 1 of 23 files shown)


zreigz:
works fine :+1:

PR #3532, opened 2021-07-12
fix reset of opa integration checkbox

kgroschoff:
The OPA integration checkbox sometimes got reset during cluster creation due to a reload of the admin settings.

### Release note

```release-note
NONE
```


kgroschoff:
PTAL @maciaszczykm / @floreks

Issue #3536, opened 2021-07-13
Network config is not working for AWS


maciaszczykm:
**What happened?**: After specifying network config for AWS cluster it is not applied. It is also not displayed in the wizard summary.

**What should happen?**: It should work the same for all providers. DigitalOcean was working when it was checked. We need to check all the providers.

PR #3537, opened 2021-07-13
Use merge when applying cluster network changes to avoid overrides

maciaszczykm:
Fixes https://github.com/kubermatic/dashboard/issues/3536.

Previous implementation was based on a wrong assumption that cluster setter in the service is always invoked with full cluster objects. In reality sometimes partial cluster objects are sent and this was causing overwrites of cluster network. I've changed pure assignment to the merge method, however it uses a different customizer. I'm not sure why in the code above the changed one we are ignoring the source array and using destination one. In my case I needed to do the opposite.

### Release note
```release-note
NONE
```


kgroschoff (commented):
/lgtm

PR #3563, opened 2021-07-29
Remove status indicator from service account view

### LINKED ISSUE #2804: Clarify/Remove liveness indicator for Service Account tokens
screenshot-2020-11-25-165930
The dashboard does not explain what it means for a Service Account to be "active" (the green dots in the screenshot). I also cannot make it _not_ be active. SSH keys do not have an activity indicator, so I guess it would be possible to not have it for Service Accounts as well?

Just remove the indicator alltogether.

kgroschoff:
Remove status indicator from service account view

Fixes #2804

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/serviceaccount/component.spec.ts
+++ b/src/app/serviceaccount/component.spec.ts
@@ -82,10 +82,6 @@ describe('ServiceAccountComponent', () => {
     expect(component).toBeTruthy();
   });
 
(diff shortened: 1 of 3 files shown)


kgroschoff:
PTAL @maciaszczykm / @floreks

PR #3598, opened 2021-08-09
[MLA] Rename User Cluster Alert Rules to User Cluster Recording & Alerting Rules

jiachengxu:
This PR renames `User Cluster Alert Rules` to `User Cluster Recording & Alerting Rules`
### Release note

```release-note
NONE
```


kgroschoff (commented):
/lgtm
/approve

Issue #3604, opened 2021-08-10
post-dashboard-master-push-ce is failing


maciaszczykm:
post-dashboard-master-push-ce is failing for some time now.

PR #3562, opened 2021-07-28
Add Allowed Registries

kgroschoff:
### What this PR does / why we need it
Adds Allowed Registries to UI. Note that this is **only** an ee feature.

### Which issue(s) this PR fixes

Fixes #3518

### Special notes for your reviewer

allowed1
allowed2

### Release note

```release-note
EE: Added Allowed Registries
```

PR #3605, opened 2021-08-10
Fix CE build, improve configs and add missing build check

maciaszczykm:
It changes the way we build and serve different app editions. We can no longer use the `getEditionDirName()` method when picking modules to import as the compiler is checking for the available directories then. The value inside `import()` needs to be constant and we can achieve that by using the file replacements in `angular.json`. Besides that, it adds a few improvements in our config files and job to check if community edition build passes too.

Fixes https://github.com/kubermatic/dashboard/issues/3604.

### Release note
```release-note
NONE
```


kgroschoff (commented):
From my side it looks good.

We just have to keep in mind to also change it for the Allowed Registries: https://github.com/kubermatic/dashboard/pull/3562
I don't mind which pr we merge first though.

maciaszczykm:
PTAL

PR #3612, opened 2021-08-13
Add checkbox for assigning Azure availability sets

maciaszczykm:
Zrzut ekranu 2021-08-17 o 09 56 00
Zrzut ekranu 2021-08-17 o 09 53 11
Zrzut ekranu 2021-08-17 o 09 53 01

Closes https://github.com/kubermatic/kubermatic/issues/7421.

Confirmed that checkbox should be checked by default. Besides that, we will avoid keeping this field in preset.

### Release note
```release-note
Added checkbox for assigning Azure availability sets.
```


maciaszczykm:
@moadqassem I would like to discuss it before proceeding. 

If you use your own provider credentials everything is pretty straight-forward and it's clear how it should work. You go to the wizard, specify the cloud spec, on the next step you have zones input enabled or disabled and there is nothing to clarify.

Unfortunately, it is not the case when user would like to use preset. Currently we have a rule that if you use preset all other provider fields are disabled and values are assigned in the API (whole cloud spec). This is not possible with "assign availability zones" as we need to know on the next step what is the value of it but presets cannot be read in the UI. Anyways, at the moment this new field is not part of the preset. What we can do is to change the API to be able to use preset with this field at the same time (at the moment it seems to be either preset or direct values) and adjust the UI to be able to always select the checkbox (even if the preset was selected). So it would look something like this:

Zrzut ekranu 2021-08-13 o 15 03 16

What do you think?

src/app/shared/entity/cluster.ts
```diff
@@ -125,6 +125,7 @@ export class AzureCloudSpec {
   tenantID: string;
   vnet: string;
   loadBalancerSKU: string;
+  assignAvailabilitySet: string;
```
moadqassem, line 128: Shouldn't this one be boolean?

maciaszczykm: I haven't pushed my latest changes as I was still checking out how to solve the issue described above and the pull request is still work in progress. I have it fixed locally.

src/app/wizard/step/provider-settings/provider/extended/azure/component.ts
```diff
@@ -95,17 +97,25 @@ export class AzureProviderExtendedComponent extends BaseFormValidator implements
...
+      Object.values(Controls)
+        .filter(control => control !== Controls.AssignAvailabilitySet)
+        .forEach(control => this._enable(!preset, control))
+    );
 
     merge(this._clusterSpecService.providerChanges, this._clusterSpecService.datacenterChanges)
       .pipe(takeUntil(this._unsubscribe))
-      .subscribe(_ => this.form.reset());
+      .subscribe(_ => {
+        // TODO: This causes reset of all default values set in the init stage. It should be fixed.
```
kgroschoff, line 115: In case you haven't done yet: Can you please open an issue, so that we don't forget about it?

kgroschoff (commented):
/lgtm

src/app/wizard/step/provider-settings/provider/extended/azure/component.ts
maciaszczykm: Sure, I will do.

moadqassem (commented):
/approve
/lgtm

PR #3615, opened 2021-08-16
Update aws spot instance max price to be bigger than 0

### LINKED ISSUE #3613: AWS spot instance max price should be float instead of integer
**User Story**
As a user I would like to assign a decimal number as the max price for my aws spot instance instead of providing only integers

**Acceptance Criteria**
- The field `Spot instance max price` should accept cents as well(float)

floreks:
- Added `largerThan` validator
- Updated spot instance validation to allow numbers bigger than 0

Fixes #3613

### Release note

```release-note
NONE
```


floreks:
1. Update the invalid `SpotInstancePersistentRequest` description
2. See if we can use some custom formatted for the price display.

/hold

PR #3633, opened 2021-08-26
OPA design follow up

### LINKED ISSUE #3583: Admin Settings OPA design follow up
- [x] I think OPA should be under Presets in the sidebar
- [x] Distribute table content like this
 image
20%, 20%, 14%, 26%, 10%, 10%

- [x] Bring tabs in the same line as button like this
 image
- [x] make active tab text medium weight
- [x] make grey line below tabs darker
- [x] left-align content of `Applies To` column
- [x] Remove headline in card (redundant)
- [ ] Truncate `applies to` column, add uncollapse/unfold functionality 
image

Question
- Is it better to write `All` instead of `*` ?
- Maybe we should have Default Constraints as the first tab rather than Templates? Opinions?

kgroschoff:
OPA design follow up

Fixes #3583

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/settings/admin/opa/default-constraints/component.ts
+++ b/src/app/settings/admin/opa/default-constraints/component.ts
@@ -9,7 +9,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Component, OnChanges, OnDestroy, OnInit, ViewChild, SimpleChanges} from '@angular/core';
+import {Component, OnChanges, OnDestroy, OnInit, ViewChild, SimpleChanges, AfterViewChecked} from '@angular/core';
 import {MatDialog, MatDialogConfig} from '@angular/material/dialog';
 import {MatPaginator} from '@angular/material/paginator';
 import {MatSort} from '@angular/material/sort';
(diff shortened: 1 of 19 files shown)


kgroschoff:
cd-improvements

PTAL @floreks 
I know the toggle option is a bit more static (especially the provider part) than I would like to have it, but styles became more complex each second... :unamused:

src/app/settings/admin/opa/default-constraints/component.ts
```diff
@@ -74,6 +77,13 @@ export class DefaultConstraintComponent implements OnInit, OnChanges, OnDestroy
     });
   }
 
+  ngAfterViewChecked(): void {
+    const box = document.querySelector('.applies-to-column') as HTMLElement;
```
floreks, line 81: This can most likely be achieved with the combination of `ngClass` and status check of the component. Also, using `document` without importing breaks the tree-shaking algorithms. This should be optimized.

floreks:
To be honest, this new component and so much DOM manipulation and calculation directly from code feels super hacky. We should think about a better solution here.

kgroschoff:
Hm, okay. I think it would be good to have some improvements in the upcoming release. What do you think about removing the toggle logic for now and just merge the other improvements?

floreks:
> What do you think about removing the toggle logic for now and just merge the other improvements?

That would be good.

kgroschoff:
Closing this one in favor of https://github.com/kubermatic/dashboard/pull/3643

PR #3636, opened 2021-08-26
Fix Azure issues

maciaszczykm:
- Fixes "assign availability set" behavior, now selecting it will block availability zones usage
- Allows selecting only one availability zone
Zrzut ekranu 2021-08-26 o 14 55 27

/assign @kgroschoff @floreks 
/cc @moadqassem 

### Release note

```release-note
NONE
```


kgroschoff (commented):
/lgtm

Issue #12, opened 2017-01-09
Create Breadcrumb


j3ank:
Display breadcrumb and make sure that every case needs to be covered. 

Logic:
- Should the breadcrumb displays the same logic as the the url routes?
- Should the breadcrumb acting as a progressbar and how much information should be shown?
- Should be the cluster name displayed in the breadcrumb? 

Design: 
- What happens if there is too much text and not enough space?
- Should it be this present like now? 
- How is the breadcrumb acting on a responsive design without just hiding it?

PR #3639, opened 2021-08-30
Improve buttons

### LINKED ISSUE #3624: Button Follow Up
Note: All buttons need icons now, ideally

- [x] Disabled buttons icon should be the same color as the text (instead of white) (same in dark mode)
- [x] Create Preset Dialog (Admin Settings -> Presets -> Create) `Next` is missing the arrow icon, other buttons in Admin Settings are also missing their buttons
- [x] Connect Icon (link/chain) width: 24px
- [x] Cancel Icon (x) and Arrows <- & -> should be 24 by 24 px size
- [x] Share, Download (Cluster Details Page) should be 24 by 24 px size
- [x] Add second (24px big) delete icon and use it in `delete` buttons
- [x] Add checkmark icon to last step of create (cluster create wizard, preset create in admin settings) add checkmark to `Got It` buttons (like the one for changelog), Add checkmark to the `change` button for the dialog for updating cluster version, Add checkmark to `Revoke Token` button in Revoke Token dialog
- [ ] add service account/add member/add ssh, add checkmark icon to "add" button inside dialog, add ressource to button label, i. e. `Add Member` instead of just `Add`
- [x] link icon not vertically centered in this button image

Icons (Checkmark and 24px Delete)
https://drive.google.com/drive/folders/17hAaVBhIvmfSFU3dMYp7zAypSG2Robgu?usp=sharing

kgroschoff:
Multiple ui improvements for buttons 

Fixes #3624

### Release note

```release-note
NONE
```


kgroschoff:
PTAL @floreks

src/assets/images/icons/trash-24.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
floreks, line 1: Why the `trash-24`? The point of having SVG is to be able to easily scale it without losing the quality.

src/assets/images/icons/check.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
floreks, line 1: If this is a checkmark then we already have a `tick` icon.

src/assets/css/_images.scss
```diff
@@ -397,3 +405,7 @@
 .km-os-image-flatcar {
   @include mixins.background-image("/assets/images/os/flatcar.svg", 24px);
 }
+
+.with-text {
```
floreks, line 409: I'd rather use something like we have for table column width, just for the text.

https://github.com/kubermatic/dashboard/blob/bfbb48dd3f1e0612d2ea8bc686d57cd721959d0f/src/assets/css/global/_main.scss#L277-L283

src/assets/images/icons/trash-24.svg
cschieder: Hi, actually I can add something here, the problem is that on 4k monitors this would be correct, but on HD screens when freely scaling small svgs they end up blurry or suboptimal in one way or another when used in different sizes than they were created for, that's why the same icons end up having different versions

floreks: If that's the case then the original SVG has to be invalid. Vector graphic by design is scalable infinitely as it is only a definition of vectors (paths) and not pixels. Correctly created SVG should not be kept in multiple sizes. There should be always only a single file and it should be scaled with CSS.

cschieder: No, icons do not work at any size just because they're svgs. Yes, they're vector, yes they scale in every direction basically infinitely. But because of HD displays what can happen is that on certain sizes vector lines end up in a halfpixel step, if that happens it ends up blurry.  (left 24px trash icon, right small trash icon upscaled)
image
image

floreks: This issue is irrelevant to the display resolution. It would still be present even on a 16k display. The only difference is that it would not be noticeable since the pixel density is so big that when antialiasing is applied you just can't see this effect.

Originally, this is a result of incorrect grid alignment. Since the icon was originally prepared for the 18px then whenever it is scaled to the size that does not align with the original grid size it might end up in this half-pixel state and then when antialiasing is applied the half-pixel is rendered as a semi-translucent pixel.

That's why it is so important to design the original icon at a set size that always aligns with the grid size chosen for the application. This is also the reason why i.e. the base grid size of angular material is 8px and the default size of their icons is 24px. 

To sum up, we should have a defined grid size, use a base icon size that aligns with the grid and then scale icons only to the sizes that match the original grid.

kgroschoff: Not sure why, but the icons are acting strange when only using the existing or new svg 🤔 

In button case:
left: trash-24 (24px) / right: trash scaled to 24px
delete2 delete

left: trash (18px) / right: trash-24 scaled to 18px
delete242 delete24

src/assets/images/icons/check.svg
kgroschoff: you are right, totally forgot. 
Just noticed, that the new check icon seems to be thinner than the existing one. Or maybe it's just the same behaviour as above 🤔

src/assets/images/icons/trash-24.svg
cschieder: Given the current priorities, we cannot pursue this right now. I made a note for potential perusal at a later date

kgroschoff: @floreks So should we merge it now and improve later?

floreks: I'd go with the old way for now as I am afraid that if we keep this change it will never be picked up again and improved. The "blurry" icon might just be annoying enough to force us to properly work on this in the near future :slightly_smiling_face:

cschieder: No, let's not release blurry icons, that gives the UI an unprofessional look. Before we even decide to rework our icons, I want to do research and confirm that this is/there is a better way of doing icon sets. However, the priority of that is low, the benefit of it is small compared to the effort required, especially considering that there are other tasks that require our attention first. I understand your concern about implementing hundreds of different icons sizes, but it should remain an exception rather than the rule either way because it's not often that an icon is needed in different sizes anyway.

floreks: If we'd merge this then there are a couple of changes that need to be done first.

1. Use SVG image stacking in order to use a single file for every icon. Thanks to [SVG framgnet identifier](https://caniuse.com/svg-fragment) it's easy to use and can be handled in a generic way in the code.

Example SVG with merged trash icons (18, 24)
```svg
<?xml version="1.0" encoding="utf-8"?>
<svg xmlns="http://www.w3.org/2000/svg">
  <style>
    g {
      display: none;
    }

    g:target {
      display: inline;
    }
  </style>

  <g id="18">
    <path
      d="M9 0C8.44772 0 8 0.447715 8 1V2H3C2.44772 2 2 2.44772 2 3C2 3.55228 2.44772 4 3 4H15C15.5523 4 16 3.55228 16 3C16 2.44772 15.5523 2 15 2H10V1C10 0.447715 9.55228 0 9 0Z"
      fill="#4D555C"/>
    <path fill-rule="evenodd" clip-rule="evenodd"
          d="M3.18048 5C2.56254 5 2.0925 5.55487 2.19409 6.1644L3.86076 16.1644C3.94112 16.6466 4.35831 17 4.84715 17H13.1529C13.6417 17 14.0589 16.6466 14.1393 16.1644L15.806 6.1644C15.9075 5.55487 15.4375 5 14.8196 5H3.18048ZM4.36094 7L5.69428 15H12.3058L13.6391 7H4.36094Z"
          fill="#4D555C"/>
  </g>

  <g id="24">
    <path fill-rule="evenodd" clip-rule="evenodd"
          d="M12 2C11.4477 2 11 2.44772 11 3V5H5C4.44772 5 4 5.44772 4 6C4 6.55228 4.44772 7 5 7H19C19.5523 7 20 6.55228 20 6C20 5.44772 19.5523 5 19 5H13V3C13 2.44772 12.5523 2 12 2ZM6.51408 10L8.591 19H15.4089L17.4859 10H6.51408ZM5.25702 8C4.61394 8 4.13803 8.59824 4.28263 9.22486L6.82109 20.2249C6.9258 20.6786 7.32983 21 7.79549 21H16.2045C16.6701 21 17.0741 20.6786 17.1789 20.2249L19.7173 9.22486C19.8619 8.59824 19.386 8 18.7429 8H5.25702Z"
          fill="#4D555C"/>
  </g>
</svg>
```

Example usage in scss (It can be made more generic with some generation function)
```scss
.km-icon-delete {
  @include mixins.mask-image("/assets/images/icons/trash.svg#18", 18px);

  &.i-24 {
    @include mixins.mask-image("/assets/images/icons/trash.svg#24", 24px);
  }
}
```

2. Properly adjust prepared icons to the grid. Currently 18px icon has smaller borders than 24px version. It causes inconsistent scaling (24px version is smaller compared to the 18px upscaled to 24px)

trash-24
trash-18

This is the result of a slightly "fixed" SVG. Can you tell for sure which one is based on which icon version?
trash

floreks:
@kgroschoff PTAL to make sure I haven't missed anything.

kgroschoff:
@floreks LGTM

kgroschoff:
@floreks needed to rebase due to multiple merge issues. PTAL

src/assets/css/_images.scss
```diff
@@ -51,6 +51,18 @@
   @include mixins.mask-image("/assets/images/icons/clear.svg", 24px);
 }
 
+.km-icon-check {
+  @include mixins.mask-image("/assets/images/icons/check.svg#16", 16px);
+
+  &.i-12 {
+    @include mixins.mask-image("/assets/images/icons/check.svg#16", 12px);
```
maciaszczykm, line 58: Shouldn't it be `#12`?

maciaszczykm:
/hold

One question added.

src/assets/css/_images.scss
maciaszczykm: Nevermind, I've verified that there is no such version in the SVG 😅

PR #3643, opened 2021-09-01
Improve ui for opa default constraints

kgroschoff:
Improve ui for opa default constraints 

### Release note

```release-note
NONE
```


kgroschoff:
PTAL @floreks 
Removed everything that was related to this toggle thingy. 

Looks like this now, I think it is much better that the current status on dev:
constraints-default

PR #3649, opened 2021-09-02
Avoid setting custom page title when user is not logged in

floreks:
It was causing consistent notification error popup on the login screen.

### Release note

```release-note
NONE
```


kgroschoff (commented):
/lgtm

PR #3655, opened 2021-09-03
Add Backup Buckets

### LINKED ISSUE #3652: Bucket Settings for Backups
Add view in Admin Settings, to manage bucket settings for backups.

kgroschoff:
Add Backup Buckets to Admin Settings

Fixes #3652

### Release note

```release-note
Added Backup Buckets to Admin Settings
```


kgroschoff:
PTAL @floreks 

Few informations that might be important:
We decided to display all available seeds by default in this table, that's why e.g. the asian seed is not displaying more that it's name atm, because we didn't configured anything so far.
Since we already display all seeds by default and there isn't any specific "create" endpoint, we go without an `Add` button and just use the two edit buttons in each row. First one (with the pen) is for editing the bucket settings (Bucket + Endpoint) and the second one (new credentials icon) is for editing the credentials. They have seperate endpoints and we generally wanted to have two seperate dialogs for it.

Atm the `Bucket` isn't used in backend, so for now we can insert a random name. It will be updated, once we have a real bucket.
But please **do not** change the credentials atm. They are in use and it might break stuff in the background. However, for security reasons the credentials will never be displayed in the edit dialog, even though they exists.

backup-buckets

update-settings

update-credentials

kgroschoff:
/hold
we can't merge this pr until the patch seed endpoint is fixed. it is causing the datacenter issues

kgroschoff:
@floreks / @maciaszczykm Even though the api bug isn't fixed yet, do you have time for a code-based review (today or tomorrow)? Since tomorrow is my last working day before vacation, I guess it would be good to make this pr as ready as possible, so that you don't have to take over in the end to get it into the release (or maybe just a tiny bit).

src/app/core/services/datacenter.ts
```diff
@@ -30,6 +30,8 @@ export class DatacenterService {
   private _seedSettings$ = new Map<string, Observable<SeedSettings>>();
   private _seedSettingsRefresh$ = new Subject<void>();
   private _seedsRefresh$ = new Subject<void>();
+  private _adminSeeds$: Observable<any[]>;
```
floreks, line 33: `any` type?

kgroschoff: Ah, good catch. Forgot to change during development, will adjust in all places

src/app/core/services/datacenter.ts
```diff
@@ -113,4 +115,29 @@ export class DatacenterService {
...
+      this._adminSeeds$ = merge(this._adminSeedsRefresh$, this._refreshTimer$)
+        .pipe(switchMap(_ => this._getAdminSeeds()))
+        .pipe(shareReplay({refCount: true, bufferSize: 1}));
+    }
+
+    return this._adminSeeds$;
+  }
+
+  private _getAdminSeeds(): Observable<AdminSeed[]> {
+    const url = `${this._restRoot}/admin/seeds`;
```
floreks, line 131: Is this a new endpoint? How does it differ from the original seeds endpoint?

src/app/settings/admin/bucket-settings/edit-bucket-setting-dialog/component.ts
```diff
@@ -0,0 +1,82 @@
...
+    });
+  }
+
+  edit(): void {
+    const backupRestore: BackupRestoreConfiguration = {
+      [Controls.Bucket]: this.form.get(Controls.Bucket).value,
+      [Controls.Endpoint]: this.form.get(Controls.Endpoint).value,
+    };
+
+    let configuration = new AdminSeed();
```
floreks, line 58: The configuration object is overridden in the next line. There is no need to create a new object here.

src/app/settings/admin/bucket-settings/edit-bucket-setting-dialog/component.ts
```diff
@@ -0,0 +1,82 @@
...
+    configuration.spec.backupRestore = backupRestore;
+
+    this._datacenterService.patchAdminSeed(configuration.name, configuration).subscribe(_ => {
+      this._matDialogRef.close();
+      this._notificationService.success('Bucket settings were successfully edited');
+      this._datacenterService.refreshAdminSeeds();
+    });
+  }
+
+  private _setBucketName(): string {
```
floreks, line 69: Why this setter returns a value? It can be simplified to a getter.

```ts
private get _bucketName(): string {
    return !!this.data.seed.spec.backupRestore ? this.data.seed.spec.backupRestore.s3BucketName : '';
}
```

src/app/core/services/datacenter.ts
kgroschoff: We're now having
1. `seeds` via `/api/v1/seed` -> used for every user group and only returns an array of seed names
2. `seedSettings` via `/api/v2/seeds/{seedName}/settings` -> returns some settings (e.g. mla) that are needed in wizard and therefor need to be readable for all user groups
3. "new" `adminSeeds` via `/api/v1/admin/seeds` -> returns all seeds including its whole settings (and also the backupRestore values). This ep is only available for admins, as it includes configurations like e.g. proxy-settings.

kgroschoff:
/hold cancel

@floreks / @maciaszczykm Bug is fixed in the backend. Ready for merge, unless you have remarks. [also feel free to take over this pr in case you need to change something during my vacation]

PR #3658, opened 2021-09-06
temporary skip provider + opa e2e tests

kgroschoff:
Temporary skip provider e2e tests as they are blocking ci too much atm

### Release note

```release-note
NONE
```

### DIFF
--- a/cypress/plugins/index.ts
+++ b/cypress/plugins/index.ts
@@ -14,12 +14,8 @@ import failFast from 'cypress-fail-fast/plugin';
 import {configuration} from './cy-ts-preprocessor';
 
 export default async (on, config) => {
+  config.ignoreTestFiles = '**/integration/providers/*.spec.ts';
   on('file:preprocessor', webpack(configuration));
   failFast(on, config);
-  return config;
-};
-
-module.exports = (on, config) => {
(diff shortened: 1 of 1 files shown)


cypress/plugins/index.ts
```diff
@@ -18,3 +18,8 @@ export default async (on, config) => {
   failFast(on, config);
   return config;
 };
+
+module.exports = (on, config) => {
```
floreks, line 22: I don't think this will work actually with the above configuration. 

1. It is using the old style of config initialization and I think it is not supported with our current cypress ts config
2. It tries to override the above correct config

Simply try to add this line to the above function.

kgroschoff: You're right, that was stupid. 🙈

floreks:
https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_dashboard/3639/pre-dashboard-test-e2e-ce/1434845237959725056

Looking at the logs it seems that the provider tests are still running and not being skipped.

floreks:
Weird I followed a link from this PR. Maybe I pasted wrong link from my clipboard :)

PR #3662, opened 2021-09-08
Replace tick icon

kgroschoff:
Replace tick icon by the new check icon, as it is no longer available and neither steps in wizard nor the health status in cluster details view was displaying checkmarks anymore.

### Release note

```release-note
NONE
```


kgroschoff:
PTAL @maciaszczykm / @floreks

PR #3695, opened 2021-09-15
Add changelog for v2.18

floreks:
As in the title

### Release note

```release-note
NONE
```

### DIFF
--- a/src/assets/config/changelog.json
+++ b/src/assets/config/changelog.json
@@ -1,1 +1,125 @@
-{}
+{
+  "externalChangelogURL": "https://github.com/kubermatic/kubermatic/blob/release/v2.18/CHANGELOG.md#kubermatic-218",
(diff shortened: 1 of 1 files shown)


src/assets/config/changelog.json
```diff
@@ -1 +1,125 @@
...
+      "links": [
+        {
+          "url": "https://docs.kubermatic.com/kubermatic/v2.18/guides/monitoring_logging_alerting/user_cluster/",
+          "caption": "MLA"
+        }
+      ]
+    },
+    {
+      "category": "added",
+      "description": "Cluster Templates for Deploying Optimal Clusters Instantly",
```
maciaszczykm, line 30: It's a bit inconsistent because some of the descriptions are using title case but some are not. I guess this is a copy from original changelog so we should raise this point next time.

PR #3709, opened 2021-09-21
Expose CNI config in wizard

wozniakjan:
By default, all our clusters have Canal CNI. Backend already exposes CNI configuration knobs and internally we can spin up a cluster with Cilium. This exposes the CNI config to the end user.

depends on: https://github.com/kubermatic/kubermatic/pull/7853, https://github.com/kubermatic/kubermatic/pull/7851, https://github.com/kubermatic/kubermatic/pull/7862
closes: https://github.com/kubermatic/kubermatic/issues/7577

### Release note

```release-note
Allow users to select one of the two supported CNIs - Canal or Cilium
```

### DIFF
--- a/src/app/shared/entity/cluster.ts
+++ b/src/app/shared/entity/cluster.ts
@@ -223,7 +223,6 @@ export class ClusterNetwork {
 
 export class CNIPlugin {
   type: string;
-  version: string;
 }
 
 export class NetworkRanges {
--- a/src/app/shared/entity/cluster.ts
+++ b/src/app/shared/entity/cluster.ts
@@ -212,7 +212,7 @@ export class ClusterSpec {
   mla?: MLASettings;
   containerRuntime?: ContainerRuntime;
   clusterNetwork?: ClusterNetwork;
-  cniPlugin?: CNIPlugin;
+  cniPlugin?: CNIPluginConfig;
 }
 
 export class ClusterNetwork {
@@ -221,7 +221,7 @@ export class ClusterNetwork {
   services?: NetworkRanges;
 }
 
-export class CNIPlugin {
+export class CNIPluginConfig {
   type: string;
 }
 
@@ -234,9 +234,9 @@ export enum ProxyMode {
   iptables = 'iptables',
 }
 
-export enum CNIPlugins {
-  canal = 'canal',
-  cilium = 'cilium',
+export enum CNIPlugin {
+  Canal = 'canal',
+  Cilium = 'cilium',
 }
 
 export class AuditLoggingSettings {
--- a/src/app/shared/entity/cluster.ts
+++ b/src/app/shared/entity/cluster.ts
@@ -212,12 +212,18 @@ export class ClusterSpec {
   mla?: MLASettings;
   containerRuntime?: ContainerRuntime;
   clusterNetwork?: ClusterNetwork;
+  cniPlugin?: CNIPlugin;
 }
 
 export class ClusterNetwork {
   pods?: NetworkRanges;
   proxyMode?: ProxyMode;
   services?: NetworkRanges;
+}
+
+export class CNIPlugin {
+  type: string;
+  version: string;
 }
 
 export class NetworkRanges {
@@ -227,6 +233,11 @@ export class NetworkRanges {
 export enum ProxyMode {
   ipvs = 'ipvs',
   iptables = 'iptables',
+}
+
+export enum CNIPlugins {
+  canal = 'canal',
+  cilium = 'cilium',
 }
 
 export class AuditLoggingSettings {
--- a/src/app/wizard/step/cluster/component.ts
+++ b/src/app/wizard/step/cluster/component.ts
@@ -64,7 +64,6 @@ enum Controls {
   PodsCIDR = 'podsCIDR',
   ServicesCIDR = 'servicesCIDR',
   CNIPlugin = 'cniPlugin',
-  CNIPluginVersion = 'cniPluginVersion',
 }
 
 @Component({
@@ -94,8 +93,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
   asyncLabelValidators = [AsyncValidators.RestrictedLabelKeyName(ResourceType.Cluster)];
   proxyMode = ProxyMode;
   cniPlugin = CNIPlugins;
-  csiPluginVersions = [];
-  availableProxyModes = [];
+  availableProxyModes = ['ipvs', 'iptables'];
   readonly Controls = Controls;
   private _datacenterSpec: Datacenter;
   private _seedSettings: SeedSettings;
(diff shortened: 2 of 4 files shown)


src/app/wizard/step/cluster/component.ts
```diff
@@ -247,6 +257,14 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
     return AdmissionPluginUtils.getPluginName(name);
   }
 
+  updateCNIPluginOptions() {
+    if (this.controlValue(Controls.CNIPlugin) === 'cilium') {
+      this.availableProxyModes = ['ipvs', 'iptables', 'eBPF'];
```
rastislavs, line 262: hmm, I put lowercase `ebpf` here: https://github.com/kubermatic/kubermatic/pull/7861

wozniakjan: I will make it lowercase too

src/app/shared/entity/cluster.ts
```diff
@@ -229,6 +234,11 @@ export enum ProxyMode {
   iptables = 'iptables',
 }
 
+export enum CNIPlugins {
```
maciaszczykm, line 237: For enum I'd use singular form, so `CNIPlugin`.

src/app/wizard/step/cluster/component.ts
```diff
@@ -193,6 +198,10 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       .valueChanges.pipe(takeUntil(this._unsubscribe))
       .subscribe(() => (this._clusterSpecService.admissionPlugins = this.form.get(Controls.AdmissionPlugins).value));
 
+    this.control(Controls.CNIPlugin).valueChanges.subscribe(() => {
+      this.updateCNIPluginOptions();
+    });
```
maciaszczykm, line 203: Please change to one-liner:

```ts
this.control(Controls.CNIPlugin).valueChanges.subscribe(_ => this.updateCNIPluginOptions());
```

Also, add unsubscribe (see other subscriptions for examples).

src/app/wizard/step/cluster/component.ts
```diff
@@ -247,6 +257,14 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
     return AdmissionPluginUtils.getPluginName(name);
   }
 
+  updateCNIPluginOptions() {
+    if (this.controlValue(Controls.CNIPlugin) === 'cilium') {
```
maciaszczykm, line 261: Use enum in comparision instead of the string value.

src/app/wizard/step/cluster/component.ts
```diff
@@ -284,6 +302,11 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
   private _getClusterEntity(): Cluster {
     const pods = this.controlValue(Controls.PodsCIDR);
     const services = this.controlValue(Controls.ServicesCIDR);
+    const cniPluginType = this.controlValue(Controls.CNIPlugin);
+    let cniPlugin = null;
+    if (cniPluginType !== '') {
+      cniPlugin = {type: cniPluginType};
+    }
```
maciaszczykm, line 309: Short form:

``` ts
const cniPlugin = cniPluginType !== '' ? {type: cniPluginType} : null;
```

src/app/wizard/step/cluster/template.html
```diff
@@ -52,13 +52,25 @@
         <mat-card-title>Network Configuration</mat-card-title>
       </mat-card-header>
       <div fxLayout="column">
+        <mat-form-field class="km-dropdown-with-suffix">
+          <mat-label>CNI Plugin</mat-label>
+          <mat-select [formControlName]="Controls.CNIPlugin"
+                      class="km-select-ellipsis"
+                      disableOptionCentering>
+            <mat-option [value]="cniPlugin.canal">canal</mat-option>
```
maciaszczykm, line 60: On official pages both Canal and Cilium are starting with upper letters. Can we follow that in the UI?

maciaszczykm:
I've added some remarks. Really good job for a first thing done in frontend :) 

I'll test it later.

src/app/shared/entity/cluster.ts
wozniakjan: I had a problem that it the clashed with `CNIPlugin` class and I thought since the json schema for API has `cniPlugin`, that I should rather conform there.

How about I make the class `CNIPluginConfig`, so the enum can be `CNIPlugin` without the names clashing?

src/app/wizard/step/cluster/template.html
wozniakjan: yes, I think we can display the proper names but then need to pass lowercase to the backend. I just thought that is our convention, since we show `ipvs` despite the fact [it's all caps](https://docs.projectcalico.org/networking/use-ipvs) elsewhere

src/app/shared/entity/cluster.ts
maciaszczykm: I didn't notice the class name. Renaming sounds good but it can also stay as it is. It's minor thing.

src/app/wizard/step/cluster/template.html
maciaszczykm: No, the IPVS was not intentional I think. Let's use official names.

src/app/wizard/step/cluster/component.ts
```diff
@@ -258,7 +260,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
   }
 
   updateCNIPluginOptions() {
-    if (this.controlValue(Controls.CNIPlugin) === 'cilium') {
+    if (this.controlValue(Controls.CNIPlugin) === CNIPlugin.cilium) {
```
maciaszczykm, line 263: Please also change the enum keys to use upper case. The internal value can be the same as now.

src/app/wizard/step/cluster/component.ts
```diff
@@ -198,9 +198,11 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       .valueChanges.pipe(takeUntil(this._unsubscribe))
       .subscribe(() => (this._clusterSpecService.admissionPlugins = this.form.get(Controls.AdmissionPlugins).value));
 
-    this.control(Controls.CNIPlugin).valueChanges.subscribe(() => {
-      this.updateCNIPluginOptions();
-    });
+    this.control(Controls.CNIPlugin)
+      .valueChanges.pipe(takeUntil(this._unsubscribe))
+      .subscribe(_ => {
```
maciaszczykm, line 203: You can also remove the `{}`.

src/app/wizard/step/cluster/component.ts
```diff
@@ -129,6 +133,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       [Controls.ProxyMode]: this._builder.control(''),
       [Controls.PodsCIDR]: new FormControl('', [CIDR_PATTERN_VALIDATOR]),
       [Controls.ServicesCIDR]: new FormControl('', [CIDR_PATTERN_VALIDATOR]),
+      [Controls.CNIPlugin]: new FormControl(''),
```
maciaszczykm, line 136: Should it be required value?

wozniakjan: doesn't need to be, we autoconfigure on backend if it's empty

maciaszczykm:
I've tested in manually. Works fine. Last thing that seems to be missing is entry in the wizard summary:

Zrzut ekranu 2021-09-23 o 11 50 20

wozniakjan:
thanks! good catch with the cluster summary, I think it should be implemented in the last commit :)

cluster summary

PR #3742, opened 2021-10-01
Prevent multiple clicks on buttons

### LINKED ISSUE #3697: Buttons are not blocked after first click
**What happened?**: Buttons are not blocked after first click so if they are clicked fast enough multiple requests to the backend may be send.

**What should happen?**: Even if user clicks save or other button multiple times only 1 request is send.

**How to reproduce?**: See cluster templates save button or backup create button.

**Environment**
- UI Version: master
- API Version: master
- Domain: dev
- Others: Chrome on macOS

kgroschoff:
Prevent multiple clicks on buttons

Fixes #3697

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/backup/list/automatic-backup/add-dialog/template.html
+++ b/src/app/backup/list/automatic-backup/add-dialog/template.html
@@ -103,7 +103,8 @@ limitations under the License.
 
   <mat-dialog-actions>
     <button mat-flat-button
-            (click)="save()"
+            kmThrottleClick
+            (throttleClick)="save()"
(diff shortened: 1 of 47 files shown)


kgroschoff:
PTAL @maciaszczykm :)
For now I "only" added the directive to buttons that trigger an api request, but in theory we could add it to all other buttons, too. I personally don't see the need for it atm (or better said I don't think it is as "risky" as for the ones with an api request), but open to add it to the remaining ones if we want to do it for all. 😉

PR #3755, opened 2021-10-07
Add validation for floats

### LINKED ISSUE #3664: Minimum spot instance price input accepts invalid value?
**What happened?**:
Zrzut ekranu 2021-09-8 o 14 49 10

**What should happen?**: Should the field be invalid in that case?

**How to reproduce?**: Go to wizard, pick AWS as provider and select spot instance checkbox.

**Environment**
- UI Version: master
- API Version: master
- Domain: dev
- Others: Chrome on macOS

kgroschoff:
Add validation for floats

Related to #3664

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/node-data/extended/provider/aws/component.ts
+++ b/src/app/node-data/extended/provider/aws/component.ts
@@ -16,6 +16,7 @@ import {pushDown} from '@shared/animations/push';
 import {NodeCloudSpec, NodeSpec} from '@shared/entity/node';
 import {NodeData} from '@shared/model/NodeSpecChange';
 import {BaseFormValidator} from '@shared/validators/base-form.validator';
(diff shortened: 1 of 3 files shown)


floreks:
Actually, string validation for float is not what we need here. Since it is a field of type `number` it already accepts only digits and `.` or `,`. Example entry provided by Marcin that is acceptable now actually returns a `control.value` set to `null`. Another issue is that this new validator does not allow an empty value and basically makes it required since providing a value and then removing it returns `null`.

kgroschoff:
@floreks Oh I've missed your comment. Good catch with the empty value that should definitely be allowed, even after change. Nevertheless, since this solution seems to be not the correct one, should I close this pr and we take a look into this issue at a later point again?

PR #3784, opened 2021-10-18
Fix license header

xrstf:

This PR fixes the format of the Kubermatic license headers, which got squished when they were originally introduced in this repo. This includes a few files were the line endings were changed from Windows to Unix. A new `pre-dashboard-verify-boilerplate` Prow job ensures that the headers stay consistent from now on.

This PR removes the npm-based license check, mainly because it wasn't able to handle CE and EE. And because I think it make sense if we use the same tooling for boilerplate checking throughout the org on Github.

This PR also fixes the last few "loodse" occurences.

### Release note
```release-note
NONE
```


maciaszczykm:
@xrstf Is there any script to insert these automatically?

PR #3816, opened 2021-10-29
Improve azure error handling and more

### LINKED ISSUE #3756: Error message block the Next button during user cluster creation in Azure
**What happened?**:
I wanted to create a user cluster in azure without using the presets and use my own credentials. An error message kept showing up regarding RouteTableClient is not found that blocked the next button and the back button so I can't proceed with the cluster creation.

**What should happen?**:
The error message shouldn't appear since the there is no real error.

**How to reproduce?**:
- Create an azure user cluster in KKP 
- In the credentials step don't use the presets and enter the credentials manually 
- Choose a resource group
- The error message will keep showing up blocking you from pressing Next or Back

**Environment**
- UI Version:
- API Version:
- Domain:
- Others:

floreks:
- Add error mapping entry for the resource group related errors. It will now display a single error notification instead.
- Restore a shebang line in our bash scripts
- Improve the notification overlay area not to block visible elements that were placed below the notification element
- Slightly increase the debounce time to increase the chance of avoiding multiple overlapping calls that might result in error notification

Fixes #3756

### Release note

```release-note
NONE
```

### DIFF
--- a/hack/check-licenses.sh
+++ b/hack/check-licenses.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 # Copyright 2020 The Kubermatic Kubernetes Platform contributors.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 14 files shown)


maciaszczykm:
Good improvement, error message is more clear now and there is more time until it shows.

/lgtm

Issue #3619, opened 2021-08-17
Update cluster creation time on cluster detail page


xrstf:
**What is the problem, where is it and why is it an issue?**
I created a cluster using KKP. I kept my browser tab open and forgot about it. A few hours later I open the tab again and it says "Cluster created: a few seconds ago."

**Suggest how to fix it, if possible**
If this was a static time ("created: 2:02pm" or whatever), it would be fine to keep it, but if we already go the extra mile ( :-) ) and make it relative, I propose we also try to keep it up-to-date. Maybe we could simply redraw this info after every NodeDeployment fetch or whatever regular updating you're already doing in the background? It doesn't need to have second precision, updating once per minute would be more than fine.

I am using Firefox on Ubuntu.

PR #3823, opened 2021-11-05
Fix relative time display

maciaszczykm:
Fixes https://github.com/kubermatic/dashboard/issues/3619.

### Release note
```release-note
NONE
```

### DIFF
--- a/src/app/shared/pipes/relativetime.ts
+++ b/src/app/shared/pipes/relativetime.ts
@@ -15,9 +15,9 @@
 import {Pipe, PipeTransform} from '@angular/core';
 import moment from 'moment';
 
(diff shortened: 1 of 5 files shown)


src/app/shared/pipes/relativetime.ts
```diff
@@ -15,9 +15,9 @@
 import {Pipe, PipeTransform} from '@angular/core';
 import moment from 'moment';
 
-@Pipe({name: 'kmRelativeTime'})
+@Pipe({name: 'kmRelativeTime', pure: false})
```
maciaszczykm, line 18: This is the solution. Small pipes can be marked as impure. That means it can produce different results with the same input what is exactly the case here.

Issue #3694, opened 2021-09-15
Error notifications show after creating cluster with OPA enabled


maciaszczykm:
**What happened?**: After creating OpenStack cluster with OPA enabled I keep getting following error notifications when I'm on the cluster details:

```
no matches for kind "allowedregistry" in version "constraints.gatekeeper.sh/v1beta1"
```

```
no matches for kind "K8sPSPAllowPrivilegeEscalationContainer" in version "constraints.gatekeeper.sh/v1beta1"
```

OPA Gatekeeper controller and audit statuses are still pending but perhaps there is a way to not request for the data that produces errors until everything will be up and running?

**What should happen?**: There should be less errors in the UI.

**How to reproduce?**: As above.

**Environment**
- UI Version: master
- API Version: master
- Domain: dev
- Others: Chrome on macOS

PR #3851, opened 2021-11-12
Improve OPA loading check

maciaszczykm:
Fixes https://github.com/kubermatic/dashboard/issues/3694.

Previous check was preventing from loading the data if the cluster had no machine deployments created. Machine deployments are loaded in the same `combineLatest` as OPA stuff.

### Release note
```release-note
Fix a bug where OPA data was not retrieved when there were no machine deployments in the cluster.
```

### DIFF
--- a/src/app/cluster/cluster-details/component.ts
+++ b/src/app/cluster/cluster-details/component.ts
@@ -159,7 +159,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
                 : of([] as MasterVersion[])
             )
             .concat(
-              this._canReloadNodes()
+              this.isClusterRunning
                 ? [
                     this._clusterService.addons(this.projectID, this.cluster.id),
                     this._clusterService.nodes(this.projectID, this.cluster.id),
@@ -169,7 +169,7 @@ export class ClusterDetailsComponent implements OnInit, OnDestroy {
                 : [of([] as Addon[]), of([] as Node[]), of([] as MachineDeployment[]), of({} as ClusterMetrics)]
             )
             .concat(
-              this._canReloadNodes() && this.isMLAEnabled()
+              this.isClusterRunning && this.isMLAEnabled()
                 ? [
                     this._mlaService.alertmanagerConfig(this.projectID, this.cluster.id),
                     this._mlaService.ruleGroups(this.projectID, this.cluster.id),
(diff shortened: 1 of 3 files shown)


kgroschoff (commented):
I remember that it was only possible to create a new constraint (without getting a lot of errors) if there are machine deployments available 🤔 
But maybe we should put in some information/disable the add button in this case 🤔

maciaszczykm:
@kgroschoff If that is the case then yes, let's put this info somewhere. @lsviben can you take a look and provide us more details about it?

lsviben:
> @kgroschoff If that is the case then yes, let's put this info somewhere. @lsviben can you take a look and provide us more details about it?

Sure. 

What happens is that OPA Gatekeeper creates CRDs for Constraints based on Constraint Templates. We sync the CTs from the seed cluster to the user cluster.

The Gatekeeper-controller needs to be running to be able to generate and create the Constraint CRDs based on the Constraint Templates, if there is no machine deployment, its not running. 

When you create Constraints in the UI, they create a KKP Constraint in the seed-clusters user-cluster namespace. This gets synced to the user-clustr in the user-cluster-controller-manager. If there is no CRD, it fails. (but this should not be UI concern, it will retry until if finnaly succeds)

But where you could see errors is the LIST endpoint for Constraints. The LIST endpoint gets all the KKP Constraints in the  seed-clusters user-cluster namespace, and then based on them, lists the OPA Constraints in the user cluster, and gets their status(it contains the violations).

 So this LIST fails if the CRDs are not created yet(no machine dep -> no running gatekeeper -> no CRDs). Possibly it can be changed to ignore the fails and just not fill out the status in this case. This way we avoid the errors in UI

maciaszczykm:
@lsviben What if we would add a check for OPA Gatekeeper Controller instead of checking for Machine Deployment availability? For me it was very strange to see check for existing MD in a place that doesn't seem related at all.

lsviben:
> @lsviben What if we would add a check for OPA Gatekeeper Controller instead of checking for Machine Deployment availability? For me it was very strange to see check for existing MD in a place that doesn't seem related at all.
> 
> Btw., it doesn't work if initial MD has 0 replicas too:
> 
> Zrzut ekranu 2021-11-15 o 17 00 34
> 
> I will adjust the error message.

You mean check if Gatekeeper Controller is healthy? That would work. I agree that checking MD is not intuitive

maciaszczykm:
OPA needs like a lot of time to be ready but the errors seem to be gone. That's why I didn't change error messages to anything custom.

@kgroschoff PTAL

PR #3860, opened 2021-11-16
Reduce height of formfields

### LINKED ISSUE #3385: Input Height Investigation
Because of our high textfield density on places like cluster create, edit pages and more it would be very nice to check if we could reduce the height of the textfields. 

To test with we can try a height of 45px? Meaning the part of the textfield that has a border specifically, not counting the height that gets added by the label

kgroschoff:
Reduce height of formfields

height4

Fixes #3385

### Release note

```release-note
NONE
```


kgroschoff:
Sorry @maciaszczykm , forgot to add the master version, can you lgtm again, please?

Issue #3891, opened 2021-11-22
Update connect cluster dialog


maciaszczykm:
Update connect cluster dialog and add a wizard that allows to import/connect external clusters.

Design: https://github.com/kubermatic/dashboard/issues/3808

PR #3892, opened 2021-11-22
Update import cluster dialog & add GKE support

maciaszczykm:
Zrzut ekranu 2021-11-30 o 13 23 42
Zrzut ekranu 2021-12-2 o 11 10 16
Zrzut ekranu 2021-12-2 o 09 13 43

Closes https://github.com/kubermatic/dashboard/issues/3891.
Closes https://github.com/kubermatic/dashboard/issues/3888.

- GKE zones are not supported by API yet.
- API is throwing some errors when entering the page of newly imported external clusters. 

### Release note
```release-note
Added support for importing GKE clusters.
```


maciaszczykm:
/assign @kgroschoff @floreks 

This is ready for a review. Please do some tests too.

floreks:
We need a dark theme icon for the GKE @cschieder 
image

Also, the text in the GKE icon looks very small. It is hard to read.

floreks:
A couple of remarks about the dialog data flow:
- it is not possible to get back to the previous step and adjust the provided data
- in case of error (wrong SA provided) we end up in the last step with the spinner loading forever and an error notification without any way to fix this and we have to close and reopen the dialog starting from the beginning

image

cschieder:
Hi @floreks , 
I agree that this is unfortunate. The dialog was never intended to have a backtracking function so that part is fine. I would suggest adding some sort of validation to the credentials and only letting the user proceed if the data entered is valid, @maciaszczykm mentioned we probably need a new endpoint for that. Can we add that? @zreigz

floreks:
Does it mean that the `Back` button in the preset management dialog (admin settings) should be removed for consistency?

zreigz:
It's possible to add credential validation endpoint

floreks:
Could we simply try requesting the current endpoint and see if it returns an error before proceeding?

src/app/shared/components/add-external-cluster-dialog/steps/credentials/provider/gke/template.html
```diff
@@ -0,0 +1,26 @@
...
+limitations under the License.
+-->
+
+<form [formGroup]="form"
+      fxLayout="column">
+  <mat-form-field fxFlex>
+    <mat-label>Service Account</mat-label>
+    <input matInput
+           [formControlName]="Controls.ServiceAccount"
+           type="text"
```
kgroschoff, line 23: Shouldn't we use `type="password"` to "hide" the credentials?

kgroschoff (commented):
Agree on @floreks with the icon size, it is very small for GKE.

For the back thingy, I think this dialog is more or less a little wizard, so for me it eather needs validation or a back button 🤔

src/app/shared/components/add-external-cluster-dialog/steps/credentials/provider/gke/template.html
floreks: Isn't it base64 encoded?

kgroschoff: true, good point. 
just wondering if we want to handle it the same as in the wizard, where we never display any "real" characters for these kind of fields.

maciaszczykm: It is.

maciaszczykm:
Updated the logo part.

@kubermatic/sig-ui @kubermatic/sig-api Do we add the API endpoint for validation or reuse the current one (it can be seen as less clean)?

src/app/shared/components/add-external-cluster-dialog/steps/credentials/provider/gke/template.html
maciaszczykm: Changed, as we use "password" in wizard too. @kgroschoff good spot :)

floreks:
If it's not too much trouble I am fine with adding a new API endpoint. We can live without it otherwise.

cschieder:
@floreks about your question with the back button in preset create - you're right it would be inconsistent. I have some other thoughts about how to improve that dialog - once I have those written down as an issue we can include removing the back button there

zreigz:
OK I will add an endpoint to validate SA for the GKE provider

maciaszczykm:
@floreks PTAL

maciaszczykm:
@kgroschoff @floreks PTAL

PR #3940, opened 2021-12-09
Only display cni versions greater than current one

kgroschoff:
Only display cni versions greater than current one in dialog

### Release note

```release-note
NONE
```


kgroschoff:
Small improvement for the cni dialog, only display versions that are greater than the current one.
PTAL @maciaszczykm / @floreks

PR #3948, opened 2021-12-13
Add support for EventRateLimit admission plugin

ahmedwaleedmalik:
Add support for admission plugin `EventRateLimit`. Right now the only allowed limit type is `Namespace` for the plugin. 

Fixes https://github.com/kubermatic/edgematic/issues/77

- Wizard
wizard

- Wizard Info Tooltip
info_err

- Wizard Summary
wizard_summary

- Edit Cluster
edit_cluster

### Release note

```release-note
Support for EventRateLimit admission plugin
```


floreks:
I think we need some input from @cschieder about the design improvements. It doesn't match our current styling.

cschieder:
New feedback after talking with @floreks, he had some good points

- Change it to "Namespace Event Rate Limit" in the dropdown and "Namespace Event Rate Limit Configuration" in the headline
- Add a description what this is and what it does to the info icon tooltip

src/app/shared/components/event-rate-limit/style.scss
```diff
@@ -0,0 +1,25 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@use 'variables';
+
+form {
+  display: flex;
```
floreks, line 18: Same as before. Most of the time whole flex layout can be controlled by using `fxFlex`.

src/app/shared/components/number-stepper/style.scss
```diff
@@ -43,5 +43,5 @@ input[type='number']::-webkit-outer-spin-button {
 }
 
 .mat-form-field {
-  min-width: 150px;
+  min-width: 130px;
```
floreks, line 46: Do not change this globally as it will update our whole UI and many other places. You can create a custom class that when applied changes the width.

src/app/shared/components/event-rate-limit/template.html
```diff
@@ -0,0 +1,46 @@
...
+                       mode="errors"
+                       label="Burst"
+                       min="1"
+                       step="10"
+                       required>
+    </km-number-stepper>
+    <km-number-stepper [formControlName]="Controls.CacheSize"
+                       id="km-event-rate-limit-cache-size-input"
+                       mode="errors"
+                       label="CacheSize"
```
floreks, line 40: s/CacheSize/Cache Size

floreks (changes requested):
Few additional remarks from my side

floreks:
One more thing. Making the stepper width smaller makes the error overlap below text.

image

cschieder:
Are there some recommended values we can set there by default instead of just setting 1? Also we can change the error message to "Minimum value is 1" not sure if the "acceptable" is needed

ahmedwaleedmalik:
Most of the feedback is handled. The following is left at the moment:

> Making the stepper width smaller makes the error overlap the below text.

Yes, that is an issue that needs to be fixed. I reverted it to the old value of `min-width` but the issue still persists. I really don't want to opt for one input per line for this numeric input though.

> Change it to "Namespace Event Rate Limit" in the dropdown 

That's probably why I resorted to adding a `disabled` drop-down to show the limit type i.e. namespace. @floreks I'll have to add a special case [here](https://github.com/ahmedwaleedmalik/dashboard/blob/event-rate-limit-admission-plugin/src/app/shared/utils/admission-plugin-utils/admission-plugin-utils.ts#L27) for this. Are you okay with that?

embik:
> Change it to "Namespace Event Rate Limit" in the dropdown

A note on this, the API supports all types for the EventRateLimit plugin (namespace, cluster, user and source), but we decided to go a MVP route and only expose the namespace settings for now. But it's feasible to introduce more fields for it if customers/users are asking for it. I'm not sure I think renaming this is a good idea, given that the names are usually based on the admission plugin names, and this one _is_ named "EventRateLimit".

src/app/shared/components/cluster-summary/style.scss
```diff
@@ -69,3 +69,12 @@
     margin-left: -15px;
   }
 }
+
+.km-event-rate-limit {
```
floreks, line 73: As mentioned `km-` prefix is for the globally available styles only. Here is it a local style.

s/km-event-rate-limit/event-rate-limit

src/app/shared/components/cluster-summary/style.scss
```diff
@@ -69,3 +69,12 @@
...
+  display: flex;
+  font-size: 14px;
+}
+
+.km-event-rate-limit-values {
```
floreks, line 78: s/km-event-rate-limit-values/event-rate-limit-value

src/app/shared/components/cluster-summary/style.scss
```diff
@@ -69,3 +69,12 @@
     margin-left: -15px;
   }
 }
+
+.km-event-rate-limit {
+  display: flex;
```
floreks, line 74: Flex usage was also mentioned in another comment. Should be replaced with `fxFlex`.

src/app/shared/components/event-rate-limit/style.scss
```diff
@@ -0,0 +1,19 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+@use 'variables';
+
+.no-margin {
+  margin: 0 !important;
```
floreks, line 18: Do not use `!important` modifier. We are trying to get rid of it at all from the codebase as it often causes conflicts and makes it hard to debug.

floreks:
@embik @ahmedwaleedmalik since all values are actually supported and we only block other types now, I'd vote for having the disabled combobox with preselected `Namespace` type as initially proposed. I assume that after the release we will be enabling all options and it will be much simpler for us to do this.

floreks (changes requested):

ahmedwaleedmalik:
Feedback handled and screenshots updated.

floreks:
@cschieder any objections regarding the updated design?

cschieder:
Is there no way to put the number inputs next to each other like initially planned? @floreks

floreks:
There is always a way. The question is do we want error message to be 5 lines long due to very small width?

cschieder:
The error values that will barely happen because we pre-fill it with existing values? Also we can optimise wording and they can span the width of the + and - boxes as well

floreks:
> The error values that will barely happen because we pre-fill it with existing values? Also we can optimise wording and they can span the width of the + and - boxes as well

Even if they happen rarely they should still be clearly displayed in a readable format, without overlapping other elements like before.

We can't make them span to the full length of the component as they are handled by angular and are part of the input only. They will always match input width.

floreks:
@ahmedwaleedmalik as a result of discussion with @cschieder we'd like to get back to the single-line design.

Also, the number stepper error message can be changed

`Maximum value is ...` -> `Maximum is ...`
`Minimum values is ...` -> `Minimum is ...`
`Expected provided number to be an integer.` -> `Whole number expected`

If you need any help with updating width only for this single view let me know. I can do this.

ahmedwaleedmalik:
> If you need any help with updating width only for this single view let me know. I can do this.

@floreks I have managed the rest but the width for the form elements is still pending. Do let me know when you have a few minutes for that, I'll setup the environment at my end.

floreks:
### Updated

ahmedwaleedmalik:
@floreks thank you for taking care of this. Can you please attach a screenshot for the `edit cluster` view as well?

floreks:
@ahmedwaleedmalik I would if our dev would actually provision any cluster :joy: I tried creating multiple different clusters but they are permanently pending.

ahmedwaleedmalik:
/retest

Some tests for `kubevirt` are failing. Trying one more time before I try and resolve the issue since I'm expecting this to be a flake/temp issue.

ahmedwaleedmalik:
/retest

Tests for `kubevirt` are passing and the one that is failing seems like a flake; `tories/service-accounts.spec.ts > should add token`

PR #3961, opened 2021-12-15
Enable konnectivity by default

kgroschoff:
Enable konnectivity by default

Fixes https://github.com/kubermatic/kubermatic/issues/8129

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/wizard/step/cluster/component.ts
+++ b/src/app/wizard/step/cluster/component.ts
@@ -141,7 +141,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       [Controls.AuditPolicyPreset]: new FormControl(''),
       [Controls.UserSSHKeyAgent]: new FormControl(true),
       [Controls.OPAIntegration]: new FormControl(false),
-      [Controls.Konnectivity]: new FormControl(false),
+      [Controls.Konnectivity]: new FormControl(true),
       [Controls.MLALogging]: new FormControl(false),
       [Controls.MLAMonitoring]: new FormControl(false),
       [Controls.AdmissionPlugins]: new FormControl([]),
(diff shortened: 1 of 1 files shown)


kgroschoff:
PTAL @maciaszczykm / @floreks

src/app/wizard/step/cluster/component.ts
```diff
@@ -342,6 +342,8 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
     const cniPluginType = this.controlValue(Controls.CNIPlugin);
     const cniPluginVersion = this.controlValue(Controls.CNIPluginVersion);
     const cniPlugin = cniPluginType ? {type: cniPluginType, version: cniPluginVersion} : null;
+    const konnectivity = this.isKonnectivityEnabled ? this.controlValue(Controls.Konnectivity) : null;
```
maciaszczykm, line 345: Does this mean that if `this.isKonnectivityEnabled` will be false then even if my checkbox in the UI will be marked it will have no effect? If that is the case then it will be very confusing.

If Konnectivity is disabled in general then the checkbox should be hidden or disabled and empty.

kgroschoff: Took it from the issue `Note: in case that the Konnectivity feature gate in KKP is not enabled, this checkbox should not be shown at all (which should already be the case) and the value should not be set in the API call.`
Atm the checkbox isn't visible in the ui if `!this.isKonnectivityEnabled`

rastislavs: if the Konenctivity feature gate is disabled then I would prefer to not show the checkbox in the UI (make it hidden / not visible).
In that case it is not super-important whether the UI sends null or false to backend - both should work fine.

maciaszczykm: I didn't see it as this was done not in this pull request. In the end this line might not be needed but it's okay.

PR #3963, opened 2021-12-16
Update name validation to be more compliant with RFC1123 DNS names

embik:

The UI validation for MachineDeployment names was allowing uppercase letters, which would subsequently fail the MachineDeployment creation at the Kubernetes API in the backend. This PR tightens the validation logic for the MachineDeployment fields. Apart from uppercase letters, it also blocks dashes in the start and end of the name, as that's not valid under RFC1123 DNS names either (https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names).

In addition, as cluster names are used for name generation of MachineDeployments, they need tighter validation too. Otherwise, users can create a cluster that cannot generate MachineDeployments (if no name is passed for them).

The error message was tested across the various UI elements. I experimented with better wording, but that broke into multi-line output for editing resources, which looks ugly.

I was also looking at adding this to the KKP API on the backend side, but there's the major refactor of validation in KKP going on right now, so I'd like to postpone that until that PR has settled.

Fixes https://github.com/kubermatic/kubermatic/issues/8301

### Release note

```release-note
Cluster and MachineDeployment names are validated to be lowercase, alphanumerical and with dashes in between
```

### DIFF
--- a/src/app/cluster/cluster-details/edit-cluster/component.ts
+++ b/src/app/cluster/cluster-details/edit-cluster/component.ts
@@ -105,7 +105,7 @@ export class EditClusterComponent implements OnInit, OnDestroy {
       [Controls.Name]: new FormControl(this.cluster.name, [
         Validators.required,
         Validators.minLength(this._nameMinLen),
-        Validators.pattern('[a-zA-Z0-9-]*'),
+        Validators.pattern('[a-z0-9]+[a-z0-9-]*[a-z0-9]+'),
       ]),
       [Controls.ContainerRuntime]: new FormControl(this.cluster.spec.containerRuntime || ContainerRuntime.Containerd, [
         Validators.required,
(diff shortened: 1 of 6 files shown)


ahmedwaleedmalik:
Can you please move the regex to `src/app/shared/constants/constants.ts` and add a const for it? Since it's shared between multiple components and specifying the string value is more error-prone.

embik:
@ahmedwaleedmalik I can try. This is my first time ever touching any frontend code, so I was happy I even got as far as finding the regexes. I'll need a bit figuring out how to do constants, import from a different module, etc.

ahmedwaleedmalik:
You have already done a remarkable job, it was just a small nit ;)

floreks:
You can create an issue to refactor i.e. patterns to constants (possibly other shared things also). I'd handle it a bit differently and probably create either namespace or static class to gather them all together.

PS. Regular constants do not require any magic to import. You simply `export const x` in file A and then import it in file B.

PR #3971, opened 2021-12-20
Admin Rule Groups

### LINKED ISSUE #3873: MLA: Add a Tab in admin panel for KKP admins to manage rule groups
Add a tab in admin panel "MLA", and this will be used for KKP admins to manage rule group, similar to the rule group UI in cluster’s “Monitoring, Logging & Alerting” tab
For API endpoints, we can use following endpoints:
```
GET /api/v2/seeds/{seed_name}/rulegroups
GET /api/v2/seeds/{seed_name}/rulegroups/{rulegroup_id}
POST /api/v2/seeds/{seed_name}/rulegroups
PUT  /api/v2/seeds/{seed_name}/rulegroups/{rulegroup_id}
DELETE /api/v2/seeds/{seed_name}/rulegroups/{rulegroup_id}
```

kgroschoff:
Adds new view "Rule Groups" to Admin Settings

Fixes #3873

### Release note

```release-note
Admins can define default Rule Groups in Admin Settings
```


floreks:
There is some kind of race condition as the `rule groups are not always loading.

image

image

kgroschoff:
Sorry @floreks, missed your comment.
I can't reproduce the issue, but i've moved the "get Admin Rule Groups" part, because I can imagine, that this is an issue.
Can you verify if this was the problem?

I don't know if one of you is working until the end of the year, but just in case yes: Would be great if one of you can have a look. I will be back in January, so if you want to hijack the PR, it'll be fine with me. - cc @maciaszczykm

kgroschoff:
@floreks @maciaszczykm PTAL

src/app/settings/admin/rule-groups/component.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    this._userService.currentUserSettings.pipe(takeUntil(this._unsubscribe)).subscribe(settings => {
+      this.settings = settings;
+      this.paginator.pageSize = settings.itemsPerPage;
+      this.dataSource.paginator = this.paginator; // Force refresh.
+    });
+
+    this._datacenterService.seeds.pipe(takeUntil(this._unsubscribe)).subscribe(seeds => {
+      this.seeds = seeds;
+      this._getAdminRuleGroups();
```
floreks, line 77: This change creates a new issue. It will cause a subscription leak and create lots of the same subscriptions every time seeds get updated.

A pipe should be used here possibly with a `combineLatest` to control the rule groups.

kgroschoff: Understood the problem, but unfortunately my attemps are failing atm. Eather code is not correct or it makes it even worse, but no idea how to fix it. :/ Could you help out, please?

src/app/settings/admin/rule-groups/component.ts
```diff
@@ -0,0 +1,200 @@
...
+      this.adminRuleGroups.length > this.paginator.pageSize
+    );
+  }
+
+  hasNoData(): boolean {
+    return _.isEmpty(this.adminRuleGroups);
+  }
+
+  getName(data: string): string {
+    return MLAUtils.getRuleGroupName(data);
```
floreks, line 104: This is a bit frightening, to be honest. In order to fill out a single name column, we will have to decode every string data object to YAML, transform it to JSON, cast it to object, and get the name.

This definitely requires an API change. I see that the `RuleGroup` CRD has a name that we should be using instead of this workaround.

image

src/app/settings/admin/rule-groups/component.ts
floreks: Optimized

kgroschoff:
/hold
we need to wait for the backend pr to be merged first.

kgroschoff:
/hold cancel

@floreks PTAL, I've added the name field for the rulegroup spec.

src/app/settings/admin/rule-groups/component.ts
```diff
@@ -0,0 +1,186 @@
...
+    this._unsubscribe.next();
+    this._unsubscribe.complete();
+  }
+
+  isPaginatorVisible(): boolean {
+    return (
+      this.adminRuleGroups &&
+      this.adminRuleGroups.length > 0 &&
+      this.paginator &&
+      this.adminRuleGroups.length > this.paginator.pageSize
```
maciaszczykm, line 113: `this.adminRuleGroups?.length > this.paginator?.pageSize` seems to cover it all assuming that page size is at least equal to 0.

src/app/settings/admin/rule-groups/template.html
```diff
@@ -0,0 +1,176 @@
...
+              </mat-option>
+            </mat-select>
+          </mat-form-field>
+        </div>
+        <div fxFlex></div>
+        <button mat-flat-button
+                color="quaternary"
+                fxLayoutAlign="center center"
+                (click)="add()">
+          <i class="km-icon-mask km-icon-add icon-add"></i>
```
maciaszczykm, line 94: Are both `km-icon-add` and `icon-add` needed?

src/app/settings/admin/rule-groups/component.ts
kgroschoff: yep, sounds good to me 👍

src/app/settings/admin/rule-groups/template.html
kgroschoff: For some reason the add icon is having too much margin on this button, so I needed to add some styling to it. I can also directly use `.km-icon-add` for it, but I thought we wanted to use the km-prefix only for global styling 😅

floreks: True, but you can simply apply some additional styling on the component level to the global class. It will not override anything for other components.

kgroschoff: Alright. Then - changed :)

PR #3980, opened 2021-12-21
Support for OSM in dashboard

ahmedwaleedmalik:
Support to enable experimental feature `Operating System Manager` in the dashboard.

I have intentionally not added support to edit this field after cluster creation i.e. in `edit-cluster` dialog.

osm

### Release note

```release-note
Support to enable experimental feature OSM
```


maciaszczykm:
I'd like to get an opinion on the design here. I'd change "[Experimental]" text to some badge like we have for external clusters or labels. WDYT @cschieder?

maciaszczykm:
We can leave the final design update for later, let's do something like these:

Zrzut ekranu 2021-12-22 o 09 23 44
Zrzut ekranu 2021-12-22 o 09 23 26

It's either the first one, which is easy to notice, or the second one where the info can go into tooltip.

ahmedwaleedmalik:
How about the first one with a tooltip? The thing is we do not have some info up there and the tooltip seems like the appropriate place for that. I like the first one in general because it's more noticable.

maciaszczykm:
Fine for me.

ahmedwaleedmalik:
Updated UI

updates_osm

maciaszczykm:
LGTM but you have to fix static check.

src/app/wizard/step/cluster/style.scss
```diff
@@ -22,6 +22,10 @@ $toggle-height: 45px;
   margin: 4px 8px;
 }
 
+.km-label-primary {
```
maciaszczykm, line 25: One last thing is to check if it can be removed and if you can use `fxLayoutGap="3px"` in the template instead.

ahmedwaleedmalik: I did try that but it doesn't work. Because `margin` is inherited from `km-label-primary` and it ruins the layout.

PR #4071, opened 2022-01-14
Remove unused code

maciaszczykm:
Removes unused code and simplifies some methods.

### Release note
```release-note
NONE
```


maciaszczykm:
@kgroschoff @floreks It looks like full tests are not passing but this is not connected to this change. Can you review?

PR #4084, opened 2022-01-18
Fix mla tab

kgroschoff:
Fixes mla tab, as the current `isLoading()` function basically every time returned true. Especially in case there aren't any default rule groups it is not possible that this tab ever gets enabled.

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/details/cluster/mla/component.ts
+++ b/src/app/cluster/details/cluster/mla/component.ts
@@ -31,6 +31,6 @@ export class MLAComponent {
   @Input() addons: Addon[];
 
   isLoadingData(): boolean {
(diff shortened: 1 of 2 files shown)


kgroschoff:
PTAL @maciaszczykm @floreks

src/app/cluster/details/cluster/mla/component.ts
```diff
@@ -31,6 +31,6 @@ export class MLAComponent {
   @Input() addons: Addon[];
 
   isLoadingData(): boolean {
-    return _.isEmpty(this.alertmanagerConfig) || _.isEmpty(this.ruleGroups) || !this.isClusterRunning;
+    return !this.isClusterRunning;
```
maciaszczykm, line 34: Please use it directly in the template instead.

kgroschoff: 👍

Issue #4042, opened 2022-01-10
Review notification messages and ensure that some of the actions trigger instant refresh 


maciaszczykm:
**What happened?**: For example, when removing machine deployment user gets notification that "it was removed" but in the list it is still green and running. Status switches to red and deleting in a few seconds.

**What should happen?**:
- status should be updated sooner (in the subscribe we can trigger resource refresh)
- notification messages should be more accurate (i.e. deleting instead of deleted)

**Environment**
- UI Version: master
- API Version: master
- Domain: dev
- Others:

PR #4096, opened 2022-01-19
Minor fixes and notification updates

maciaszczykm:
See commits.

Feel free to post suggestions related to the notification messages. It's an attempt to be more consistent but I think we still need to meet and create some patterns for the messages and them use them.

A step towards https://github.com/kubermatic/dashboard/issues/4042.

https://github.com/kubermatic/dashboard/issues/4097 will cover the refesh part as it is a bit bigger change.

### Release note
```release-note
NONE
```


src/app/backup/list/restore/component.ts
```diff
@@ -127,7 +127,7 @@ export class RestoreListComponent implements OnInit, OnDestroy {
         )
       )
       .subscribe(_ => {
-        this._notificationService.success(`Successfully deleted restore object ${restore.name}`);
+        this._notificationService.success(`Deleted restore object ${restore.name}`);
```
kgroschoff, line 130: For e.g. automatic backups we use `Deleted the ${backup.name} automatic backup` (so basically Action-Name-Resource)
and here we use `Deleted restore object ${restore.name}` (Action-Resource-Name)

I guess we should think about one order in general, shouldn't we?
[Haven't checked on all yet, but was the first one where I noticed it.]

kgroschoff: cc @cschieder

maciaszczykm: IMO the first form is better.

cschieder: Iirc we use action-name-resource basically everywhere, like kristin said. another example: "Delete silly-clustername cluster" Imo it reads well.

maciaszczykm:
@kgroschoff @floreks PTAL

PR #4099, opened 2022-01-19
Enhance wording + add warnings for default destinations

kgroschoff:
Today during the day we noticed, that the default destinations need some more warnings and the wording could be optimised.

If a destination is marked as default already and user wants to unset the checkbox following warning will appear:
delete-warning

If a user wants to select a new destination as default (no matter if via add or edit dialog), following warning will appear:
overwrite-warning

### Release note

```release-note
NONE
```


src/app/settings/admin/bucket-settings/destinations/destination-dialog/component.ts
```diff
@@ -124,4 +134,8 @@ export class DestinationDialog implements OnInit, OnDestroy {
         return 'Destination was successfully edited';
     }
   }
+
+  private _isDefault(): boolean {
+    return !!this.currentDefault && this.currentDefault === this.data?.destination?.destinationName;
```
maciaszczykm, line 139: Is `!!this.currentDefault` really needed?

kgroschoff:
@maciaszczykm can you lgtm again? there was a merge conflict

PR #4110, opened 2022-01-20
Update provider/ssh keys fields key display

### LINKED ISSUE #4107: Labels missing in Cluster Detail view?

On the cluster detail page, the provider and the SSH keys are not labelled, whereas everything else is. This looks ugly for people like me, who zoom in a bit, as the horizontal alignment gets disrupted by the seemingly out of place provider logo and the free floating, undescribed "No keys assigned".

screenshot-2022-01-20-112154
screenshot-2022-01-20-112206

Add labels to these 2 blocks like every other block (ID, Created, Seed, Region, ...).

I'm using Firefox on Ubuntu. Tested on dev.

floreks:
Added key labels to the provider and ssh keys fields.

Fixes #4107

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/details/cluster/template.html
+++ b/src/app/cluster/details/cluster/template.html
@@ -163,13 +163,15 @@ limitations under the License.
           <div key>Container Runtime</div>
           <div value>{{cluster.spec.containerRuntime}}</div>
         </km-property>
-        <km-property mode="value">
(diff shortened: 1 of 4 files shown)


floreks:
cc @xrstf @cschieder

PR #4121, opened 2022-01-24
Move update indicator

### LINKED ISSUE #4113: Consistent placement of "updates available" indicator
The placement of the "updates available" indicator of Control Plane Version and CNI Plugin Version is inconsistent. The former is on the Kubernetes logo and the latter on the right hand side of the virtual drop down field.

image

Place both "update alvailable" indicators on the right hand side of the virual drop down field.

kgroschoff:
As decided in slack, we wanted to move the update available indicator next to the dropdown arrow.
Also slightly reduced the width of the whole version picker, as we also shortened the name to "Control Plane" and there is plenty of space left.

Fixes #4113

ua4

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/details/shared/version-picker/style.scss
+++ b/src/app/cluster/details/shared/version-picker/style.scss
@@ -64,6 +64,10 @@
     .mat-form-field {
       font-size: variables.$font-size-subhead;
       max-width: 140px;
+
+      &.cluster-external {
+        max-width: 200px;
+      }
     }
   }
 
--- a/src/app/cluster/details/shared/version-picker/style.scss
+++ b/src/app/cluster/details/shared/version-picker/style.scss
@@ -41,20 +41,29 @@
 
         margin: 0 auto;
       }
+    }
 
-      .km-icon-update-available,
-      .km-icon-warning,
-      .km-icon-error {
-        @include mixins.size(16px);
+    .update-available-icon {
+      @include mixins.size(16px);
 
-        margin: 0 0 12px 22px;
-        position: absolute;
-      }
+      position: absolute;
+      right: 28px;
+      top: -9px;
+    }
+
+    .warning-icon,
+    .error-icon {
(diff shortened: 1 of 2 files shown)


kgroschoff:
PTAL @maciaszczykm @floreks

maciaszczykm:
@kgroschoff Have you checked it with external clusters? I remember GKE has pretty long version.

kgroschoff:
good hint, will check 👍

kgroschoff:
@maciaszczykm Added a class for external clusters, as size differs a lot and I think the wide size on "normal" clusters looks strange. PTAL :)

src/app/cluster/details/shared/version-picker/style.scss
```diff
@@ -41,20 +41,33 @@
...
+    }
+
+    .warning-icon,
+    .error-icon {
+      @include mixins.size(16px);
+
+      margin-right: 0;
+      position: absolute;
+      right: 28px;
+      top: -13px;
```
maciaszczykm, line 61: One last thing is if we can style this without position absolute? It should be possible as right now matSuffix is used.

src/app/cluster/details/shared/version-picker/style.scss
```diff
@@ -41,20 +41,33 @@
...
-        @include mixins.size(16px);
+    .update-available-icon {
+      @include mixins.size(16px);
 
-        margin: 0 0 12px 22px;
-        position: absolute;
-      }
+      position: absolute;
+      right: 28px;
+      top: -9px;
```
maciaszczykm, line 51: As above.

src/app/cluster/details/shared/version-picker/style.scss
kgroschoff: without position absolute the icon will always be placed on the right of the dropdown arrow :/

PR #4123, opened 2022-01-24
Update event grouping and more information section logic

### LINKED ISSUE #4116: Cluster Event List - Improvements
image

Two points
1) when I last talked with @toschneck he told me how the grouping of the notifications is rather slow or sometimes doesn't seem to happen at all, then we have the same event in the list a few times
2) resource id and cluster should probably not have a line break

floreks:
- Additional custom events grouping by the message/type/id
- Expand additional information section by default when cluster is not in running state

Fixes #4116

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/details/cluster/component.ts
+++ b/src/app/cluster/details/cluster/component.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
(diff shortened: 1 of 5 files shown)


floreks:
I'd just override this as it's a weird flake.

xrstf:
Shall I?

PR #4131, opened 2022-01-26
Migrate to kubermatic.k8c.io

xrstf:
This PR updates the API groups used for e2e tests to the new groups that are soon going to be used by KKP.

Currently, before https://github.com/kubermatic/kubermatic/pull/8783 is merged, this PR will always fail.

This is part of kubermatic/kubermatic#6471

### Release note
```release-note
NONE
```

### DIFF
--- a/hack/e2e/fixtures/seed.yaml
+++ b/hack/e2e/fixtures/seed.yaml
@@ -40,7 +40,7 @@ spec:
       location: Frankfurt
       country: DE
       spec:
-         bringyourown: {}
+        bringyourown: {}
     alibaba-eu-central-1a:
       location: Frankfurt
       country: DE
@@ -59,6 +59,7 @@ spec:
       spec:
         hetzner:
           datacenter: nbg1-dc3
+          network: kubermatic-e2e
     vsphere-ger:
       location: Hamburg
       country: DE
--- a/hack/e2e/fixtures/seed.yaml
+++ b/hack/e2e/fixtures/seed.yaml
@@ -22,7 +22,7 @@ data:
 
 ---
 kind: Seed
-apiVersion: kubermatic.k8s.io/v1
+apiVersion: kubermatic.k8c.io/v1
 metadata:
   name: __SEED_NAME__
   namespace: kubermatic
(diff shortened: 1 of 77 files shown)


maciaszczykm:
Updated names ❤️

hack/e2e/fixtures/seed.yaml
```diff
@@ -116,23 +117,24 @@ spec:
...
-          dns_servers:
+          authURL: https://api.cbk.cloud.syseleven.net:5000/v3
+          availabilityZone: dbl1
+          dnsServers:
           - 37.123.105.116
           - 37.123.105.117
-          enforce_floating_ip: true
-          ignore_volume_az: false
+          enforceFloatingIP: true
+          ignoreVolumeAZ: false
```
maciaszczykm, line 132: I can see these values are in https://github.com/kubermatic/dashboard/blob/master/cypress/fixtures/datacenters.json too.

hack/e2e/fixtures/seed.yaml
```diff
@@ -84,7 +85,7 @@ spec:
       spec:
         gcp:
           region: europe-west3
-          zone_suffixes:
+          zoneSuffixes:
```
maciaszczykm, line 88: It's also part of https://github.com/kubermatic/dashboard/blob/master/src/app/shared/entity/datacenter.ts.

maciaszczykm:
Not all sources and fixtures are adjusted at the moment. The full tests can fail due to some provider issues but they also provide info what is not working, so let's not ignore them completely.

hack/e2e/fixtures/seed.yaml
xrstf: What are the files like datacenter.ts actually based on? The REST API in KKP still has the exact same types. Unless you use the CRDs 1:1 in the dashboard, I am not sure that the entity files need to be changed here.

maciaszczykm: Ah, okay. So it's not updated in the API. Then ignore this comment.

xrstf: Yeah we only changed the CRDs but kept the API _mostly_ the same. There are a few changes like https://github.com/kubermatic/dashboard/pull/4131/commits/dab8d7823e55d87326e3efa63908b758c7d3fad5 (and this one needs a bit of you guys' love, cause I don't know how to properly handle a string[] for the Presets). ^^

floreks:
Someone finally did it and got rid of all those snake_case relics :heart: Thank you :slightly_smiling_face:

xrstf:
Thank @kron4eg, he spearheaded the cleanup for our tags :)

xrstf:
@maciaszczykm  /approve instead of /approved pls ^^

PR #4149, opened 2022-02-02
Review and update the notification messages

### LINKED ISSUE #4042: Review notification messages and ensure that some of the actions trigger instant refresh

### LINKED ISSUE #4161: Subscription leak on the service account tokens
The view requires complete refactoring to fix the issue. Every time user clicks on the SA to open a list of tokens a new subscription is created without closing the previous one. Since they all use refresh timers and are never closed we end up with the leak.

floreks:
- Updated in general the way we construct the notification messages to `[action] the [resource name] [resource type]`
- Updated all async delete/remove messages to use deleting/removing form as we cannot guarantee that the resource will be actually deleted after showing the notification.
- Fixed an issue with the project switch on cluster list view where the reload animation was not triggered.
- Fixed an issue with the subscription leak on the SA list view and refactored the logic.

API part: https://github.com/kubermatic/kubermatic/pull/8908

#### ~~TODO~~
~~Finish reviewing the resource refresh after showing the notification. It also requires a `deletionTimestamp` handling. Some of the resources do not provide the proper `deletionTimestamp` as the API does not return it. It will require further API review and update.~~

Closes #4042
Fixes #4161

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/shared/entity/backup.ts
+++ b/src/app/shared/entity/backup.ts
@@ -94,8 +94,10 @@ type BackupStatusPhase = string;
 export const BackupStatusPhaseCompleted = 'Completed';
 
 export class EtcdRestore {
+  // ObjectMeta
+  creationTimestamp?: Date;
+  deletionTimestamp?: Date;
   name?: string;
-
   spec: EtcdRestoreSpec;
   status: EtcdRestoreStatus;
 }
--- a/src/app/shared/entity/backup.ts
+++ b/src/app/shared/entity/backup.ts
@@ -94,7 +94,6 @@ type BackupStatusPhase = string;
 export const BackupStatusPhaseCompleted = 'Completed';
 
 export class EtcdRestore {
-  // ObjectMeta
   creationTimestamp?: Date;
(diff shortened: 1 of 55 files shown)


src/app/shared/entity/backup.ts
```diff
@@ -94,8 +94,10 @@ type BackupStatusPhase = string;
 export const BackupStatusPhaseCompleted = 'Completed';
 
 export class EtcdRestore {
+  // ObjectMeta
```
maciaszczykm, line 97: Let's remove it or if you prefer not to let's make it more clear that it's for the timestamps and the name.

PR #4158, opened 2022-02-08
Use textarea instead of input field

### LINKED ISSUE #3946: Improve input fields for bigger data structures
Update the type of the input fields for things like GKE service account or KubeVirt kubeconfig (containing long values) to be text areas instead of single line inputs.

Additionally it would be good to detect if data is base64 encoded and encode it if that is required.

kgroschoff:
Use textarea instead of input field for Kubevirt Kubeconfig and GCP/GKE Service Account

Fixes #3946

### Release note

```release-note
NONE
```


kgroschoff:
PTAL @maciaszczykm @floreks 
- Chiara said that 3 is a good row size.
- I've only added base64 check for GCP/GKE atm, as it seems like backend is fine with kubevirt even if it's not base64. However, I'm also fine with adding the check to kubevirt if needed.

kgroschoff:
PTAL @maciaszczykm @floreks

kgroschoff:
@floreks good catch. Initally the form value is null, therefor the isValid() check for base64 is false and null will be encoded :trollface: 
Added an additional check.

src/app/settings/admin/presets/dialog/steps/settings/provider/gcp/component.ts
```diff
@@ -74,13 +74,19 @@ export class GCPSettingsComponent extends BaseFormValidator implements OnInit, O
...
       network: this.form.get(Controls.Network).value,
       subnetwork: this.form.get(Controls.Subnetwork).value,
     } as GCPPresetSpec;
   }
+
+  private get _serviceAccountValue(): string {
+    let serviceAccountValue = this.form.get(Controls.ServiceAccount).value;
+    if (!!serviceAccountValue && !isValid(this.form.get(Controls.ServiceAccount).value)) {
+      serviceAccountValue = encode(this.form.get(Controls.ServiceAccount).value);
+    }
```
floreks, line 88: Is there a reason why the variable is assigned at the first line and then it is not being really used in the `if` and during the `encode`.

kgroschoff: if Monday counts... otherwise no. Fixed.

Issue #3937, opened 2021-12-09
Update KubeVirt creation wizard for UI


mfranczy:
We are in process of changing KubeVirt Settings to be more friendly and usable. 
A new diagram has been prepared that shows how a new form should look like:

![new UI.drawio.png](https://images.zenhubusercontent.com/5dc5412ea240680001cc35aa/17568a83-c360-4439-a481-d33ad5fc7b05)

Additionaly we prepared two new endpoints to list storage classes and virtual machine flavors from KubeVirt infrastructure clusters.
Those are:
* list virtual machine presets https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/routes_v2.go#L3575-L3597 and https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/routes_v2.go#L5168-L5188 (note that in the endpoint we don't use virtual machine preset name but virtual machine flavor, it's because we prepare for a new feature from KubeVirt)
* list storage classes https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/routes_v2.go#L3599-L3621 and https://github.com/kubermatic/kubermatic/blob/master/pkg/handler/v2/routes_v2.go#L5190-L5210

More details on the "Operating System" field, acceptable values are an http url or `namespace/name` format. For instance when user wants to clone a disk he should put a value `default/some-pvc-with-disk`, it always must be a pair of namespace with name.

hdurand0710:
Additional needs for **VM affinity**
References: https://github.com/kubermatic/kubermatic/issues/8742


![VM affinity UI.drawio.png](https://images.zenhubusercontent.com/613f8d176dae552873cbbca8/fd275ee2-fafb-4e59-b5ed-4eeee6a83b39)

PR #4178, opened 2022-02-11
Update KubeVirt machine deployment settings

maciaszczykm:
Zrzut ekranu 2022-02-23 o 16 32 12
Zrzut ekranu 2022-02-23 o 16 35 05
Zrzut ekranu 2022-02-23 o 16 27 37

Closes https://github.com/kubermatic/dashboard/issues/3937.

- `KubeVirtVMInstancePreset` contains at the moment only name, namespace and marshaled spec, so CPUs and memory inputs will not be updated with the values from it. It would be complex anyways as I have noticed that some of the flavors do not contain both values.
- There is assumption in here and also in the API that all the flavors are stored in the default namespace. To change that first API needs to accept parameter other than `flavorName` in `KubeVirtNodeSpec`.
- `KubeVirtVMInstancePreset` will be automatically selected once the API will send its response back. Let me know if you would like to change this behavior.

### Release note
```release-note
Update KubeVirt machine deployment settings.
```


src/app/node-data/basic/provider/kubevirt/flavor-details/component.ts
```diff
@@ -0,0 +1,39 @@
...
+  styleUrls: ['./style.scss'],
+})
+export class FlavorDetailsDialogComponent {
+  constructor(@Inject(MAT_DIALOG_DATA) private _data: DialogDataInput) {}
+
+  get flavor(): KubeVirtVMInstancePreset {
+    return this._data.flavor;
+  }
+
+  get spec(): any {
```
maciaszczykm, line 36: `any` is used because we don't know the exact format of the unmarshaled object. It's internal API structure.

PR #4179, opened 2022-02-14
Improve Nutanix wizard form

maciaszczykm:
Improve Nutanix wizard form by adding debounce and watching not only for basic form changes but also extended one. It's now similar to what we have in AWS wizard.

### Release note
```release-note
NONE
```


kgroschoff (commented):
/lgtm

PR #4194, opened 2022-02-18
Disable OIDC kubeconfig flag if feature gate OIDCKubeCfgEndpoint is disabled

### LINKED ISSUE #4193: Block Enable OIDC Kubeconfig flag when FG not present
The `EnableOIDCKubeconfig` flag is settable in the UI at the following path: `admin panel > Interface > Enable OIDC Kubeconfig`; such a flag is stored in the `kubermaticSetting` CRD. When this option is enabled, and the user wants to download the kubeconfig for a specific cluster, the following path is used: "/api/v1/kubeconfig?cluster_id=xxx&project_id=yyy&user_id=zzz". To support such functionality, a Feature Gates must be set in the `kubermaticConfiguration` resource: `OIDCKubeCfgEndpoint`. If this Feature Gates isn't enabled, the path mentioned above isn't loaded in the router, leading to a 404 when trying to download the Kubeconfig.

The `admin panel > Interface > Enable OIDC Kubeconfig` flag shouldn't be settable if the FG `OIDCKubeCfgEndpoint`  isn't enabled .

ahmedwaleedmalik:

Disable OIDC kubeconfig flag if feature gate OIDCKubeCfgEndpoint is disabled.

Fixes #4193

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/core/services/feature-gate.ts
+++ b/src/app/core/services/feature-gate.ts
@@ -21,6 +21,7 @@ import {catchError, shareReplay, switchMap} from 'rxjs/operators';
 
 export interface FeatureGates {
   konnectivityService?: boolean;
+  OIDCKubeCfgEndpoint?: boolean;
 }
 
 @Injectable({
--- a/src/app/core/services/feature-gate.ts
+++ b/src/app/core/services/feature-gate.ts
@@ -21,7 +21,7 @@ import {catchError, shareReplay, switchMap} from 'rxjs/operators';
 
 export interface FeatureGates {
   konnectivityService?: boolean;
-  OIDCKubeCfgEndpoint?: boolean;
+  oidcKubeCfgEndpoint?: boolean;
 }
 
 @Injectable({
--- a/src/app/settings/admin/interface/template.html
+++ b/src/app/settings/admin/interface/template.html
@@ -92,10 +92,12 @@ limitations under the License.
              class="entry-label">
           <span>Enable OIDC Kubeconfig</span>
           <div class="km-icon-info km-pointer"
-               matTooltip="Use OIDC provider as a proxy for kubeconfig download."></div>
+               matTooltip="Use OIDC provider as a proxy for kubeconfig download. Requires OIDCKubeCfgEndpoint feature gate to be enabled."></div>
         </div>
         <mat-checkbox [(ngModel)]="settings.enableOIDCKubeconfig"
+                      [disabled]="!isOIDCKubeCfgEndpointEnabled"
                       (change)="onSettingsChange()"
+                      [matTooltip]="!isOIDCKubeCfgEndpointEnabled ? 'OIDCKubeCfgEndpoint feature gates needs to enabled.' : null"
                       id="km-enable-oidc-setting"></mat-checkbox>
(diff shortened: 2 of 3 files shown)


src/app/core/services/feature-gate.ts
```diff
@@ -21,6 +21,7 @@ import {catchError, shareReplay, switchMap} from 'rxjs/operators';
 
 export interface FeatureGates {
   konnectivityService?: boolean;
+  OIDCKubeCfgEndpoint?: boolean;
```
floreks, line 24: Are you sure that the API has been updated to return this?

ahmedwaleedmalik: https://github.com/kubermatic/kubermatic/pull/9034/files#diff-997ca39511bbd531bc6ed68d834e78ce8194439f367a1cf0130e0748d3033f15R647

floreks: For the future please link any blocking external PRs and hold the PR so that we won't merge anything too early.

ahmedwaleedmalik:
/hold
Blocked by https://github.com/kubermatic/kubermatic/pull/9034

src/app/settings/admin/interface/template.html
```diff
@@ -92,10 +92,12 @@
              class="entry-label">
           <span>Enable OIDC Kubeconfig</span>
           <div class="km-icon-info km-pointer"
-               matTooltip="Use OIDC provider as a proxy for kubeconfig download."></div>
+               matTooltip="Use OIDC provider as a proxy for kubeconfig download. Requires OIDCKubeCfgEndpoint feature gate to be enabled."></div>
         </div>
         <mat-checkbox [(ngModel)]="settings.enableOIDCKubeconfig"
+                      [disabled]="!isOIDCKubeCfgEndpointEnabled"
                       (change)="onSettingsChange()"
+                      [matTooltip]="!isOIDCKubeCfgEndpointEnabled ? 'OIDCKubeCfgEndpoint feature gates needs to enabled.' : null"
```
maciaszczykm, line 100: It may be not clear where the flag can be found.

maciaszczykm: Still I am not sure about using internal names like `OIDCKubeCfgEndpoint` or `KubermaticConfiguration` in the UI. Not all end users will know them, it's clear for devs but it might be not for all.

maciaszczykm: @cschieder what do you think, how it should look like in your opinion?

ahmedwaleedmalik: What'd you suggest then? Just keep it disabled without any additional info?

cschieder: Could we call it by some name that describes it's functionality, in general maybe? (both here and in the documentation) And then say "Feature is disabled. Check out our documentation about xxx to learn more." Or the like?

ahmedwaleedmalik: Hmm, will [this](https://docs.kubermatic.com/kubermatic/master/tutorials_howtos/oidc_provider_configuration/share-_clusters_via_delegated_oidc_authentication/) be adequate? Thought we had support for anchor tags in our docs but I think we don't 🤷‍♂️

floreks: I am not sure if the link inside the tooltip will actually work.

PR #4195, opened 2022-02-18
Fix invalid marshaling for requiredEmails in preset creation

ahmedwaleedmalik:

Fix invalid marshaling for `requiredEmails` in preset creation.

Screenshot 2022-02-21 at 12 47 57 PM
Screenshot 2022-02-21 at 12 48 07 PM

### Release note

```release-note
NONE
```


ahmedwaleedmalik:
> This is a workaround. We should add proper support for an array of emails, not dump a raw input string as a single element array.

My understanding from the existing code base was that it's supposed to only take one value. I'll update this to an array.

floreks:
Since the API was updated to expect an array then we should also do it.

src/app/settings/admin/presets/dialog/steps/preset/template.html
```diff
@@ -30,18 +30,33 @@
...
-  <mat-form-field fxFlex>
-    <mat-label>Restrict Preset Group to Domain</mat-label>
-    <input matInput
-           [formControlName]="controls.Domain"
-           [pattern]="domainRegex"
-           type="text"
-           title="domain"
-           autocomplete="off">
-    <mat-error *ngIf="form.get(controls.Domain).hasError( 'pattern')">
+  <mat-form-field class="domains-chip-list"
```
floreks, line 33: I would reuse a `km-tag-list` and adjust it if needed to accept more parameters. This is basically a copy of the code that is already there. We can have a single generic shared component.

src/app/settings/admin/presets/module.ts
```diff
@@ -13,12 +13,13 @@
...
 
 import {NgModule} from '@angular/core';
+import {MatChipsModule} from '@angular/material/chips';
 import {PresetListComponent} from '@app/settings/admin/presets/component';
 import {AdminSettingsPresetsRoutingModule} from '@app/settings/admin/presets/routing';
 import {SharedModule} from '@shared/module';
 
 @NgModule({
-  imports: [SharedModule, AdminSettingsPresetsRoutingModule],
+  imports: [SharedModule, AdminSettingsPresetsRoutingModule, MatChipsModule],
```
floreks, line 22: Why is it imported here? `SharedModule` holds all shared modules, so that we don't have to add them to any other component module.

src/app/settings/admin/presets/dialog/steps/preset/template.html
ahmedwaleedmalik: Maybe that is something that we can do in a follow-up, separate PR? Because we are currently using `mat-chip-list` at other places as well. That can be refactored in one go to use a more generic shared component.

src/app/settings/admin/presets/module.ts
ahmedwaleedmalik: Didn't know that we were already using this.

src/app/settings/admin/presets/dialog/steps/preset/template.html
ahmedwaleedmalik: Btw that stuff needs to be refactored anyways since right now if you try to create a data-center, it doesn't perform any validations for emails.

ahmedwaleedmalik: Updated

ahmedwaleedmalik:
@floreks After the refactoring the name `tag-list` doesn't really make sense here. I think we should come up with a more generic name that de-couples it from the term `tag`.

src/app/shared/components/tag-list/style.scss
```diff
@@ -17,3 +17,9 @@
 :host {
   width: 100%;
 }
+
+.mat-standard-chip {
+  .mat-chip-remove {
+    background: black;
```
maciaszczykm, line 23: Does it work with dark and light themes?

maciaszczykm: Anyways it should be exported to theme.scss.

floreks:
> @floreks After the refactoring the name `tag-list` doesn't really make sense here. I think we should come up with a more generic name that de-couples it from the term `tag`.

I think we could simply go with `km-chip-list`. @maciaszczykm WDYT?

src/app/shared/components/tag-list/style.scss
ahmedwaleedmalik: Good pointer, you are right it should be exported to theme.

maciaszczykm:
> I think we could simply go with km-chip-list. @maciaszczykm WDYT?

It's a good name.

src/app/shared/components/tag-list/component.ts
```diff
@@ -70,6 +91,35 @@ export class TagListComponent implements OnDestroy, ControlValueAccessor {
...
+    const set = new Set(tags.value);
+    if (set.size !== tags.value.length) {
+      return {
+        alreadyExists: {valid: true},
+      };
+    }
+    return null;
+  }
+
+  validatePattern(tags: FormControl) {
```
maciaszczykm, line 108: Type is missing.

PR #4210, opened 2022-02-23
Modifying template.html

holgers66:
Corrected description of Bring Your Own Cluster (kubeadm) steps. kubeadm does not need to be installed on a local machine (many customers even prohibit that). It should be installed on a control plane and run from there.

Corrects parts of template.html

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/details/cluster/machine-deployment-list/template.html
+++ b/src/app/cluster/details/cluster/machine-deployment-list/template.html
@@ -206,12 +206,12 @@ limitations under the License.
     <ol *ngIf="!!cluster.spec.cloud.bringyourown"
         class="kubeadm-manual">
       <li>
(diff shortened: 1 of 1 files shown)


embik:
Maybe I'm missing something, but that does not sound right. The point of the BYOC/kubeadm provider is that the control plane is managed by KKP and you only need to join nodes; you don't need a separate kubeadm-based control plane. And you need `kubeadm` to create the token, that's just the mechanic, or you do it manually without `kubeadm`. But why would you be allowed to install `kubectl` and not `kubeadm`?

holgers66:
@floreks @embik 

You're both correct. I made a mistake in the modification. I'll cancel the two PRs and create new ones with the correct text. Thanks.

PR #4215, opened 2022-02-24
fixing height for add and edit machine deployment form

### LINKED ISSUE #4205: part of the form for adding MD can't be seen
**What happened?**:
when you try to add a MD the form is seems to be taller than the screen and you can't see all the form so you can't see the add machine button .
Screenshot from 2022-02-22 12-07-56

**What should happen?**:
fixing the form size for Add or Edit MD 
**How to reproduce?**:
check if all the form is visible through different resolution   
**Environment**
- UI Version: KKP 2.19
- API Version: KKP 2.19
- Domain:
- Others: checked on screen with resolution of 1366 x 768

ahmadhamzh:
in some screen the form for add or edit a machine deployment isn't fully showed so this pull request adjust a height so it can be fully shown on the screen.

“closes #4205 ”

```release-note
NONE
```


src/app/cluster/details/cluster/edit-cluster/style.scss
```diff
@@ -15,6 +15,7 @@
 @use 'variables';
 
 .edit-cluster-form {
+  height: 60vh;
```
maciaszczykm, line 18: Instead of changing the height I would consider changing max-height instead, so it will prevent the window being too big but also keep the current behavior.

floreks:
Also, I think that this is a global issue right now. `.mat-dialog-content` max height is defined globally in the `material/_main.scss`. We should update it globally so that the dialog can be shrunk with the screen.

src/app/node-data/dialog/style.scss
```diff
@@ -13,9 +13,9 @@
 // limitations under the License.
 
 form {
-  height: 90vh;
+  max-height: 90vh;
 
   mat-dialog-content {
-    height: 70vh;
+    max-height: 70vh;
```
maciaszczykm, line 19: Do we need to adjust max heights of both elements here?

maciaszczykm:
I've tested it and it seems to work much better now.

/lgtm
/approve

src/app/shared/components/cluster-summary/template.html
```diff
@@ -41,7 +41,7 @@
        [fxFlex]="flipLayout ? '100%' : '50%'"
        fxLayout="column"
        class="step">
-    <div class="header counter-3">Settings</div>
+    <div class="header counter-2">Settings</div>
```
floreks, line 44: Why is it changed? `Settings` step is 3rd and `Cluster` step is 2nd. It was ok.

floreks (changes requested):

src/app/shared/components/cluster-summary/template.html
maciaszczykm: It was changed as the order in the dialog was 1 3 2 I believe.

maciaszczykm: Instead of changing numbers it should be sorted, right?

floreks: It should be based on `flipLayout` variable to make sure that in the wizard when there are 2 columns it will also be ok.

maciaszczykm: cc @ahmadhamzh

ahmadhamzh: i checked if there 2 column and it looks good but there is some provider the summary for them don't show the settings and Initial Nodes so it's better to bee sorted am on it....

maciaszczykm: I didn't check the code yet, but please note that the steps in the summary should have the same order as in the wizard. So first datacenter and provider etc.

floreks: It's about this view
image

If you change the counter then the step number will not match the section and it has to match. We could also simply disable the counter for the dialog mode and not display it at all.

ahmadhamzh: in the last commit below it will be like this :

Screenshot from 2022-03-02 05-11-18

and when create cluster from template  :

Screenshot from 2022-03-02 05-14-34

and the summary for cluster if there is no settings and initial nodes will be like this :

Screenshot from 2022-03-02 05-16-47

so i think all are good now ??

maciaszczykm:
Tested, works as expected.

/lgtm
/approve

maciaszczykm:
@ahmadhamzh The order is correct right now, but unfortunately the flow is not as @cschieder has proposed.

Right now the flow is:
1 2
3 4

But it should be:
1 3
2 4

src/app/shared/components/cluster-summary/template.html
ahmadhamzh: in the wizard summary i don't believe it could work like this
13
24
unless the dialog would look like it was
1
3
2
4
and also how can you disable the counter from just the dialog ??
the numbing come from the summary component which we call it in the dialog and in the wizard so i think its either we call the counter in both positions or disable it in both right ?
@floreks

ahmadhamzh: or the dialog could look like this 
Screenshot from 2022-03-03 15-46-45

floreks: We could easily disable the counter inside the dialog and only show the title. There are a couple of different ways this could be done. One of them is passing a `mode` parameter to the component that defines if it works in standalone or dialog mode and then adjusting the UI based on that.

src/app/shared/components/cluster-summary/template.html
```diff
@@ -14,7 +14,7 @@
 limitations under the License.
 -->
 
-<div class="km-wizard-summary"
+<div [ngClass]="showNumbering ? 'km-wizard-summary' : 'km-wizard-dialog-summary'"
```
maciaszczykm, line 17: This name was not the best before. I would rather not put "wizard" in it at all.

src/app/shared/components/cluster-summary/style.scss
```diff
@@ -46,6 +46,13 @@ $steps: 5;
   }
 }
 
+.km-wizard-dialog-summary {
+  .header {
+    font-size: variables.$font-size-subhead-lg;
+    margin-bottom: 16px;
+  }
+}
```
maciaszczykm, line 54: This code is the same as above. Instead you could only extract the `@include counter-x;` part to separate rule.

src/app/shared/components/cluster-from-template/content/template.html
```diff
@@ -33,7 +33,8 @@
                       [sshKeys]="sshKeys"
                       [datacenter]="datacenter"
                       [seedSettings]="seedSettings"
-                      [flipLayout]="false"
+                      [flipLayout]="true"
```
maciaszczykm, line 36: Is the flip layout still needed?

src/assets/css/global/_main.scss
```diff
@@ -442,7 +442,8 @@ km-dashboard {
   }
 }
 
-.km-wizard-summary {
+.km-wizard-summary,
+.km-wizard-dialog-summary {
```
maciaszczykm, line 446: Let's return to `km-wizard-summary` there.

src/app/shared/components/cluster-summary/template.html
```diff
@@ -14,7 +14,8 @@
 limitations under the License.
 -->
 
-<div class="km-wizard-summary"
+<div class="summary"
```
maciaszczykm, line 17: It is used globally so let's name it `km-wizard-summary` instead of `summary`.

src/assets/css/global/_main.scss
maciaszczykm: Is `km-wizard-dialog-summary` still needed?

PR #4255, opened 2022-03-03
CCM/CSI info addition 

mlavacca:
1. Changing the "External CCM" label to "External CCM/CSI migration". This information should reflect the status of the migration to the external CCM/CSI.
2. Addition of a new field in the "additional cluster information > MISC" section related to the external CCM/CSI setting for the current cluster.

### Release note

```release-note
Addition of a new field in the "additional cluster information > MISC" section related to the external CCM/CSI setting for the current cluster.
```

### DIFF
--- a/src/app/cluster/details/cluster/template.html
+++ b/src/app/cluster/details/cluster/template.html
@@ -340,8 +340,12 @@ limitations under the License.
                                  [value]="cluster.spec.enableUserSSHKeyAgent">
             </km-property-boolean>
 
+            <km-property-boolean label="External CCM/CSI"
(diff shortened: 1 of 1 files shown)


src/app/cluster/details/cluster/template.html
```diff
@@ -340,8 +340,12 @@
                                  [value]="cluster.spec.enableUserSSHKeyAgent">
             </km-property-boolean>
 
+            <km-property-boolean label="External CCM/CSI"
+                                 [value]="cluster.status.externalCCMMigration === 'NotNeeded'">
```
floreks, line 344: There is enum available in the template that can be used instead of hardcoded string. `externalCCMMigrationStatus.NotNeeded`

PR #4295, opened 2022-03-10
support Redirects to external cluster view

### LINKED ISSUE #4265: Allow redirects to the external clusters view
At the moment after i.e. disconnecting external cluster users get redirected to the cluster list view. It displays always the first tab, so the internal clusters. It would be nice to be able to redirect directly to the second tab too.

ahmadhamzh:
At the moment after disconnecting external cluster users get redirected to the cluster list view. It displays always the first tab, so the internal clusters. this PR redirect the user to the external cluster list view..

"closes #4265"

```release-note
NONE
```


maciaszczykm:
It is a working solution but I don't like the part where you have to use cluster service to switch to another tab. Instead I would check if we can reuse some of the Angular concepts, i.e. routing to fragments (see: https://angular.io/api/router/UrlCreationOptions#fragment). If that is not possible then perhaps we can register routes per tabs on cluster list view? 

cc @floreks

src/app/cluster/details/external-cluster/component.ts
```diff
@@ -180,6 +180,7 @@ export class ExternalClusterDetailsComponent implements OnInit, OnDestroy {
   disconnect(): void {
     this._clusterService.showDisconnectClusterDialog(this.cluster, this.projectID).subscribe(_ => {
       this._router.navigate(['/projects/' + this.projectID + '/clusters']);
+      this._clusterService.changeIndexClusterList(1);
```
maciaszczykm, line 183: Having `1` as param is non-intuitive. If this will be the end solution we would have to register enum for tabs.

src/app/cluster/list/template.html
```diff
@@ -17,7 +17,7 @@
 <ng-container [ngSwitch]="areExternalClustersEnabled">
   <mat-tab-group *ngSwitchCase="true"
                  animationDuration="0ms"
-                 [selectedIndex]="clusterIndex">
+                 [selectedIndex]="clusterIndex? 1 : 0">
```
maciaszczykm, line 20: This is the part that we can still improve I think. At the moment there is no difference what will be the value of fragment, if it is not null then the tab will be switched.

src/app/cluster/list/component.ts
```diff
@@ -27,11 +27,15 @@ export class ClustersComponent implements OnInit, OnDestroy {
...
   areExternalClustersEnabled = false;
 
-  constructor(private readonly _settingsService: SettingsService, private readonly _clusterService: ClusterService) {}
+  constructor(private readonly _settingsService: SettingsService, private route: ActivatedRoute) {}
 
-  clusterIndex = this._clusterService.clusterIndex;
+  clusterIndex = '';
 
   ngOnInit(): void {
+    this.route.fragment.subscribe((fragment: string) => {
```
maciaszczykm, line 35: It is a leak. Besides that, I'm not sure if we need to subscribe for that. If there is a option to get current value once and use it, I would do it.

src/app/cluster/list/component.ts
```diff
@@ -18,6 +18,11 @@ import {SettingsService} from '@core/services/settings';
 import {map, takeUntil} from 'rxjs/operators';
 import {ActivatedRoute} from '@angular/router';
 
+export enum clusterTab {
```
maciaszczykm, line 21: s/clusterTab/ClusterTab

src/app/cluster/list/component.ts
```diff
@@ -18,6 +18,11 @@ import {SettingsService} from '@core/services/settings';
...
+  cluster,
```
maciaszczykm, line 22: s/cluster/Cluster

src/app/cluster/list/component.ts
```diff
@@ -18,6 +18,11 @@ import {SettingsService} from '@core/services/settings';
...
+  externalCluster,
```
maciaszczykm, line 23: s/externalCluster/ExternalCluster

src/app/cluster/list/component.ts
```diff
@@ -26,7 +32,9 @@ export class ClustersComponent implements OnInit, OnDestroy {
   private _unsubscribe: Subject<void> = new Subject<void>();
   areExternalClustersEnabled = false;
 
-  constructor(private readonly _settingsService: SettingsService) {}
+  constructor(private readonly _settingsService: SettingsService, private route: ActivatedRoute) {}
+
+  clusterIndex = this.route.snapshot.fragment;
```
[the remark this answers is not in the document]
floreks: Is route data actually even available before the class is constructed?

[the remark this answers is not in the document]
ahmadhamzh: well it's working so i think yes it's available maybe because declaration is first thing to read in js right ?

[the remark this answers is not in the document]
maciaszczykm: Actually it's working because the constructor is first and route already has the updated state.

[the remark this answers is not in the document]
maciaszczykm: It's working because the constructor is first and route already has the updated state.

maciaszczykm:
/hold

@ahmadhamzh I think you may need to adjust e2e test for external clusters to make the checks pass.

cypress/pages/clusters.po.ts
```diff
@@ -259,6 +259,17 @@ export class ClustersPage {
     cy.get('div').should(Condition.Contain, 'No clusters available.');
   }
 
+  static verifyNoExternalClusters(): void {
+    if (Mocks.enabled()) {
+      cy.intercept({method: RequestType.GET, path: Endpoint.Clusters}, []);
```
maciaszczykm, line 264: You are intercepting the internal cluster requests here, for external clusters use `Endpoint.ExternalClusters`.

PR #4325, opened 2022-03-16
Fix issue with Openstack preset creation 

### LINKED ISSUE #4323: Domain is not required when using application credentials
**What happened?**:

Right now `domain` is a required field when we are using application credentials for OpenStack. This field is not required when using application credentials, so it should be removed.

Screenshot 2022-03-16 at 1 16 26 PM

### LINKED ISSUE #4324: Unable to create preset for openstack
Screenshot 2022-03-16 at 1 31 40 PM

floreks:
- Fix an issue with Openstack preset creation

Fixes #4324

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/settings/admin/presets/dialog/steps/settings/component.ts
+++ b/src/app/settings/admin/presets/dialog/steps/settings/component.ts
@@ -12,18 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Component, forwardRef, OnInit, Input} from '@angular/core';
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik:
/hold

• Make Domain optional when app credentials are used

Please handle this in a separate PR since this shouldn’t be a part of 2.20. We need to work on this in KKP, API, and machine-controller first.

xrstf:
/override pre-dashboard-test-e2e-ce

under protest

Issue #4344, opened 2022-03-22
Investigate and stabilize e2e tests


floreks:
**What happened?**:
Our e2e tests are randomly failing too often. We have to stabilize them as it often blocks our work.

Disabled in: https://github.com/kubermatic/dashboard/pull/4337#

floreks:
Closing in favor of #4371

PR #4364, opened 2022-03-25
Base for the e2e rewrite

### LINKED ISSUE #4372: Investigate and fix full e2e container issues
Currently, the full e2e tests are failing at a very early stage. It should be investigated and fixed.

floreks:
- Update kind/k8s node version
- Update full e2e container and remove kind cache usage as it was breaking the tests
- Provide a generic skeleton for the e2e handling
- Rewrite ServiceAccount e2e test as an example
- Update mocking logic to use different mocks based on user actions instead of always returning the same mocked data
- Update the mocked/real tests handling logic and move them to specific strategy handling classes
- Fix `license-check-and-add` configuration as it was failing with error

There are still some leftovers in the code and all tests except the new one have been disabled but I'd leave it for the moment and remove\update chunk by chunk while rewriting more tests.

Related to the #4344
Fixes #4372

### Release note

```release-note
NONE
```


maciaszczykm:
Love it. 

/lgtm

PR #4370, opened 2022-03-29
Add autofocus to dialog where needed

### LINKED ISSUE #4229: Add autofocus to dialogs
**What would you like to be added? / User Story**: When opening dialogs (i.e. create project dialog) it would be nice to add autofocus on the some input (in this case project name) to improve the UX. Let's check all the dialogs that are available.

ahmadhamzh:

add auto focus directive to input filed when some dialog showed
    
"closes #4229"

```release-note
NONE
```


src/app/cluster/details/cluster/add-machine-network/component.spec.ts
```diff
@@ -35,19 +35,21 @@ describe('AddMachineNetworkComponent', () => {
...
-      providers: [
-        WizardService,
-        {provide: ClusterService, useClass: ClusterMockService},
-        {provide: MatDialogRef, useClass: MatDialogRefMock},
-        NotificationService,
-      ],
-      teardown: {destroyAfterEach: false},
-    }).compileComponents();
-  }));
+  beforeEach(
```
maciaszczykm, line 38: It looks that you have some different version of dependencies installed. Run `npm ci` each time you download latest changes from master. To check it now you can run `npm ci && npm run fix`.

src/app/shared/directives/autofocus/directive-autofoocusDiallog.ts
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Directive, ElementRef, OnInit} from '@angular/core';
+
+@Directive({
+  selector: '[kmAutofocusDialog]',
+})
+export class AutofocusDirectiveDialog implements OnInit {
```
floreks, line 20: I'd prefer reusing and slightly adjusting the already existing `kmAutofocus` directive instead of creating almost exact copy of it.

floreks, line 20: I'd prefer reusing and slightly adjusting the already existing `kmAutofocus` directive instead of creating almost exact copy of it.

src/app/shared/directives/autofocus/directive.ts
```diff
@@ -27,15 +27,15 @@ export class AutofocusDirective implements AfterViewInit, OnDestroy {
...
-    if (!this.opened) {
-      throw new Error('[opened] event binding is undefined');
+    if (this.opened) {
+      this.opened.pipe(takeUntil(this._unsubscribe)).subscribe(opened => {
+        if (opened) {
+          setTimeout(() => this._el.nativeElement.focus());
+        }
+      });
+    } else {
+      setTimeout(() => this._el.nativeElement.focus());
```
floreks, line 37: Extract this common part to a method

src/app/shared/directives/autofocus/directive.ts
```diff
@@ -27,15 +27,15 @@ export class AutofocusDirective implements AfterViewInit, OnDestroy {
   constructor(private readonly _el: ElementRef) {}
 
   ngAfterViewInit(): void {
-    if (!this.opened) {
-      throw new Error('[opened] event binding is undefined');
+    if (this.opened) {
+      this.opened.pipe(takeUntil(this._unsubscribe)).subscribe(opened => {
```
floreks, line 31: Since it should only work when `opened` is `true` then use a `filter` pipe here to avoid `if` inside the subscribe.

src/app/shared/directives/autofocus/directive.ts
```diff
@@ -26,16 +26,21 @@ export class AutofocusDirective implements AfterViewInit, OnDestroy {
 
   constructor(private readonly _el: ElementRef) {}
 
+  focus() {
```
maciaszczykm, line 29: nit: Place lifecycle hook methods (in this case `ngAfterViewInit`) above the rest.

src/app/shared/directives/autofocus/directive.ts
```diff
@@ -26,16 +26,21 @@ export class AutofocusDirective implements AfterViewInit, OnDestroy {
...
+  }
+
   ngAfterViewInit(): void {
-    if (!this.opened) {
-      throw new Error('[opened] event binding is undefined');
+    if (this.opened) {
+      this.opened
+        .pipe(takeUntil(this._unsubscribe))
+        .pipe(filter(open => open))
+        .subscribe(_ => {
```
maciaszczykm, line 38: nit: You can skip {} if there is just one line inside.

src/app/shared/directives/autofocus/directive.ts
floreks: Also, this method should be private as it will not be directly used inside the template.

src/app/shared/directives/autofocus/directive.ts
```diff
@@ -26,16 +26,21 @@ export class AutofocusDirective implements AfterViewInit, OnDestroy {
...
+  focus() {
+    setTimeout(() => this._el.nativeElement.focus());
+  }
+
   ngAfterViewInit(): void {
-    if (!this.opened) {
-      throw new Error('[opened] event binding is undefined');
+    if (this.opened) {
+      this.opened
+        .pipe(takeUntil(this._unsubscribe))
```
floreks, line 36: `takeUntil` pipe should always be a last pipe before the `subscribe`

src/app/shared/directives/autofocus/directive.ts
floreks: I'd actually pass the focus method directly as subscribe parameter.
```typescript
...
.subscribe(this._focus.bind(this))
```

src/app/shared/components/add-ssh-key-dialog/template.html
```diff
@@ -25,8 +25,8 @@
              [formControlName]="controls.Name"
              type="text"
              autocomplete="off"
-             cdkFocusInitial>
-      <mat-error *ngIf="form.get(controls.Name).hasError('required')">
+             kmAutofocus>
+      <mat-error *ngIf="form.controls.name.hasError('required')">
```
floreks, line 29: Revert this line to the previous form.

PR #4374, opened 2022-03-31
support Redirects to external cluster view 

### LINKED ISSUE #4265: Allow redirects to the external clusters view

ahmadhamzh:

At the moment after disconnecting external cluster users get redirected to the cluster list view. It displays always the first tab, so the internal clusters. this PR redirect the user to the external cluster list view .

"closes #4265 "

```release-note
NONE
```


maciaszczykm:
Tested. Works as expected.

/lgtm
/approve

PR #4375, opened 2022-03-31
[release/v2.18] Bump version to 2.18.9

embik:

We need to cut another set of KKP patch releases. This creates the dashboard version for that.

### Release note

```release-note
NONE
```


floreks:
Needs to be overridden as e2e will not pass.

Issue #4383, opened 2022-04-04
Project Summary 


cschieder:
https://www.figma.com/file/dJQKgUPzIDND36o2Nq8dct/Loodse-UI?node-id=6446%3A49393
![Clusters](https://user-images.githubusercontent.com/49145827/163332427-284f7f76-8beb-4f4c-a6a8-ecd30dd0a88d.png)

maciaszczykm:
https://github.com/kubermatic/kubermatic/issues/9639 is done.

PR #4399, opened 2022-04-12
Add project overview

maciaszczykm:
Zrzut ekranu 2022-04-22 o 13 08 11

**Notes:** 
- Project quota is not done as API is not ready yet.
- Cluster location map is not done yet as there not too many components for that and API has to be checked.
- Number of cluster machine deployments is not done as API is not ready yet
- Menu entry is not registered as the view is not 100% ready.
- Ideally we should avoid loading templates, backups, ssh keys and service accounts and use some new API endpoint for returning the number stats.

Related to https://github.com/kubermatic/dashboard/issues/4383.

### Release note
```release-note
NONE
```


src/app/project-overview/clusters-overview/component.ts
```diff
@@ -0,0 +1,113 @@
...
+export class ClustersOverviewComponent implements OnInit, OnChanges {
+  @Input() clusters: Cluster[] = [];
+  @Input() clusterHealth: Health[] = [];
+  @Input() externalClusters: ExternalCluster[] = [];
+  @Input() externalClustersEnabled = false;
+  clusterColumns: string[] = ['status', 'name'];
+  clusterDataSource = new MatTableDataSource<Cluster>();
+  externalClusterColumns: string[] = ['status', 'name'];
+  externalClusterDataSource = new MatTableDataSource<ExternalCluster>();
+  projectID = this._activeRoute.snapshot.paramMap.get(PathParam.ProjectID);
```
floreks, line 40: Ideally, it would be best to get the selected project from the service and update the view when it changes.

maciaszczykm:
@floreks PTAL

PR #4411, opened 2022-04-21
Open dialog api method

### LINKED ISSUE #4406: Split the methods that open dialog and api req into two methods
in order to migrate throttleClick directives to km-button component a PR been created and since the throttle is applied on the buttons that do an Api req there is some buttons trigger a method that open dialog and do the API req on the same method, so we want to split that method into two, one would open a dialog component and the other would trigger the API  req when confirmed the dialog  

ref : #4379

ahmadhamzh:

in order to migrate throttleClick directives to km-button component a PR been created and since the throttle is applied on the buttons that do an Api req there is some buttons trigger a method that open dialog and do the API req on the same method, so we want to split that method into two, one would open a dialog component and the other would trigger the API req when confirmed the dialog

"closes #4406"
```release-note
NONE
```


src/app/core/services/cluster.ts
```diff
@@ -167,21 +172,30 @@ export class ClusterService {
...
       data: {
         title: 'Disconnect Cluster',
         message: `Are you sure you want to disconnect ${cluster.name} cluster?`,
         confirmLabel: 'Disconnect',
+        throttelButton: true,
+        observable: this._deleteExternalCluster(projectID, cluster.id),
+        // next: this._router.navigate(['/projects/' + projectID + '/clusters'], {
+        //     fragment: `${ClusterListTab.ExternalCluster}`,
+        //   }).finally(() => this._notificationService.success(`Disconnected the ${cluster.name} cluster`))
       },
```
ahmadhamzh, line 186: if i uncomment lines (183-185) next will execute just when the dialog open before i click the button

ahmadhamzh:
disconnect external cluster is working now but onNext isn't working and Im not sure if observable also working as it have to be so it would be great if i can have some hint 
this is the commit with the changes 
https://github.com/kubermatic/dashboard/pull/4411/commits/af45e5937db147ec61ff3b03f6bdea00db4f2167

@maciaszczykm 
@floreks

src/app/cluster/details/cluster/add-machine-network/component.spec.ts
```diff
@@ -29,13 +29,15 @@ import {NotificationService} from '@core/services/notification';
 import {SharedModule} from '@shared/module';
 import {AddMachineNetworkComponent} from './component';
 
+const modules: any[] = [BrowserModule, HttpClientModule, BrowserAnimationsModule, RouterTestingModule, SharedModule];
```
floreks, line 32: Do not use `any`

src/app/cluster/details/cluster/add-machine-network/component.spec.ts
```diff
@@ -29,13 +29,21 @@ import {NotificationService} from '@core/services/notification';
 import {SharedModule} from '@shared/module';
 import {AddMachineNetworkComponent} from './component';
 
+const modules: (BrowserModule | RouterTestingModule)[] = [
```
maciaszczykm, line 32: You do not list all types there. You can:
1. Skip the const as it is used in once place only (preferred).
2. Remove type completely.

src/app/wizard/template.html
```diff
@@ -102,17 +102,14 @@
...
-                  *ngIf="last"
-                  kmThrottleClick
-                  (throttleClick)="create()">
-            <i class="km-icon-mask km-icon-add"></i>
-            <span>Create Cluster</span>
-          </button>
+          <km-button id="km-wizard-create-btn"
+                     *ngIf="last"
+                     icon="km-icon-add"
+                     label="Create Cluster hhhhhhhhh"
```
maciaszczykm, line 108: ?

src/app/shared/components/external-cluster-data-dialog/template.html
```diff
@@ -35,12 +35,11 @@
...
-          kmThrottleClick
-          (throttleClick)="handler()"
-          [disabled]="!form.valid || !kubeconfig">
-    <i class="km-icon-mask km-icon-edit w-24"></i>
-    <span>Edit Cluster Connection</span>
-  </button>
+  <km-button id="external-cluster-confirm-btn"
+             icon="km-icon-edit"
+             label="Edit Cluster Connection"
+             [disabled]="!form.valid || !kubeconfig"
```
maciaszczykm, line 41: Use `form.invalid` instead of `!form.valid`.

cypress/pages/v2/wizard/page.ts
```diff
@@ -63,7 +63,7 @@ export class Wizard extends PageOptions implements Page {
 
     // kubeadm does not have additional steps
     if (provider === Provider.kubeadm) {
-      this.Buttons.create.click({force: true});
```
floreks, line 66: Actually, we do not really want to click multiple buttons here. Code should be modified to click only a single button that is a `Create cluster`.

ahmadhamzh: so maybe i use an id for the button inside the km-button component

ahmadhamzh: and also is it like after the click the button already will be disabled 
i add that because the button is actually inside the component i'm targeting

ahmadhamzh:
@maciaszczykm

src/app/shared/components/confirmation-dialog/template.html
```diff
@@ -34,16 +34,26 @@
              title={{data.inputTitle}}
              (blur)="onChange($event)"
              (keyup)="onChange($event)"
-             cdkFocusInitial>
+             kmAutofocus>
     </mat-form-field>
   </mat-dialog-content>
   <mat-dialog-actions>
-    <button mat-flat-button
+    <button *ngIf="!data.throttelButton"
```
maciaszczykm, line 41: There is a typo here.

cypress/pages/wizard.po.ts
```diff
@@ -62,7 +62,7 @@ export class WizardPage {
   }
 
   static getCreateBtn(): Cypress.Chainable {
-    return cy.get('#km-wizard-create-btn');
+    return cy.get('#km-button-buttonClick');
```
maciaszczykm, line 65: Mixed naming conventions.

src/app/shared/components/button/template.html
```diff
@@ -29,6 +30,7 @@
 </button>
 <button *ngIf="iconButton"
         mat-icon-button
+        [id]="this.buttonId"
```
maciaszczykm, line 33: Applying the same id to multiple elements is invalid HTML and should be avoided.

ahmadhamzh: but the tow element wont show together it's either one of them is shown 
you mean line 33 and line 19 right ?there is *ngIf that prevent them to been shown together

ahmadhamzh: i thought maybe at some place we would have the km-button component using multiple time in the same page so they would have the same id if it's constant and in the tests we need to get specific button by id  to click the button inside the component so if they have the same id both will be clicked

src/app/wizard/component.ts
```diff
@@ -106,50 +107,30 @@ export class WizardComponent implements OnInit, OnDestroy {
...
-              `Added the ${key.name} SSH key to the cluster ${createCluster.cluster.name}`
-            )
+      .pipe(switchMap(cluster => this._clusterService.cluster(this.project.id, cluster.id)))
+      .pipe(takeUntil(this._unsubscribe));
+  }
+
+  onNext(cluster: Cluster): void {
+    this.creating = true;
+    this._notificationService.success(`Created the ${cluster.name} cluster`);
+    this._googleAnalyticsService.emitEvent('clusterCreation', 'clusterCreated');
```
floreks, line 122: Why is it there? We have removed most of the GA usages as it is not configured anyway and not properly used.

floreks:
/lgtm

Will let @maciaszczykm to give the final approval.

Issue #4404, opened 2022-04-14
Extend presets view for stats


zreigz:
This is a new endpoint for stats. Display additional info for preset list: https://github.com/kubermatic/kubermatic/pull/9596

PR #4412, opened 2022-04-21
Extend presets view for stats

KhizerRehan:

Introducing changes to the Admin Panel under Provider Presets Tab
- fetch additional stats for each preset and display information in table

updated

### Which issue(s) this PR fixes:
Fix: https://github.com/kubermatic/dashboard/issues/4404

- @cschieder can confirm
 - Placement of columns and If any UI update is required for better UX

### Release note

```release-note
 Add additional stats for each Provider Preset

```

### DIFF
--- a/src/app/settings/admin/presets/component.ts
+++ b/src/app/settings/admin/presets/component.ts
@@ -48,12 +48,13 @@ enum Column {
   styleUrls: ['./style.scss'],
 })
 export class PresetListComponent implements OnInit, OnDestroy, OnChanges {
+  readonly providers = NODE_PROVIDERS;
+
   presets: Preset[] = [];
   dataSource = new MatTableDataSource<Preset>();
   displayedColumns: string[] = Object.values(Column);
   datacenters: Datacenter[] = [];
   datacenterFilter: string;
-  readonly providers = NODE_PROVIDERS;
   providerFilter: NodeProvider;
 
   @ViewChild(MatSort, {static: true}) sort: MatSort;
--- a/src/app/settings/admin/presets/component.ts
+++ b/src/app/settings/admin/presets/component.ts
@@ -120,14 +120,20 @@ export class PresetListComponent implements OnInit, OnDestroy, OnChanges {
         }),
         takeUntil(this._unsubscribe)
       )
-      .subscribe((stats: PresetStat[]) => {
-        this.presets.forEach((preset: Preset, idx: number) => {
-          preset.associatedClusters = stats[idx].associatedClusters;
-          preset.associatedClusterTemplates = stats[idx].associatedClusterTemplates;
-        });
-        this.dataSource.data = this.presets;
-        this.isBusyCounter--;
-      });
+      .subscribe(
+        (stats: PresetStat[]) => {
+          this.presets.forEach((preset: Preset, idx: number) => {
+            preset.associatedClusters = stats[idx].associatedClusters;
+            preset.associatedClusterTemplates = stats[idx].associatedClusterTemplates;
+          });
+          this.dataSource.data = this.presets;
+          this.isBusyCounter--;
+        },
+        _ => {
+          this._notificationService.error('Could not fetch Presets data');
+          this.isBusyCounter--;
+        }
+      );
 
     this._datacenterService.datacenters.pipe(takeUntil(this._unsubscribe)).subscribe(datacenters => {
       this.datacenters = datacenters;
--- a/src/app/settings/admin/presets/component.ts
+++ b/src/app/settings/admin/presets/component.ts
@@ -108,7 +108,7 @@ export class PresetListComponent implements OnInit, OnDestroy, OnChanges {
     merge(of(true), this._presetsChanged)
       .pipe(
         switchMap(_ => this._presets$),
-        tap(presetList => {
+        tap((presetList: PresetList) => {
           this.presets = presetList.items;
         }),
         switchMap(_ => {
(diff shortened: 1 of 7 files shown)


floreks:
Column names are a bit too long IMHO. @cschieder WDYT?

src/app/settings/admin/presets/style.scss
```diff
@@ -68,3 +68,9 @@
 #km-add-preset-btn {
   margin: 2px 0;
 }
+
+mat-card-content {
+  .mat-header-cell.km-header-cell.km-text-center {
+    text-align: center;
```
maciaszczykm, line 74: Did you tried to use fxFlex for that?

src/app/settings/admin/presets/component.ts
```diff
@@ -102,11 +106,33 @@ export class PresetListComponent implements OnInit, OnDestroy, OnChanges {
...
+          this.isBusyCounter++;
+          return forkJoin(presetStats$);
+        }),
+        map((stats: PresetStat[]) => {
+          const updatePresets = _.cloneDeep(this.presets);
+          updatePresets.forEach((preset: Preset, index: number) => {
+            preset.associatedClusters = stats[index].associatedClusters;
+            preset.associatedClusterTemplates = stats[index].associatedClusterTemplates;
+          });
+          return updatePresets;
```
maciaszczykm, line 128: You can skip this map and move it to the subscribe, so cloneDeep() will not be required.

src/app/settings/admin/presets/template.html
```diff
@@ -118,10 +118,29 @@
...
+      <ng-container matColumnDef="associatedClusterTemplates">
+        <th *matHeaderCellDef
+            class="km-header-cell km-text-center p-20"
+            mat-header-cell>Associated Clusters Templates
+        </th>
+        <td *matCellDef="let element" class="km-text-center"
+            mat-cell>{{element.associatedClusterTemplates}}</td>
+      </ng-container>
+
+
```
maciaszczykm, line 139: Avoid more than 1 empty line in a row.

src/app/settings/admin/presets/component.ts
```diff
@@ -102,11 +106,33 @@ export class PresetListComponent implements OnInit, OnDestroy, OnChanges {
...
+        tap((presetList: PresetList) => {
+          this.presets = presetList.items;
+        }),
+        switchMap(_ => {
+          const presetStats$ = [];
+          this.presets.forEach((preset: Preset) => {
+            presetStats$.push(this._presetService.getPresetStatsBy(preset.name));
+          });
+          this.isBusyCounter++;
+          return forkJoin(presetStats$);
```
maciaszczykm, line 120: Fork join will wait until all subscriptions have emitted values AFAIR, why using counter if that is true?

floreks: Definitely, the API in this case is not optimal. This will put a lot of overhead on the UI and API since for every preset on the list we will make a request for the stats. I'd vote for updating the API and not using it this way.

KhizerRehan: @floreks that's correct. This is overhead on UI to have n+1 call for each preset and as no of presets will increase it will cause slight delays for page to show presets data + additional info on table. 

- +1 if it can be added up API `/presets` endpoint than from FE the extra API call be be removed.

KhizerRehan: @maciaszczykm correct on first part.

why using counter if that is true?

Reason: on legacy code the check was `!presets` to show busy counter and it will be always true as `!! []` -> true so loader doesn't appear.

Adding **isBusyCounter**: number -> counter incremented when any API call is initiated so that spinner can be shown and disable spinner on when all api calls or task is been done until we reaches final subscription close the spinner by decrement **isBusyCounter**

thanks

src/app/settings/admin/presets/component.ts
KhizerRehan: @maciaszczykm you are correct it can be skipped and logic can be removed to within in subscribe.

I intentionally added it due following reasons:

- Keep it clear in terms of code readability that we are manipulating existing preset e.g in this case adding filling up values to optional properties

- All manipulations should be done on previous pipe operators and final **subscribe** should only get **latest** data that needs to be bind.


**Note:** let me know your final thoughts i can update it.

thanks

src/assets/css/global/_main.scss
```diff
@@ -99,6 +99,10 @@ hr {
   width: 100%;
 }
 
+.km-text-center {
+  text-align: center;
```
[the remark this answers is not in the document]
KhizerRehan: thanks, i can check it out.

src/app/settings/admin/presets/style.scss
KhizerRehan: remove as styles are not required anymore due to 
- Due to `left-alignment` of text/heading
- Shorten heading of coulmns

src/app/settings/admin/presets/template.html
KhizerRehan: fixed

src/app/settings/admin/presets/component.ts
maciaszczykm: You can extract the part from the forEach to the separate method then. Using cloneDeep adds additional complexity that is not required and will slow it down.

src/app/settings/admin/presets/style.scss
KhizerRehan: @cschieder Updated screenshot is attached:

fixes as per our consensus 👍🏽 
https://github.com/kubermatic/dashboard/pull/4412#discussion_r855116716

src/app/settings/admin/presets/component.ts
maciaszczykm: Hmm, okay. Let's change the API part if we can and then we can keep it. Would be nice to add it to error handler or finalizer too.

KhizerRehan: Good Point 👍🏽

PR #4428, opened 2022-04-26
Proposal for unit test guideline

pkprzekwas:
Here are some of points / guidelines regarding unit testing. Feel free to share your thoughts and suggest more points.

-

### Release note
```release-note
NONE
```


Development.md
```diff
@@ -57,6 +57,16 @@ To automatically fix issues run `npm run fix` command.
 ## Running Unit Tests
 Run `npm test` to execute the unit tests via [Jest](https://jestjs.io/).
 
+#### Unit testing guideline
```
maciaszczykm, line 60: Use title case.

Development.md
```diff
@@ -57,6 +57,16 @@ To automatically fix issues run `npm run fix` command.
...
+- Component tests should be located in the same directory as the tested component and named `component.spec.ts`.
+- Focus on testing components that are shared across the project (see `src/app/shared`).
+- Try to cover any conditional renders in tests (e.g. displaying spinner while waiting for data).
+- If a component is expected to react on events, try to cover that in tests.
+- Avoid testing too deep. Component's children should be tested separatly.
```
maciaszczykm, line 65: s/separatly/separately

Development.md
```diff
@@ -57,6 +57,16 @@ To automatically fix issues run `npm run fix` command.
...
+- Avoid complex css queries, mark an element with an `id` instead.
```
maciaszczykm, line 66: s/css/CSS

Development.md
```diff
@@ -57,6 +57,16 @@ To automatically fix issues run `npm run fix` command.
 ## Running Unit Tests
 Run `npm test` to execute the unit tests via [Jest](https://jestjs.io/).
 
+#### Unit testing guideline
+- Component tests should be located in the same directory as the tested component and named `component.spec.ts`.
+- Focus on testing components that are shared across the project (see `src/app/shared`).
```
maciaszczykm, line 62: Perhaps it is also worth mentioning that we keep test utils and mocks in the `src/test`.

maciaszczykm:
/lgtm
/approve
/hold

@floreks @KhizerRehan @ahmadhamzh do you want to add anything?

Development.md
```diff
@@ -57,6 +57,17 @@ To automatically fix issues run `npm run fix` command.
 ## Running Unit Tests
 Run `npm test` to execute the unit tests via [Jest](https://jestjs.io/).
 
+#### Unit Testing Guideline
+- Component tests should be located in the same directory as the tested component and named `component.spec.ts`.
+- Test utils and mocks are placed in `src/test`.
+- Focus on testing components that are shared across the project (see `src/app/shared`).
```
KhizerRehan, line 63: I posted this comment but somehow it was in PENDING state-> reposting

Hi, Thanks for proposing i have opinion but i agree on some of the points:

IMHO,
Components which are isolated and pure based on Input parameters should be more **focus** on Unit Testing as said components which are `Shared` should be tested for Unit tests

 - Test should includes assertions for maximum combinations of assertions to be checked to validated truthiness of shared components rather than testing individual components.


 I am not sure but on Individual `*.spec.ts` i saw test cases like checking truthiness Which may or mayn't bring values to UI instead an extra overhead to UI IMHO.
 
 **Test cases like..**
  - Form is Valid or Not
  - Component is Initiated
  - Heading is checked or not
  - Existence of Element
  - Visibility of Element
  - Element is clickable or not
  ... etc

  I would say all above purpose can be handled in E2E cypress test cases that's all my personal opinion.
  
  **Note:**
  You can ignore it as it was discussed on TOI session with @maciaszczykm and @floreks might be chances we need to test it when checking template existence.

KhizerRehan (commented):
**Point: If a component is expected to react on events, try to cover that in tests.**

I will like to know thoughts about how you are doing testing for components that have child component and nested child components?

Any reference component to test such events?

_"If a component is expected to react on events"_
 
 - How to handle async API calls? 
 - How to handle async API calls which are interconnected like based on previous response perform next functionality?

KhizerRehan:
FYI, I posted those at the time of PR but somehow it was in PENDING state :)

Development.md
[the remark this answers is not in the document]
floreks: I find it a bit unconventional to use custom attributes dedicated purely for test purposes. In general, it's more natural to test the application as is without having to add attributes to the production code that are used only for tests. Using a unique element identifier is the best way to do this IMHO.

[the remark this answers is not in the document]
maciaszczykm: I agree with @floreks. I would rather avoid test-specific code if that is possible.

pkprzekwas:
> **Point: If a component is expected to react on events, try to cover that in tests.**
> 
> I will like to know thoughts about how you are doing testing for components that have child component and nested child components?
> 
> Any reference component to test such events?
> 
> _"If a component is expected to react on events"_
> 
> * How to handle async API calls?
> * How to handle async API calls which are interconnected like based on previous response perform next functionality?

While testing components you should focus on a component's API (e.g. props or triggering/triggered events). If a component has child components, those have their own APIs and we should **not** try to test them while testing the parent. Not being able to figure out the API for a tested component may mean that the component requires some refactoring.

Testing async API calls is preferred to be handled in **E2E** tests. In the component/unit tests we can still check how components behave based on some mocked data or see if async functions were invoked based on some component's logic.

For some more advanced flows, like the one mentioned in the last question, I strongly suggest considering **E2E** testing.

Unfortunately, I don't know if we have some good examples for my ideas in the Dashboard's codebase. Have in mind that I haven't worked much with the Dashboard, so maybe someone else will be able to point you to good examples.

ahmedwaleedmalik (approved):
/approve

I'm merging this because at least it adds some valuable insights.

PR #4478, opened 2022-05-16
Add retention parameter to metering schedule configuration

### LINKED ISSUE #4471: Handle new "retention" metering parameter in the Dashboard
Extend metering schedule configuration with the new "retention" parameter. 

- present it in the metering schedule configuration table
- extend create and update forms with new parameter

pkprzekwas:
- Add retention parameter to metering schedules. 
- Remove edit option for metering schedule name - added accidentally.

#4471 

### Release note

```release-note
NONE
Add retention parameter to metering schedule configuration.
```

Screenshot 2022-05-13 at 14 48 23

Screenshot 2022-05-16 at 12 03 37


src/app/dynamic/enterprise/metering/component.ts
```diff
@@ -70,6 +70,11 @@ export class MeteringComponent implements OnInit {
       });
   }
 
+  ngOnDestroy(): void {
```
maciaszczykm, line 73: This is unused until class implements OnDestroy.

pkprzekwas: Are you sure about it? I just checked and putting console.log in the method printed message in the console after I switched the component. What you say makes sense and I'll add this interface to the class, just wanted to point out that it seems to work even without explicitly adding the interface to class definition.

KhizerRehan: @pkprzekwas 

Yes! interfaces are optional Angular lifecycle hooks are called even you don't implement it. This is helpful for devs. TypeScript compiler warns you if you don't implement the Lifecycle methods correctly but it is still executed incase of 
- pipes, services, directive and components

[Docs](https://v8.angular.io/guide/lifecycle-hooks#interfaces-are-optional-technically)

src/app/dynamic/enterprise/metering/schedule-config/template.html
```diff
@@ -80,14 +82,31 @@
...
-                 matTooltip="Range of days captured in each report."></div>
+                 matTooltip="Number of days captured in each report."></div>
           </div>
         </th>
         <td mat-cell
             *matCellDef="let element">{{element.interval}}
         </td>
       </ng-container>
 
+      <ng-container matColumnDef="schedule">
```
KhizerRehan, line 93: IMHO, instead of duplicating control names wwe can use existing enum Controls defined


e.g Usage

[matColumnDef]="column.<NAME_OF_CONTROL_DEFINED_IN_ENUM>"

Benefits:
- Avoid duplcation
- Avoid Errors while typing
- Single point of change if control name changes from TS file

KhizerRehan:
https://user-images.githubusercontent.com/17727069/170964994-064abb55-97f5-4c6c-a5eb-6b32143860c9.mp4

IMHO, there should be some sort of validation or regex validation when entering cron expression to avoid accepting invalid expression.

**Benefits:**

- Better UX 
- Avoid network call from frontend side

cc: @cschieder

KhizerRehan:
> c6a0f02e2594c08af46b6d1bb2b657d0.mp4 
> IMHO, there should be some sort of validation or regex validation when entering cron expression to avoid accepting invalid expression.
> 
> **Benefits:**
> 
> * Better UX
> * Avoid network call from frontend side
> 
> cc: @cschieder

Same for "Edit Schedule Configuration"

src/app/dynamic/enterprise/metering/component.ts
```diff
@@ -51,7 +51,7 @@ export class MeteringComponent implements OnInit {
       .pipe(switchMap(seed => this._dcService.seedSettings(seed)))
       .pipe(map(settings => settings.metering))
```
KhizerRehan, line 52: IMO, better to use `filter` operator instead of `map`

src/app/dynamic/enterprise/metering/component.ts
maciaszczykm: I was not aware of it but anyways it would be better to have it :)

src/app/dynamic/enterprise/metering/schedule-config/template.html
```diff
@@ -46,28 +46,30 @@
       <ng-container matColumnDef="name">
```
KhizerRehan, line 46: https://github.com/kubermatic/dashboard/pull/4478/files#r884573276

Same as above for all
- matColumnDef -> `name, retention, interval, schedule, acions`

src/app/dynamic/enterprise/metering/schedule-config/add-dialog/component.ts
```diff
@@ -110,6 +130,7 @@ export class MeteringScheduleAddDialog implements OnInit, OnDestroy {
       [Controls.Group]: this._builder.control(DefaultScheduleOption.Daily, Validators.required),
       [Controls.Schedule]: this._builder.control('0 0 * * *'),
```
KhizerRehan, line 131: IMO, there should be `default` `0 0 * * *` in DefaultScheduleInterval enum

KhizerRehan:
@pkprzekwas 

I noticed constant polling? Is it something that needs to occur? I am not sure about this but is there a possibility that endpoints `/reports` gets updated frequently.  thank you

https://user-images.githubusercontent.com/17727069/171110282-64b710d9-cd92-4b53-8a90-396c460c48cb.mp4

package.json
```diff
@@ -73,6 +73,7 @@
     "buffer": "6.0.3",
     "core-js": "3.22.4",
     "country-code-lookup": "0.0.20",
+    "cron-validator": "^1.3.1",
```
KhizerRehan, line 76: Hi, kindly remove `^` from package name so to save exact package version

KhizerRehan (approved):
/lgtm

src/app/dynamic/enterprise/metering/component.ts
```diff
@@ -49,9 +49,9 @@ export class MeteringComponent implements OnInit {
       .pipe(map(seeds => (seeds.length > 0 ? seeds[0] : null)))
       .pipe(filter(seed => seed !== null))
       .pipe(switchMap(seed => this._dcService.seedSettings(seed)))
-      .pipe(map(settings => settings.metering))
+      .pipe(pluck('metering'))
```
floreks, line 52: Even though `pluck` requires a couple of characters less than the `map`, we are losing type safety and checking completely due to direct string reference of the field name. I'd rather use `map` as it allows a compile-time check of the field reference for the metering object.

PR #4481, opened 2022-05-17
Give correct Kubernetes version that removes Docker support

embik:

The message shown to users was out of sync with the actual constraint (message said 1.22, but constraint is 1.24), so this updates the error message shown in the UI when trying to deploy Kubernetes 1.24 with Docker.

### Release note

```release-note
NONE
```


KhizerRehan:
IMHO, you need to update in this path too:
`src/app/cluster/details/cluster/edit-cluster`

```html
 <mat-error *ngIf="form.get(Controls.ContainerRuntime).hasError('dockerVersionCompatibility')">
          Docker is not supported since v1.22.0.
</mat-error>
```

PR #4496, opened 2022-05-23
Show updated config instead of default values in edit machine deployment dialog 

### LINKED ISSUE #4472: "Edit Machine Deployment" not loading current configuration of a MD
### What happened
I wanted to edit existing Machine Deployment via the UI, opened the "Edit Machine Deployment" dialog and noticed that it doesn't load current settings and instead shows the defaults. I thought I am only changing the replicas number, while actually all the configuration options were changed back to defaults.
picture

### Expected behavior
Edit Machine Deployment window should load existing configuration of the MachineDeployment.

### How to reproduce
1. Create a Machine deployment with non-default configuration
2. Open "Edit Machine Deployment" in a user cluster view
3. Check if the configuration is correct before saving changes

### Environment
- UI Version: g74ba3824
- API Version: g74ba3824

### Current workaround
edit values accordingly manually

Waseem826:
Signed-off-by: Waseem Abbas <waseemabbas8261@gmail.com>

Edit machine deployment dialog was displaying default values in combobox elements instead of what was configured. This has been fixed by removing `_clearX` method calls when combobox options are loading.

Fixes #4472 

Reason for removing these clear method calls:
- `_onXLoading` method is called when combobox options are loading
- `_onXLoading` method calls `_clearX` method
- `_clearX` method calls `reset` method of combobox
- Combobox emits `changed` event with `null` value
- `null` value gets assigned to property in `_nodeDataService.nodeData.spec.cloud.X` hence replacing original value
- Combobox options finish loading so `_setDefaultX` is called
- `_setDefaultX` method loads value from `_nodeDataService.nodeData.spec.cloud.X` which was set to `null` in previous step
- `_setDefaultX` method sets default value in combobox so original configured value is never shown

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/node-data/basic/provider/alibaba/component.ts
+++ b/src/app/node-data/basic/provider/alibaba/component.ts
@@ -205,20 +205,17 @@ export class AlibabaBasicNodeDataComponent extends BaseFormValidator implements 
   }
 
   private _onInstanceTypeLoading(): void {
-    this._clearInstanceType();
     this.instanceTypeLabel = InstanceTypeState.Loading;
     this._cdr.detectChanges();
   }
 
   private _onZoneLoading(): void {
-    this._clearZone();
     this.zoneLabel = ZoneState.Loading;
     this._cdr.detectChanges();
   }
 
   private _onVSwitchLoading(): void {
     this.isLoadingVSwitches = true;
-    this._clearVSwitch();
     this._cdr.detectChanges();
   }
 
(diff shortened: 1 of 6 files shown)


Waseem826:
I only have 1 preset so I've tested with following steps and it works:
1. Select Preset A
2. Select a Zone X
3. Go back and clear selected preset
4. Select Preset A again
5. Click Next
6. Default Zone value is selected (instead of Zone X)

The reason it works is because I've only removed `_clearX` method calls from `loading` methods.

floreks:
This scenario is quite clear. The one that I have described might be problematic since the value might not longer be available after changing credentials.

Waseem826:
@maciaszczykm @floreks is there anything else that needs to be done here?

maciaszczykm:
No, if it was tested as discussed then it can be merged.

/lgtm
/approve

csengerszabo:
As I see this was not released within the last patch release of KKP, can we include this into the next one? @Waseem826

PR #4504, opened 2022-05-24
Only consider MachineDeployment running if replicas are updated

embik:

MachineDeployment health in the dashboard does not consider whether replicas where updated. That can result in incorrect UI information during a MachineDeployment upgrade (it should be showing that the MachineDeployment is updating, but as long as enough old Machines are still available, it will not).

This PR includes `md.status.updatedReplicas` in the if condition for status "Running".

### Release note

```release-note
MachineDeployment health is only considered "Running" if replicas are all updated
```

### DIFF
--- a/src/app/shared/utils/health-status.ts
+++ b/src/app/shared/utils/health-status.ts
@@ -71,7 +71,11 @@ export function getNodeHealthStatus(n: Node): HealthStatus {
 export function getMachineDeploymentHealthStatus(md: MachineDeployment): HealthStatus {
   if (md.deletionTimestamp) {
     return new HealthStatus('Deleting', StatusIcon.Error);
-  } else if (md.status && md.status.availableReplicas === md.spec.replicas) {
+  } else if (
+    md.status &&
+    md.status.availableReplicas === md.spec.replicas &&
+    md.status.availableReplicas === md.status.updatedReplicas
+  ) {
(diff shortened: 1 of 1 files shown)


src/app/shared/utils/health-status.ts
```diff
@@ -71,7 +71,11 @@ export function getNodeHealthStatus(n: Node): HealthStatus {
 export function getMachineDeploymentHealthStatus(md: MachineDeployment): HealthStatus {
   if (md.deletionTimestamp) {
     return new HealthStatus('Deleting', StatusIcon.Error);
-  } else if (md.status && md.status.availableReplicas === md.spec.replicas) {
+  } else if (
+    md.status &&
+    md.status.availableReplicas === md.spec.replicas &&
+    md.status.availableReplicas === md.status.updatedReplicas
```
maciaszczykm, line 77: Just to confirm - both checks need to be valid to consider the MD as running?

embik: IMHO, yes. It's a valid scenario to have:
- spec.replicas=2
- status.availableReplicas=2
- status.updatedReplicas=0
Right after applying a MachineDeployment change, for example. But that should be "updating", not "running". You can try it out yourself on dev by updating a MachineDeployment, the status icon should stay green for some time even though the new nodes are already provisioning in the background.

PR #4514, opened 2022-05-30
Disable dynamic Kubelet config k8s 1.24

### LINKED ISSUE #4480: disable "Dynamic kubelet config" option for nodes for Kubernetes 1.24+
Kubernetes 1.24 and higher does not support dynamic kubelet configuration. Since 1.24 just landed on KKP master, it would be great if the "dynamic kubelet configuration" option in the "initial nodes" step (and potentially in other places? not sure) could be hidden if 1.24 or higher has been selected in the previous step.

ahmadhamzh:
Kubernetes 1.24 and higher does not support dynamic kubelet configuration.
so in this PR if the control plan is 1.24 or higher then the nodes can't have the option of using dynamic kubelet configuration.
"closes #4480"

```release-note
NONE
```


Waseem826 (changes requested):

src/app/cluster/details/shared/version-change-dialog/style.scss
```diff
@@ -12,6 +12,17 @@
...
 // limitations under the License.
 
+@use 'variables';
+
 .checkbox-wrapper {
   min-height: 35px;
 }
+
+.dynamic-kubelet-config-nodes {
+  font-size: variables.$font-size-body;
```
maciaszczykm, line 22: You don't need to apply font size there as it is in the parent class. Also, check if we have some global class for making the font weight bigger (bolder).

src/app/cluster/details/shared/version-change-dialog/template.html
```diff
@@ -55,6 +55,12 @@
       Upgrade Machine Deployments
     </mat-checkbox>
   </div>
+  <div *ngIf="!isDynamicKubletConfigSupportedInUpgrade()"
+       fxLayoutAlign="center start"
+       class="warning-container">
+    <i class="km-icon-warning"></i>
+    <div class="km-text dynamic-kubelet-config-warning">Kubernetes version 1.24 and higher do not support dynamic kubelet configuration. On upgrade, dynamic kubelet configuration is going to be disabled on the machine deployments: <span class="km-text dynamic-kubelet-config-nodes">{{this.nodesSupportDynamicKubeletConfig}}</span></div>
```
maciaszczykm, line 62: Put it in the different lines to make it easier to read.

src/app/node-data/template.html
```diff
@@ -150,6 +150,8 @@
 
         <div class="os-options">
           <mat-checkbox [id]="Controls.DynamicConfig"
+                        [matTooltip]="isDynamicKubletConfigSupported()? '' : 'Kubernetes version 1.24 and higher do not support dynamic kubelet configuration.'"
```
KhizerRehan, line 153: Better to have global constant for Kubernetes version and use that variable instead of adding inline` 1.24` as mentioned here:

https://github.com/kubermatic/dashboard/pull/4514/files#r891301253

src/app/node-data/component.ts
```diff
@@ -233,6 +233,10 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     this._nodeDataService.taints = this.taints;
   }
 
+  isDynamicKubletConfigSupported(): boolean {
```
KhizerRehan, line 236: IMO, it doesn't matter but i prefer Getter to use field when simple `boolean` values needs to be considered. This comment is Opinionated can be ignored

src/app/shared/entity/cluster.ts
```diff
@@ -64,6 +64,8 @@ export enum ContainerRuntime {
 
 export const END_OF_DOCKER_SUPPORT_VERSION = '1.24.0';
 
+export const END_OF_DYNAMIC_KUBELETCONFIG_SUPPORT_VERSION = '1.24';
```
KhizerRehan, line 67:
```diff
- END_OF_DYNAMIC_KUBELETCONFIG_SUPPORT_VERSION
+ END_OF_DYNAMIC_KUBELET_CONFIG_SUPPORT_VERSION
```

src/app/cluster/details/shared/version-change-dialog/component.ts
```diff
@@ -58,6 +62,14 @@ export class VersionChangeDialogComponent implements OnInit, OnDestroy {
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(project => (this.project = project));
     this._googleAnalyticsService.emitEvent('clusterOverview', 'clusterVersionChangeDialogOpened');
+
+    this._machineDeploymentService
+      .list(this.cluster.id, this.project.id)
+      .pipe(map(nodes => nodes.filter(node => node.spec.dynamicConfig)))
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(nodes => {
+        this.nodesSupportDynamicKubeletConfig = nodes.map(node => ` MD-${node.name}`).toString();
```
Waseem826, line 71: Instead of `toString` you can use `.join(', ')` which adds extra space after `,` for better readability.

Waseem826 (changes requested):

src/app/cluster/details/shared/version-change-dialog/component.ts
```diff
@@ -92,6 +104,11 @@ export class VersionChangeDialogComponent implements OnInit, OnDestroy {
     this._dialogRef.close(true);
   }
 
+  isDynamicKubletConfigSupportedInUpgrade(): boolean {
+    const endSliceParameter = 4;
```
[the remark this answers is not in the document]
ahmadhamzh: cant do it because of linting check 
` error  No magic number: 4  @typescript-eslint/no-magic-numbers`

ahmadhamzh:
@Waseem826 @KhizerRehan

src/app/cluster/details/shared/version-change-dialog/template.html
```diff
@@ -55,6 +55,16 @@
       Upgrade Machine Deployments
     </mat-checkbox>
   </div>
+  <div *ngIf="!isDynamicKubletConfigSupportedInUpgrade"
```
Waseem826, line 58: Can you add another check here so this only shows up when at least one machine-deployment has `dynamicConfig` enabled?
For example, if there is no machine deployment with `dynamicConfig` enabled then it looks like this:

screenshot-localhost_8000-2022 06 14-19_05_41

Waseem826 (changes requested):

src/app/cluster/details/shared/version-change-dialog/template.html
ahmadhamzh: so you mean like if there is no MD using dynamic kubelet config then sow another warning, like :
`kubernetes version 1.24 and higher do not support dynamic kubelet configuration`
and if there is some MD using dynamic kubelet config then show the current warning 
@Waseem826

KhizerRehan: There are multiple scenarios:

- When `NO` machine deployment exists user tries to add new MD + Kubernetes version 1.24 or > greater should we show “Toggle button” 
(IMO not since all MD will have auto disabled dynamic kubelet configuration

- When `NO`  machine deployment exists user tries to add new MD + Kubernetes version 1.24 < and user tries to enable toggle -> WARNING should be shown. 

- When node/machine are on 1.24 -> Should we hide that toggle?because it doesn’t make sense whether update we need to show warning or not because BE will automatically disables “dynamic kubelet configuration”


cc: @cschieder @embik

ahmadhamzh: the warning is not when you create new MD it's on upgrade the kubernetes version to 1.24 in cluster
and if the current version of kubernetes is 1.24 you can't upgrade it for now 
maybe adjust it to bee shown just when upgrde to 1.24 so in the future if the user want to upgrade to (1.24.5 for example) it wouldn't show the warning

Waseem826: @ahmadhamzh I think we should hide the warning altogether when there is no machine deployment with dynamic config. i.e. `nodesSupportDynamicKubeletConfig` is empty

ahmadhamzh: yes but what if the user want to upgrade to 1.24 and he don't know that he can't use dynamic kubelet config,so we suppose to warn him 
IMO the warning suppose to show when upgrade to 1.24 ,
if any MD using dynamic kubelet config then show the current error 

and if there is no MD using dynamic kubelet config then show the error :
`kubernetes version 1.24 and higher do not support dynamic kubelet configuration`

Waseem826: Lets move this conversation to Slack (in sig-api-ui) and ask @cschieder and others about this.

src/app/cluster/details/shared/version-change-dialog/template.html
```diff
@@ -55,6 +55,16 @@
       Upgrade Machine Deployments
     </mat-checkbox>
   </div>
+  <div *ngIf="!isDynamicKubletConfigSupportedInUpgrade && this.nodesSupportDynamicKubeletConfig"
```
KhizerRehan, line 58: avoid using `this` in template.

`public/protected BOTH`

- Any getter/setter property or method can be directly used in template without `this` as template is automatically binded to that class instance

src/app/cluster/details/shared/version-change-dialog/template.html
```diff
@@ -55,6 +55,16 @@
...
+       fxLayoutAlign="center start"
+       class="warning-container">
+    <i class="km-icon-warning"></i>
+    <div class="km-text dynamic-kubelet-config-warning">
+      Kubernetes version {{endOfDynamicKubeletConfigSupportVersion}} and higher do not support dynamic kubelet configuration.
+      On upgrade, dynamic kubelet configuration is going to be disabled on the machine deployments:
+      <span class="km-text-bolder">{{this.nodesSupportDynamicKubeletConfig}}</span>
```
KhizerRehan, line 65: remove `this`

KhizerRehan (changes requested):
kindly, see comments

PR #4520, opened 2022-05-30
[release/v2.20] Bump version to 2.20.3

embik:

We (cc @xrstf) plan to cut another KKP patch release, 2.20.3. This adds the required dashboard version bump.

### Release note

```release-note
NONE
```


embik:
@maciaszczykm do you think the ce test is going to succeed or should we override it?

Issue #4529, opened 2022-06-02
Allow multiple InternalIPs in node details of machine deployments


rastislavs:
**What would you like to be added? / User Story**:

In dual-stack case, it is common for nodes to have multiple InternalIPs: one IPv4 and one IPv6, e.g.:
```
,"addresses":[{"type":"InternalIP","address":"192.168.1.254"},{"type":"InternalIP","address":"2a04:d480:2c02:6:f816:3eff:fe46:5ba4"},{"type":"ExternalIP","address":"195.192.159.176"}
```

KKP dashboard in this case shows only one of them, e.g.:

![image](https://user-images.githubusercontent.com/15926980/171680856-36314423-1baf-49d5-8576-d7acfe56a6ea.png)

In such case, both InternalIPs should be shown.

PR #4542, opened 2022-06-07
4529 display multiple internal IPs in node details of machine deployment

Waseem826:
This PR adds functionality to display multiple Internal IPs in node details of a machine deployment.

screenshot-localhost_8000-2022 06 07-12_09_58

#4529 

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/shared/utils/node.ts
+++ b/src/app/shared/utils/node.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Node} from '../entity/node';
+import {Node, NodeIPAddress} from '../entity/node';
 
 export class NodeUtils {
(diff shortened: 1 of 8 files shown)


src/app/shared/utils/node.ts
```diff
@@ -38,13 +38,13 @@ export class NodeUtils {
     return nodeCapacity ? `${nodeCapacity} ${prefixes[i - 1]}` : 'unknown';
   }
 
-  static getAddresses(node: Node): object {
-    const addresses = {};
+  static getAddresses(node: Node): NodeIPAddress {
+    const addresses = new NodeIPAddress();
     for (const i in node.status.addresses) {
```
KhizerRehan, line 43: IMO, it will be good to use `index` or `idx` instead of `i` for readability

PR #4546, opened 2022-06-07
Update Angular and other dependencies

maciaszczykm:
Closes https://github.com/kubermatic/dashboard/issues/4543.

### Release note
```release-note
NONE
```


maciaszczykm:
PTAL

package.json
```diff
@@ -94,45 +94,46 @@
...
     "@types/js-yaml": "4.0.5",
     "@types/lodash": "4.14.182",
-    "@types/node": "17.0.36",
-    "@typescript-eslint/eslint-plugin": "5.22.0",
-    "@typescript-eslint/parser": "5.22.0",
+    "@types/node": "17.0.40",
+    "@typescript-eslint/eslint-plugin": "5.27.1",
+    "@typescript-eslint/parser": "5.27.1",
     "browserlist": "1.0.1",
+    "btoa": "1.2.1",
```
Waseem826, line 110: Is this required by some other dependency/package?

Waseem826 (changes requested):

KhizerRehan:
I am not sure if this will effect due to Upgrade of v.14 to Angular:
**Angular Upgrade guide**
- If you are defining routes with pathMatch, you may have to cast it to Route or Routes explicitly. Route.pathMatch is no longer compatible with string type.

 i can "pathMatch: 'full'," reference in code

package.json
maciaszczykm: Yes, with the changes it becomes required.

KhizerRehan:
Otherwise it looks good to me :) 👍🏽

maciaszczykm:
@KhizerRehan I don't think that this affects us:

```ts
const routes: Routes = [
  {
    path: '',
    component: FrontpageComponent,
    pathMatch: 'full',
  },
  ...
```

As you can see there is type declaration and the code compiles without any issues.

Waseem826 (approved):

PR #4584, opened 2022-06-12
Download token button

### LINKED ISSUE #4541: Download toke button doesn't work for Service Accounts
### What happened
Download toke button doesn't work for Service Accounts

### Environment
- UI Version:
- API Version:
- Domain:
- Others:

ahmadhamzh:
fix the download button for downloading token file after add new token 

"closes #4541"

```release-note
NONE
```


src/app/serviceaccount/token/add/template.html
```diff
@@ -41,42 +41,40 @@
...
-      <span>Got it</span>
-    </button>
-  </mat-dialog-actions>
 </form>
+<mat-dialog-actions *ngIf="mode === modes.Create">
+  <mat-spinner fxFlexAlign="center"
+               [diameter]="30"
+               color="accent"
+               *ngIf="updating"></mat-spinner>
+  <km-button *ngIf="first"
```
KhizerRehan, line 50: Functionality seems to work but i think we can do some improvements in code that will reduce the amount of duplication:

- Have a single check
```html
 <ng-container *ngIf="!first">
   /// Place code here 
 </ng-container>

File: src/app/serviceaccount/token/add/template.html [hunk 2]
```diff
@@ -41,42 +41,40 @@
...
+             id="km-serviceaccount-token-dialog-update-btn"
+             [icon]="getIconClass()"
+             [label]="mode+' Token'"
+             [disabled]="updating || form.invalid"
+             [observable]="getObservable()"
+             (next)="onNext($event)">
+  </km-button>
+  <a *ngIf="!first"
```
KhizerRehan, line 58: Functionality seems to work but i think we can do some improvements in code that will reduce the amount of duplication:

- Have a single check
```html
 <ng-container *ngIf="!first">
   /// Place code here 
 </ng-container>

Comment (KhizerRehan) [inline src/app/serviceaccount/token/add/template.html:58]: You can use `ngTemplate` to avoide code duplication

```html
<ng-template #downloadButton>
   <CODE that is duplicated>
   
   e.g Download button, Got it buttons ( Note: "Got it" has Id so you need to use "Context" variable to have ID on 1 button and other button should not.
</ng-template>
```

KhizerRehan (changes requested):
Added Comments for code improvements

KhizerRehan (changes requested):
Added new comment

Waseem826 (approved):

PR #4588, opened 2022-06-14
Pre-select rocky linux while editing machine deployment

### LINKED ISSUE #4571: Incorrect OS pre-selected when editing machine-deployment
### What happened
When editing an existing machine-deployment, wrong operating system is pre-selected. For example for a machine-deployment with Rocky Linux, Ubuntu is pre-selected:

image

### Expected behavior
When editing a machine-deployment with particular OS, that OS should be pre-selected (e.g. "Rocky Linux" should be pre-selected if that is used for the existing machine-deployment).

### How to reproduce
Create a cluster with non-default operating system & edit the machine-deployment.

### Environment
- UI Version: latest master
- API Version: : latest master
- Domain:
- Others:

Talha-Jamil-TJ:
Signed-off-by: Talha Jamil TJ <talha.jamil@kubermatic.com>

fixes the spelling of rocky linux from "rockylinux" to "rockyLinux" in **OperatingSystem** enum

Now while editing machine deployments **Rocky Linux** will be selected by default

image

Closes #4571 

### Release note

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/details/cluster/machine-deployment-details/template.html
+++ b/src/app/cluster/details/cluster/machine-deployment-details/template.html
@@ -181,8 +181,8 @@ limitations under the License.
           <km-property-boolean *ngIf="machineDeployment.spec.template.operatingSystem.rhel"
                                [value]="machineDeployment.spec.template.operatingSystem.rhel.distUpgradeOnBoot"
                                label="Upgrade system on the first boot"></km-property-boolean>
-          <km-property-boolean *ngIf="machineDeployment.spec.template.operatingSystem.rockylinux"
-                               [value]="machineDeployment.spec.template.operatingSystem.rockylinux.distUpgradeOnBoot"
+          <km-property-boolean *ngIf="machineDeployment.spec.template.operatingSystem.rockyLinux"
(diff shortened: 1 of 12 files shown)


Waseem826 (changes requested):
While this works fine in Edit Machine Deployment modal but it doesn't work in Machine Deployments table. That's because we need to rename `rockylinux` property in `OperatingSystemSpec` class as well. And I think we should also rename all other occurrences of `rockylinux` with `rockyLinux` (CSS class names etc.) just to be consistent everywhere.

screenshot-localhost_8000-2022 06 14-20_41_11

Talha-Jamil-TJ:
> While this works fine in Edit Machine Deployment modal but it doesn't work in Machine Deployments table. That's because we need to rename `rockylinux` property in `OperatingSystemSpec` class as well. And I think we should also rename all other occurrences of `rockylinux` with `rockyLinux` (CSS class names etc.) just to be consistent everywhere.
> 
> screenshot-localhost_8000-2022 06 14-20_41_11

I have fixed this issue by changing the property is OperatingSystemSpec class.

Waseem826 (approved):
/approve

PR #4606, opened 2022-06-20
update to Go 1.18

xrstf:
This simply updates all build related things from Go 1.17 to Go 1.18. As the dashboard is 99% client-side, this should not have a noticable effect.

### Release note
```release-note
Update to Go 1.18
```


xrstf:
This was doomed to fail. See https://github.com/angular/flex-layout/issues/1407 for more context.

PR #4611, opened 2022-06-21
allow arbitrary cluster names

xrstf:
There is no technical reason to restrict the human readable name for clusters. Those names are only used for display purposes, so this PR simply removes the pattern validation for them.

Fixes https://github.com/kubermatic/kubermatic/issues/6190

### Release note
```release-note
Allow arbitrary human readable cluster names
```

### DIFF
--- a/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -40,7 +40,6 @@ import {Observable, Subject} from 'rxjs';
 import {startWith, switchMap, take, takeUntil, tap} from 'rxjs/operators';
 import * as semver from 'semver';
 import {FeatureGateService} from '@core/services/feature-gate';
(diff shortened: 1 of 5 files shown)


ahmedwaleedmalik:
/unhold

Wanted to check a certain scenario where I thought it could fail. But it should work as expected, good job 👍

Waseem826 (approved):

KhizerRehan:
@xrstf 
Can you please take care of the failing CI jobs `pre-dashboard-check-static`

```
src/app/wizard/step/cluster/component.ts
  138:51  error  Replace `⏎········Validators.required,⏎········Validators.minLength(this._minNameLength),⏎······` with `Validators.required,·Validators.minLength(this._minNameLength)`  prettier/prettier

✖ 1 problem (1 error, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

Solution:
`npm run fix:ts`

PR #4619, opened 2022-06-23
kubelet version selector fallback case fix

multi-io:
Signed-off-by: Olaf Klischat <olaf.klischat@gmail.com>

One-line fix for the Kubelet version selector dropdown component. In the fallback case, i.e. when the list of possible kubelet upgrade versions from the API contains neither the current ("selected") Kubelet version nor the current control plane version of the cluster, the code is supposed to just select the last one of the list of possible Kubelet upgrade versions. It does this by doing `control.setValue(versions.pop())`. versions.pop() however removes the last element from the list, resulting in a selector that's empty by default and is just missing the last (newest) upgrade version in the dropdown. This PR fixes this to `control.setValue(versions[versions.length - 1])`.

Example with this.versions == ["1.19.12", "1.20.14", "1.21.8"], clusterVersion == "1.21.3", this.selected == "1.21.3" --

without the fix:

image

with the fix:

image

### Release note

```release-note
Fixed empty default selection in fallback case of Kubelet version selector
```

### DIFF
--- a/src/app/node-data/kubelet-version/component.ts
+++ b/src/app/node-data/kubelet-version/component.ts
@@ -102,7 +102,7 @@ export class KubeletVersionNodeDataComponent extends BaseFormValidator implement
     }
 
     // As a fallback simply use the newest version available
-    this.form.get(Controls.Kubelet).setValue(this.versions.pop());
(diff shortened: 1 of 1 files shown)


KhizerRehan:
ok-to-test

PR #4625, opened 2022-06-28
Filter subnets and VPCs when dual-stack network is selected for AWS cloud provider

### LINKED ISSUE #4474: Filter out subnets that don't have both CIDRs (IPV4 and IPV6) if "dual stack" is selected in AWS
When dual-stack networking is selected for a user cluster, the KKP UI can help the user with listing only those VPCs and subnets, where dual-stack can be enabled (the VPC must have both IPv4 and IPv6 CIDR address allocated, as well as the subnet).

The API endpoint for listing the VPCs and subnets returns the IPV4 and IPV6 CIDRs for each subnet (attributes "ipv4cidr" and "ipv6cidr").

Therefore, when user selected dual-stack networking, KKP UI should filter VPC / subnets that have both "ipv4cidr" and "ipv6cidr" assigned.

Waseem826:
Only list subnets and VPCs that support IPv6 when dual-stack networking is enabled for AWS.

Closes #4474 

### Release note

```release-note
AWS: Only list subnets and VPCs that support IPv6 when dual-stack networking is enabled.
```

### DIFF
--- a/src/app/wizard/step/provider-settings/provider/basic/aws/component.ts
+++ b/src/app/wizard/step/provider-settings/provider/basic/aws/component.ts
@@ -117,6 +117,13 @@ export class AWSProviderBasicComponent extends BaseFormValidator implements OnIn
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(_ => this.form.reset());
 
+    this._clusterSpecService.clusterChanges
+      .pipe(map(cluster => cluster.spec.clusterNetwork?.ipFamily))
+      .pipe(distinctUntilChanged())
+      .pipe(switchMap(_ => this._vpcListObservable()))
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(this._setDefaultVPC.bind(this));
+
     merge(
       this.form.get(Controls.AccessKeyID).valueChanges,
       this.form.get(Controls.AccessKeySecret).valueChanges,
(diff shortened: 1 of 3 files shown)


src/app/wizard/step/provider-settings/provider/basic/aws/component.ts
```diff
@@ -117,6 +117,14 @@ export class AWSProviderBasicComponent extends BaseFormValidator implements OnIn
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(_ => this.form.reset());
 
+    this._clusterSpecService.clusterChanges
+      .pipe(map(cluster => cluster.spec.clusterNetwork?.ipFamily))
+      .pipe(distinctUntilChanged())
+      .pipe(filter(ipFamily => !!ipFamily))
```
KhizerRehan, line 123: Just for info! what is the purpose of filter operator are we using somewhere? I think we already have dual stack check on `_vpcListObservable`

KhizerRehan:
changes looks good to me tested for AWS 👍🏽

src/app/wizard/step/provider-settings/provider/basic/aws/component.ts
Waseem826: This was added to filter out all `null` or `undefined` values of `ipFamily` but now I think it should be removed.

PR #4628, opened 2022-06-28
Move request-type.ts to fix build warning

Talha-Jamil-TJ:

we were having a warning while creating build

image

### Release note

```release-note
NONE
```


Talha-Jamil-TJ:
> This file should be moved to `test/types` instead of `test/components/types`

File is moved to  `test/types`

Waseem826 (approved):
/approve

PR #4643, opened 2022-07-04
add-project-landing-page-option

### LINKED ISSUE #4510: Provide option to keep Cluster page as the default landing page (rather than project overview)

- [x]  Provide option to keep Cluster page as the default landing page (rather than project overview)
- [x] Show said option on the first opening of the Project overview

https://www.figma.com/file/dJQKgUPzIDND36o2Nq8dct/Loodse-UI?node-id=6976%3A71594
Clustersproject

ahmadhamzh:
adding a new option in user settings to select the project landing page so it appear first when you open project ,(clusters page or project overview page) and also show a message on the first time the user open the project overview page to inform the user with this option

Fixes #4510

```release-note
NONE
```

### DIFF
--- a/src/app/core/services/project.ts
+++ b/src/app/core/services/project.ts
@@ -93,9 +93,13 @@ export class ProjectService {
   }
 
   selectProject(project: Project): Promise<boolean> {
+    let projectLandingPage: string;
     if (project?.status === ProjectStatus.Active) {
       this.onProjectChange.emit(project);
-      return this._router.navigate([`/projects/${project.id}/overview`]);
+      this._userService.currentUser.subscribe(settings => {
+        projectLandingPage = settings.userSettings.useClustersView ? 'clusters' : 'overview';
+      });
+      return this._router.navigate([`/projects/${project.id}/${projectLandingPage}`]);
     }
 
     return this._router.navigate(['/projects']);
--- a/src/app/project-overview/component.ts
+++ b/src/app/project-overview/component.ts
@@ -33,6 +33,7 @@ import {ClusterTemplate} from '@shared/entity/cluster-template';
 import {BackupService} from '@core/services/backup';
 import {EtcdBackupConfig} from '@shared/entity/backup';
 import {Health} from '@shared/entity/health';
+import {CookieService} from 'ngx-cookie-service';
 
 @Component({
   selector: 'km-project-overview',
@@ -57,6 +58,7 @@ export class ProjectOverviewComponent implements OnInit, OnDestroy {
   private _projectChange = new Subject<void>();
   private _unsubscribe = new Subject<void>();
   private readonly _refreshTime = 15;
+  firstVisite: string;
 
   constructor(
     private readonly _projectService: ProjectService,
@@ -67,7 +69,8 @@ export class ProjectOverviewComponent implements OnInit, OnDestroy {
     private readonly _memberService: MemberService,
     private readonly _serviceAccountService: ServiceAccountService,
     private readonly _settingsService: SettingsService,
-    private readonly _appConfigService: AppConfigService
+    private readonly _appConfigService: AppConfigService,
+    private readonly _cookieService: CookieService
   ) {}
 
   ngOnInit(): void {
@@ -80,6 +83,7 @@ export class ProjectOverviewComponent implements OnInit, OnDestroy {
     this._loadSSHKeys();
     this._loadMembers();
     this._loadServiceAccounts();
+    this._checkFirstVisit();
   }
 
   ngOnDestroy(): void {
@@ -189,5 +193,15 @@ export class ProjectOverviewComponent implements OnInit, OnDestroy {
       .pipe(switchMap(() => (this.project ? this._serviceAccountService.get(this.project.id) : EMPTY)))
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(serviceAccounts => (this.serviceAccounts = serviceAccounts));
+  }
+
+  private _checkFirstVisit(): void {
+    this._cookieService.get('firstVisit')
+      ? this.hideFirstVisitMessage()
+      : this._cookieService.set('firstVisit', 'visited', null, '/', 'localhost', false, 'Lax');
+  }
+
+  hideFirstVisitMessage(): void {
+    this.firstVisite = this._cookieService.get('firstVisit');
   }
 }
--- a/src/app/project-overview/style.scss
+++ b/src/app/project-overview/style.scss
@@ -27,3 +27,13 @@
     padding: 0 30px;
   }
 }
+
+.first-visit-message {
+  margin-bottom: 20px;
+  padding: 10px;
+
+  i {
+    position: fixed;
+    right: 40px;
+  }
+}
--- a/src/app/settings/user/style.scss
+++ b/src/app/settings/user/style.scss
@@ -48,6 +48,26 @@
   }
 }
 
+.project-overview-button {
+  @include mixins.size(170px, 45px, $force: true);
+
+  i {
+    margin-right: 10px;
+    min-height: 28px;
+    min-width: 28px;
+  }
+}
+
+.clusters-button {
+  @include mixins.size(110px, 45px, $force: true);
+
+  i {
+    margin-right: 10px;
+    min-height: 28px;
+    min-width: 28px;
+  }
+}
+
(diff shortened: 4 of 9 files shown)


KhizerRehan (changes requested):
Added comments

src/app/project-overview/component.ts
```diff
@@ -190,4 +194,14 @@ export class ProjectOverviewComponent implements OnInit, OnDestroy {
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(serviceAccounts => (this.serviceAccounts = serviceAccounts));
   }
+
+  private _checkFirstVisit(): void {
+    this._cookieService.get('firstVisit')
```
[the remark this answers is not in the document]
ahmadhamzh: this is just to check if the user visit the page before or this is the first visit if first visit then we show a message to tell the user that he can change the project landing page from user settings (the message only showed when first time visit the project overview page ) i think in the future we going to delete all the message after user know the new feature (or after it become old feature)

src/app/project-overview/component.ts
```diff
@@ -190,4 +194,14 @@ export class ProjectOverviewComponent implements OnInit, OnDestroy {
...
+      ? this.hideFirstVisitMessage()
+      : this._cookieService.set('firstVisit', 'visited', null, '/', 'localhost', false, 'Lax');
```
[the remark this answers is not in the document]
ahmadhamzh: no we don't need to delete it because that mean every time the user logout the next time he login the message will show and we don't need that

src/app/settings/user/template.html
```diff
@@ -98,6 +98,26 @@
         <km-spinner-with-confirmation [isSaved]="isEqual(settings.selectedProjectID, apiSettings.selectedProjectID)"
                                       fxFlexAlign=" center"></km-spinner-with-confirmation>
       </div>
+      <div fxLayout="row">
+        <div class="entry-label"
+             fxLayoutAlign=" center">Project Landing Page
+        </div>
+        <mat-button-toggle-group [(value)]="selectedProjectLandingPage"
```
KhizerRehan, line 105: Buttons size looks inconsistent IMO wdyt?
It should be of same width as above -> System, Dark?

@cschieder ?

buttons-size

src/app/project-overview/component.ts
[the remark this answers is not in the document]
KhizerRehan: can we avoid this 3rd party package? we can use local storage simply

KhizerRehan (changes requested):
Added few more comments

src/app/settings/user/template.html
cschieder: I don't think we should make them the same width, it looks strange
image


- Let's use the same padding etc is so the width is dynamic based on content like in the following screenshot
- on your screenshot I believe the project and cluster icon look too large/a bit larger then the rest? That's odd, please take another look
- also maybe we should change the system icon to something different like a cogwheel - wdyt?

image



By the way I'd remove the "Account" headline and the thin line separator so it looks like this,
image

KhizerRehan: Agreed on:
- `CogWheel` icon on System
- Yes! adding button on `Clusters` looks good to me

src/app/project-overview/component.ts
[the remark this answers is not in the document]
KhizerRehan: ah! i thought. You have added a new package but application is already using this package. Okay cool! just wondering Why there was not `package.json` and `updated` lock file

KhizerRehan (changes requested):

Waseem826 (changes requested):

src/app/settings/user/style.scss
```diff
@@ -21,7 +21,7 @@
   }
 
   .mat-card-content {
-    padding: 30px;
+    padding-left: 30px;
```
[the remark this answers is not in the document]
ahmadhamzh: because we delete the `mat-card-header` element which it was above it and now we don't need padding but from the left

src/app/project-overview/component.ts
```diff
@@ -190,4 +195,14 @@ export class ProjectOverviewComponent implements OnInit, OnDestroy {
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(serviceAccounts => (this.serviceAccounts = serviceAccounts));
   }
+
+  private _checkFirstVisitToOverviewPageMessage(): void {
+    this._cookieService.get(this._cookieName)
+      ? this.hideFirstVisitToOverviewPageMessage()
+      : this._cookieService.set(this._cookieName, 'visited', null, '/', 'localhost', false, 'Lax');
```
KhizerRehan, line 202: I guess this won't work on Dev since `hardcode` binding of **Domain** it should be `window.location.hostname` 

cc: @Waseem826

src/app/project-overview/style.scss
```diff
@@ -27,3 +27,13 @@
     padding: 0 30px;
   }
 }
+
+.first-visit-overview-message {
+  margin-bottom: 20px;
+  padding: 10px;
+
+  i .km-icon-close {
```
KhizerRehan, line 35: Update alignment according to design + `x` icon

icon

src/app/settings/user/template.html
```diff
@@ -27,10 +27,6 @@
 </div>
 
 <mat-card>
-  <mat-card-header>
```
KhizerRehan, line 30: why this is removed?

src/app/settings/user/template.html
```diff
@@ -98,6 +94,27 @@
...
       </div>
+      <div fxLayout="row">
+        <div class="entry-label"
+             fxLayoutAlign=" center">Project Landing Page
+        </div>
+        <mat-button-toggle-group [(value)]="selectedProjectLandingPage"
+                                 (change)="changingLandingPage($event.value)"
+                                 fxLayout="row wrap">
+          <mat-button-toggle class="project-overview-button"
+                             value="overview">
```
KhizerRehan, line 105: Use View.Overview

src/app/settings/user/template.html
```diff
@@ -98,6 +94,27 @@
...
+            <i class="km-icon-mask km-icon-projects-card"></i>
+            Project Overview
+          </mat-button-toggle>
+          <mat-button-toggle class="clusters-button"
+                             value="clusters">
```
KhizerRehan, line 110: Use View.Clusters

KhizerRehan (changes requested):
Added more feedback

src/app/settings/user/template.html
ahmadhamzh: because as @cschieder suggest to remove the "Account" headline and the thin line separator so it looks like this,
177564923-da5b72cc-e58b-4514-a22f-8158edab81a5

src/app/settings/user/template.html
```diff
@@ -98,6 +94,27 @@
         <km-spinner-with-confirmation [isSaved]="isEqual(settings.selectedProjectID, apiSettings.selectedProjectID)"
                                       fxFlexAlign=" center"></km-spinner-with-confirmation>
       </div>
+      <div fxLayout="row">
+        <div class="entry-label"
+             fxLayoutAlign=" center">Project Landing Page
+        </div>
+        <mat-button-toggle-group [(value)]="selectedProjectLandingPage"
+                                 (change)="changingLandingPage($event.value)"
```
KhizerRehan, line 102: can we rename to 
```suggestion
  (change)=" onLandingPageChange($event.value)`
```

e.g we have
- onSettingsChange
- onThemeChange similarly as above

src/app/core/services/project.ts
```diff
@@ -93,12 +94,16 @@ export class ProjectService {
...
 
   selectProject(project: Project): Promise<boolean> {
+    let projectLandingPage: string;
     if (project?.status === ProjectStatus.Active) {
       this.onProjectChange.emit(project);
-      return this._router.navigate([`/projects/${project.id}/overview`]);
+      this._userService.currentUser.subscribe(settings => {
+        projectLandingPage = settings.userSettings.useClustersView ? View.Clusters : View.Overview;
+      });
+      return this._router.navigateByUrl(`/projects/${project.id}/${projectLandingPage}`);
```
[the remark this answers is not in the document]
Waseem826: Yes

PR #4671, opened 2022-07-13
Display multiple external IPs in node details of machine deployment

### LINKED ISSUE #4631: Allow multiple ExternalIPs in node details of machine deployments

In dual-stack case, it is common for nodes to have multiple ExternalIPs: one IPv4 and one IPv6.

KKP dashboard in this case shows only one of them, e.g.:

azure-external-ip

In such case, both ExternalIPs should be shown.

Related https://github.com/kubermatic/dashboard/issues/4529

Waseem826:
Display multiple external IPs in node details of machine deployment.

Fixes #4631 

**Release Note**:

```release-note
Display multiple external IPs in node details of machine deployment.
```

### DIFF
--- a/src/app/cluster/details/cluster/node-list/template.html
+++ b/src/app/cluster/details/cluster/node-list/template.html
@@ -88,7 +88,7 @@ limitations under the License.
           <ng-container *ngIf="element.status.addresses && getAddresses(element) as addresses">
             <div fxLayout="column"
                  fxLayoutGap="5px"
-                 [ngClass]="{'node-ip-addresses-spacing': (addresses.externalIP && addresses.internalIPs.length) || addresses.internalIPs.length > 1}">
+                 [ngClass]="{'node-ip-addresses-spacing': (addresses.externalIPs.length && addresses.internalIPs.length) || addresses.internalIPs.length > 1 || addresses.externalIPs.length > 1}">
               <div *ngIf="!!addresses.internalIPs.length"
                    fxLayout="row"
                    fxLayoutGap="5px">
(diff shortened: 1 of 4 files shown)


src/app/cluster/details/cluster/node-list/template.html
```diff
@@ -88,7 +88,7 @@
           <ng-container *ngIf="element.status.addresses && getAddresses(element) as addresses">
             <div fxLayout="column"
                  fxLayoutGap="5px"
-                 [ngClass]="{'node-ip-addresses-spacing': (addresses.externalIP && addresses.internalIPs.length) || addresses.internalIPs.length > 1}">
+                 [ngClass]="{'node-ip-addresses-spacing': (addresses.externalIPs.length && addresses.internalIPs.length) || addresses.internalIPs.length > 1 || addresses.externalIPs.length > 1}">
```
KhizerRehan, line 91: Shouldn't this check be simple
In both cases it will be added.
addresses.externalIPs.length && addresses.internalIPs.length -> 1, 1 => true => class added
OR
addresses.externalIPs.length || addresses.internalIPs.length  -> 0, 1 -> true => class added

IMO, just adding -> `addresses.externalIPs.length  || addresses.internalIPs.length` will work

to simple => All scenarios can be covered by Above check
- 0,0 => false 
- 0,1 => true
- > 1,1 or > 1,> 1 => true
thoughts?

KhizerRehan (changes requested):
Added small comment

src/app/cluster/details/cluster/node-list/template.html
Waseem826: We don't want to add this class in cases 0,1 or 1,0. We only want this when there are at least 2 entries in the column i.e. 1 Int. IP, 1 Ext. IP or 2 Int. IPs, 0 Ext. IP or 2 Ext.IPs, 0 Int. IP

KhizerRehan: got it! thanks

PR #4675, opened 2022-07-14
Fix the functionality to add/edit backup destinations

ahmedwaleedmalik:

This PR fixes the broken functionality for adding new “Backup Destinations”. `destination` and `configuration` in `getObservable` method were being shallow cloned and thus retaining object references. Which leads to a map that adds each input character as a key and thus breaks everything. They should be deep cloned, which fixes the issues.

Fixes #

The bug:

Screenshot 2022-07-07 at 10 19 26 PM
Screenshot 2022-07-07 at 10 20 02 PM

**Release Note**:

```release-note
NONE
```


Waseem826 (approved):
/approve

PR #4682, opened 2022-07-15
Add IPv6 subnet ID and IPv6 subnet pool for OpenStack cluster provider

### LINKED ISSUE #4469: Add IPv6 subnet and IPv6 subnet pool boxes to OpenStack cluster wizard
In case that dual-stack is enabled for an user cluster, OpenStack cluster wizard should include two additional input boxes (both optional) on the Settings page:

 - IPv6 subnet ID (so there would be two subnet ID boxes - IPv4 & IPv6, only IPv4 subnets should be listed in the IPv4 box and IPv6 subnets in the IPv6 box)
 - IPv6 subnet pool (API endpoint for listing subnet pools to be done in https://github.com/kubermatic/kubermatic/pull/9809): `/api/v2/providers/openstack/subnetpools`

These options should be visible only when dual-stack is selected in the previous steps.

Screenshot 2022-06-09 at 10 29 37

Waseem826:
Add IPv6 subnet ID and IPv6 subnet pool for OpenStack cluster provider.

screenshot-localhost_8000-2022 07 15-16_41_10

Fixes #4469 

**Release Note**:

```release-note
Add IPv6 subnet ID and IPv6 subnet pool for OpenStack cluster provider.
```

### DIFF
--- a/src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
+++ b/src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
@@ -23,39 +23,37 @@ import {
   ViewChild,
 } from '@angular/core';
 import {FormBuilder, NG_VALIDATORS, NG_VALUE_ACCESSOR} from '@angular/forms';
-import {OpenstackCredentialsTypeService} from '@app/wizard/step/provider-settings/provider/extended/openstack/service';
 import {ClusterSpecService} from '@core/services/cluster-spec';
 import {PresetsService} from '@core/services/wizard/presets';
 import {FilteredComboboxComponent} from '@shared/components/combobox/component';
-import {OpenstackNetwork, OpenstackSecurityGroup, OpenstackSubnet} from '@shared/entity/provider/openstack';
+import {Cluster} from '@shared/entity/cluster';
+import {
+  OpenstackNetwork,
+  OpenstackSecurityGroup,
+  OpenstackSubnet,
+  OpenstackSubnetPool,
+} from '@shared/entity/provider/openstack';
 import {NodeProvider} from '@shared/model/NodeProviderConstants';
 import {BaseFormValidator} from '@shared/validators/base-form.validator';
 import _ from 'lodash';
 import {EMPTY, merge, Observable, onErrorResumeNext} from 'rxjs';
 import {catchError, debounceTime, filter, map, switchMap, takeUntil, tap} from 'rxjs/operators';
+import {OpenstackCredentialsTypeService} from '../service';
+import {IPVersion} from '../shared/types/ip-version';
+import {
+  IPv4SubnetIDState,
+  IPv6SubnetIDState,
+  IPv6SubnetPoolState,
+  NetworkState,
+  SecurityGroupState,
+} from '../shared/types/state';
 
 enum Controls {
   SecurityGroup = 'securityGroup',
   Network = 'network',
-  SubnetID = 'subnetID',
-}
-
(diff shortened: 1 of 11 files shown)


src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
```diff
@@ -181,8 +224,13 @@ export class OpenstackProviderExtendedDefaultCredentialsComponent
     this._unsubscribe.complete();
   }
 
-  subnetIDDisplayName(id: string): string {
-    const subnetID = this.subnetIDs.find(subnetID => subnetID.id === id);
+  ipv4SubnetIDDisplayName(id: string): string {
```
KhizerRehan, line 227: can we move to shared dir `src/app/wizard/step/provider-settings/provider/extended/openstack/shared`?
to avoid duplication? or I guess we will still need methods to to set values within a component.

thoughts?

- ipv4SubnetIDDisplayName
- ipv6SubnetIDDisplayName
- _loadSecurityGroups
- _loadNetworks
- _loadSubnetIDs

src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
```diff
@@ -23,39 +23,37 @@ import {
...
-  Ready = 'Network',
-  Loading = 'Loading...',
-  Empty = 'No Networks Available',
-}
-
-enum SubnetIDState {
-  Ready = 'Subnet ID',
-  Loading = 'Loading...',
-  Empty = 'No Subnet IDs Available',
+  IPv4SubnetID = 'subnetID',
```
KhizerRehan, line 54:
```suggestion
  IPv4SubnetID = 'ipv4subnetID',  ? // IMO, it will make clear that we are setting value for `IPv4`
```
similarly in extended too.

i guess it won't effect because we are setting value from `onIPv4SubnetIDChange` which recieves value in an event

KhizerRehan (commented):
General Feedback otherwise looks good to me

KhizerRehan:
@Waseem826 i can see `IPv6` enabled on cloud spec on `/api/v1/dc` api call but I can't see Dual Stack option on while creating cluster? am i doing something wrong?

openstack

Waseem826:
> @Waseem826 i can see `IPv6` enabled on cloud spec on `/api/v1/dc` api call but I can't see Dual Stack option on while creating cluster? am i doing something wrong?
> 
> openstack
>  

Which datacenter?

src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
Waseem826: There's lot more code duplication between this component and `applications/component.ts` so we have to refactor that in a separate ticket. Regarding these methods, this structure has been followed in other provider components too so we'll have to refactor that in separate ticket if we want.

KhizerRehan:
> > @Waseem826 i can see `IPv6` enabled on cloud spec on `/api/v1/dc` api call but I can't see Dual Stack option on while creating cluster? am i doing something wrong?
> > openstack
> 
> Which datacenter?

DualStack option only available in `Syseleven fes`
2b50b13db64128047f619afeaf520c35

src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
Waseem826: Fixed.

PR #4705, opened 2022-07-21
Disallow IPVS proxy mode for Cilium CNI

### LINKED ISSUE #4687: Disallow IPVS for Cilium CNI

Because of known issue of Cilium CNI and IPVS kube-proxy mode (https://github.com/kubermatic/kubermatic/issues/8767) we should not advise our users to use it for production. Therefore, we should not offer `ipvs` as a Proxy Mode option in case that Cilium CNI is selected during cluster creation:

image

Instead, as a special case for CIlium, if `ipvs` is returned as the default proxy mode, we should:
 - select `iptables` if Konnectivity is NOT enabled
 - select `ebpf` if Konnectivity IS enabled.

Waseem826:
Disallow `ipvs` proxy mode when `Cilium` CNI is selected.

Fixes #4687 

**Release Note**:

```release-note
Disallow IPVS proxy mode when `Cilium` CNI is selected.
```

### DIFF
--- a/src/app/wizard/step/cluster/component.ts
+++ b/src/app/wizard/step/cluster/component.ts
@@ -127,6 +127,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
   private _datacenterSpec: Datacenter;
   private _seedSettings: SeedSettings;
   private _settings: AdminSettings;
+  private _defaultProxyMode: ProxyMode;
   private readonly _minNameLength = 5;
   private readonly _defaultAllowedIPRange = '0.0.0.0/0';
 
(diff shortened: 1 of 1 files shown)


src/app/wizard/step/cluster/component.ts
```diff
@@ -480,6 +469,30 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
...
+        if (proxyModeControl.value === ProxyMode.ipvs) {
+          newValue = ProxyMode.ebpf;
+        }
+      } else {
+        this.availableProxyModes = [ProxyMode.iptables];
+        newValue = ProxyMode.iptables;
+      }
+    } else {
+      this.availableProxyModes = [ProxyMode.ipvs, ProxyMode.iptables];
+      if (proxyModeControl.pristine || !this.availableProxyModes.includes(proxyModeControl.value)) {
```
Waseem826, line 487: Reason for adding `pristine` check is to avoid unintentional value changes when user switches between `Canal` and `Cilium`. When user switches to `Cilium`, `iptables` or `ebpf` proxy mode will be selected automatically but then if they switch back to `Canal` then `ipvs` will be reselected. Since Proxy Mode is hidden under Advanced settings so the user might not even notice this change (since they never manually changed Proxy Mode value).
This is demonstrated in attached video:


https://user-images.githubusercontent.com/13975988/180172378-1be2ea6e-f630-44e9-a1b9-26900099b353.mov

PR #4710, opened 2022-07-22
migrate to yq4

xrstf:
It's time to migrate to yq4. See https://mikefarah.gitbook.io/yq/upgrading-from-v3 for more information.

I chose to make it explicit what version we're using for now, in order to slowly get rid of yq3 over the next few PRs. It is a bit annoying that devs would need a  `yq4` binary on their machines, but I suspect many have v3 and v4 around, like I do.

I have no idea why exactly we had to YAML-to-JSON the kubeconfig, and also why it should not have been encoded. So I simply removed `yq` here alltogether.

Fixes #

**Release Note**:

```release-note
NONE
```

### DIFF
--- a/hack/e2e/run-tests.sh
+++ b/hack/e2e/run-tests.sh
@@ -42,7 +42,7 @@ if [ $CYPRESS_MOCKS != "true" ]; then
   source hack/e2e/setup-kubermatic-in-kind.sh
 
   export CYPRESS_SEED_NAME="${SEED_NAME}"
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/lgtm

ahmedwaleedmalik:
@xrstf the full-e2e tests will/can be ignored. They are broken but they should be fixed in a separate PR. I've already communicated that to `sig-ui`.

Waseem826 (approved):
/approve

PR #4716, opened 2022-07-25
Remove delete functionality for imported cluster

KhizerRehan:

This PR doesn't show `delete` icon  when cluster is `imported`:

- on List page
- on Details page

https://user-images.githubusercontent.com/17727069/180807427-59936c98-4137-44b5-93c9-e04fecfd31a9.mp4

Fixes #

```txt
{
    "isImported": "true" || "false"
    "project-id": "jgm2x2q9rc"
}
```
Since `string` value is returned from Backend so had to implicit check of string on UI

**Release Note**:

```release-note
NONE
```

### DIFF
--- a/src/app/cluster/details/external-cluster/template.html
+++ b/src/app/cluster/details/external-cluster/template.html
@@ -35,7 +35,8 @@ limitations under the License.
         <span>Disconnect</span>
       </button>
 
-      <button id="km-delete-external-cluster-btn"
+      <button *ngIf="!cluster?.labels || cluster.labels['isImported'] !== 'true'"
+              id="km-delete-external-cluster-btn"
               mat-flat-button
               color="tertiary"
               fxLayoutAlign="center center"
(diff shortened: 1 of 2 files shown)


src/app/cluster/details/external-cluster/template.html
```diff
@@ -35,7 +35,8 @@
         <span>Disconnect</span>
       </button>
 
-      <button id="km-delete-external-cluster-btn"
+      <button *ngIf="cluster?.labels['isImported'] === 'false'"
```
Waseem826, line 38: What if there is no `isImported` label or no labels at all? I think this check should be like this:

```suggestion
      <button *ngIf="!cluster?.labels || cluster.labels['isImported'] !== 'true'"
```

Waseem826 (changes requested):

Waseem826 (approved):
/approve

PR #4717, opened 2022-07-25
Skip polling for machine deployments when cluster is marked for deletion

ahmedwaleedmalik:

On the list cluster view, we keep polling for machine deployments even if the cluster is marked for deletion. This is just a wasteful API call that results in an undesired error message.

Fixes #

Screenshot 2022-07-25 at 10 58 24 PM

**Release Note**:

```release-note
NONE
```


Waseem826 (approved):
/approve

PR #4726, opened 2022-07-28
Incorrect number of secondary disks

### LINKED ISSUE #4721: Kubevirt: incorrect number of secondary disk added during cluster creation
### What happened
In kubevirt user cluster creation wizard, add 2 secondary disks:
Wizard in Project 'sankalp' (2)

Then remove one secondary disk:
Wizard in Project 'sankalp' (3)

After clicking on next, still there are 2 secondary disks added which is incorrect:
Wizard in Project 'sankalp' (4)

Also for the existing cluster, after clicking on edit Machine Deployment, UI doesn't display the existing number of secondary disks:
Cluster 'serene-almeida' in Project 'sankalp'
If the user will click on `Save Changes`, all the existing secondary disks will be removed which is not desirable.

### Expected behavior
A correct number of secondary disks should be added.

### How to reproduce
Try to create a cluster with kubevirt provider and add/remove secondary disks as described above.

### Environment
- UI Version: master
- API Version: master
- Domain:
- Others:

Waseem826:
This PR fixes 2 issues:

1. When removing secondary disks from initial nodes step in wizard, value was not being updated in `nodeData`.
2. Secondary Disks were not shown in edit machine deployment dialog.

Fixes #4721 

**Release Note**:

```release-note
Display configured secondary disks in edit machine deployment dialog.
```

### DIFF
--- a/src/app/shared/components/combobox/component.ts
+++ b/src/app/shared/components/combobox/component.ts
@@ -144,5 +144,6 @@ export class FilteredComboboxComponent extends BaseFormValidator implements OnIn
 
   writeValue(value: string | string[]) {
     this.form.get(ComboboxControls.Select).setValue(value, {emitEvent: false});
+    this.selected = value;
   }
 }
--- a/src/app/node-data/basic/provider/kubevirt/component.ts
+++ b/src/app/node-data/basic/provider/kubevirt/component.ts
@@ -159,6 +159,16 @@ export class KubeVirtBasicNodeDataComponent
     this.form.valueChanges
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(_ => (this._nodeDataService.nodeData = this._getNodeData()));
+
+    this.form
+      .get(Controls.SecondaryDisks)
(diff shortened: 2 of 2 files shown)


src/app/shared/components/combobox/component.ts
```diff
@@ -144,5 +144,6 @@ export class FilteredComboboxComponent extends BaseFormValidator implements OnIn
 
   writeValue(value: string | string[]) {
     this.form.get(ComboboxControls.Select).setValue(value, {emitEvent: false});
+    this.selected = value;
```
Waseem826, line 147: Reason:
If `this.selected` is not passed on from parent component then when `ngOnChanges` method is called (due to `options` Input being changed) it overrides the selected value because of:
```
    this.form.get(ComboboxControls.Select).setValue(this.selected);
```
`writeValue` is called when value of this form control is updated in the parent component. So by updating `this.selected` here we make sure that the value is not overridden.

Talha-Jamil-TJ (approved):
/lgtm

PR #4739, opened 2022-07-29
List custom operating system profiles

Waseem826:
screenshot-localhost_8000-2022 07 29-20_33_55

Fixes #

**Release Note**:

```release-note
NONE
```

### DIFF
--- a/src/app/node-data/component.ts
+++ b/src/app/node-data/component.ts
@@ -27,15 +27,18 @@ import {ClusterSpecService} from '@core/services/cluster-spec';
 import {DatacenterService} from '@core/services/datacenter';
 import {NameGeneratorService} from '@core/services/name-generator';
 import {NodeDataService} from '@core/services/node-data/service';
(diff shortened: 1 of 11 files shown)


src/app/node-data/component.ts
```diff
@@ -286,6 +311,32 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
...
+                project.id
+              );
+            })
+          )
+        : this._datacenterSpec?.spec
+        ? this._osmService.getOperatingSystemProfilesForSeed(this._datacenterSpec.spec.seed)
+        : EMPTY;
+
+      profiles$
+        .pipe(takeUntil(this._unsubscribe))
```
Waseem826, line 331: Added this so that if API call is pending and component is destroyed, API call is cancelled.

ahmedwaleedmalik (approved):
/approve

Thank you for taking care of this, good job 👍

PR #4764, opened 2022-08-03
Fixed quota-widget width on extended view

### LINKED ISSUE #4750: Fix Quota widget in small screens

Currently, the Quota widget extends to other right half on Wizard half on `initial Nodes` step.

https://user-images.githubusercontent.com/17727069/182336877-3ce44cc0-d3ef-4bbe-aa28-03f1fa10e8d6.mp4

- Use breakpoints for small screen

Talha-Jamil-TJ:

Fixes incorrect width of quota widget progress bars on small screens

> Before 
image

> After
image

Fixes #4750

**Release Note**:

```release-note
NONE
```

### DIFF
--- a/src/app/dynamic/enterprise/quotas/quota-widget/template.html
+++ b/src/app/dynamic/enterprise/quotas/quota-widget/template.html
@@ -20,20 +20,21 @@
 END OF TERMS AND CONDITIONS
 -->
 
-<div class="km-quota-widget"
+<div class="km-quota-widget-container"
      *ngIf="quotaDetails">
   <ng-container *ngTemplateOutlet="showQuotaWidgetDetails ? quotaWidgetDetails : quotaWidget"></ng-container>
 </div>
 
 <ng-template #quotaWidget>
-  <mat-card *ngIf="showAsCard; else progressQuota">
+  <mat-card *ngIf="showAsCard; else quotaWidgetWithoutCard">
     <mat-card-content>
-      <ng-container *ngTemplateOutlet="progressQuota"></ng-container>
+      <ng-container *ngTemplateOutlet="quotaWidgetWithoutCard"></ng-container>
     </mat-card-content>
   </mat-card>
 
-  <ng-template #progressQuota>
-    <div fxLayout="row"
+  <ng-template #quotaWidgetWithoutCard>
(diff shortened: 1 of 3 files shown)


src/app/dynamic/enterprise/quotas/quota-widget/template.html
```diff
@@ -54,7 +55,7 @@
       </ng-container>
     </div>
 
-    <mat-card class="km-detail-view"
+    <mat-card class="km-compact-detail-on-hover"
```
Waseem826, line 58:
```suggestion
    <mat-card class="km-quota-detail"
```

src/assets/css/global/_main.scss
```diff
@@ -618,13 +618,31 @@ km-cni-version {
...
+        margin: 2px 5px;
+      }
     }
   }
 
   .mat-progress-bar .mat-progress-bar-fill {
     transition: none;
   }
+
+  .km-extended-widget {
```
Waseem826, line 637: This style looks redundant because this is already handled by `km-quota-widget` style above. Or am I not understanding it right?

src/assets/css/global/_main.scss
```diff
@@ -618,13 +618,31 @@ km-cni-version {
...
 .km-quota-widget {
-  .km-quota-item {
+  km-property {
+    width: 100%;
+
     .container {
-      margin: 2px 5px;
+      width: 100%;
+
+      &:not(last-child) {
```
Waseem826, line 627: This selector has no effect apparently. It applies this margin to all `km-property` elements.

Waseem826 (changes requested):

src/app/dynamic/enterprise/quotas/quota-widget/template.html
Talha-Jamil-TJ: I think `km-quota-hover-details` will be more appropriate.

Waseem826: `hover` should not be in class name. If we want to style something on hover then it should be done using `:hover` selector

Waseem826 (approved):
/approve

PR #4800, opened 2022-08-10
Change autocomplete to km select

### LINKED ISSUE #4774: Update list fields to use km-select in AKS/EKS/GKE
Update fields which have list to with km-select:
- this will fix the issue of removing the selected value to get the list again
- also get the delete sign to remove the selected value
- should also fix the issue of popup `GKE Zone cannot be empty`

ahmadhamzh:
change the autocomplete component in some field that just need to show a list to km-select component and fix the  issue for trimming .patch (except .0 ))
Fixes #4774 #4821

**What type of PR is this?**
/kind bug

-->
```release-note
NONE
```
**Documentation**:
```documentation
NONE
```

### DIFF
--- a/src/app/external-cluster-wizard/steps/external-cluster/provider/aks/component.ts
+++ b/src/app/external-cluster-wizard/steps/external-cluster/provider/aks/component.ts
@@ -251,8 +251,11 @@ export class AKSClusterSettingsComponent
   }
 
   private _updateExternalClusterModel(): void {
-    const indexPosition = 2;
-    const version = this.controlValue(Controls.KubernetesVersion)?.main;
+    let version = this.controlValue(Controls.KubernetesVersion)?.main;
+    const versionSplitArr = version?.split('.');
+    if (versionSplitArr?.length && !(versionSplitArr[2] > 0)) {
+      version = versionSplitArr[0] + '.' + versionSplitArr[1];
+    }
 
     const config = {
       name: this.controlValue(Controls.Name),
@@ -265,7 +268,7 @@ export class AKSClusterSettingsComponent
       } as ExternalCloudSpec,
       spec: {
         aksclusterSpec: {
-          kubernetesVersion: version?.slice(0, version.indexOf('.', indexPosition)),
+          kubernetesVersion: version,
           location: this.controlValue(Controls.Location),
           machineDeploymentSpec: {
             name: this.controlValue(Controls.NodePoolName),
@@ -277,7 +280,7 @@ export class AKSClusterSettingsComponent
             } as AgentPoolBasics,
           } as AKSMachineDeploymentCloudSpec,
         } as AKSClusterSpec,
-        version: version?.slice(0, version.indexOf('.', indexPosition)),
+        version: version,
       } as ExternalClusterSpec,
     } as ExternalClusterModel;
 
--- a/src/app/external-cluster-wizard/steps/external-cluster/provider/eks/component.ts
+++ b/src/app/external-cluster-wizard/steps/external-cluster/provider/eks/component.ts
@@ -147,34 +147,35 @@ export class EKSClusterSettingsComponent
       this._externalMachineDeploymentService.isAddMachineDeploymentFormValid = this.form.valid;
     });
 
-    this.form
-      .get(Controls.Vpc)
-      .valueChanges.pipe(debounceTime(this._debounceTime))
-      .pipe(takeUntil(this._unsubscribe))
-      .subscribe(_ => {
-        const vpc = this.controlValue(Controls.Vpc)?.main;
-        if (vpc) {
-          this._onVPCSelectionChange(vpc);
-        }
-        this.control(Controls.SubnetIds).setValue([]);
-        this.control(Controls.SecurityGroupsIds).setValue([]);
-      });
-
     if (this.isDialogView()) {
       const version = this.cluster.spec.version;
-      this.control(Controls.SecurityGroupsIds).removeValidators(Validators.required);
+      this.control(Controls.Vpc).clearValidators();
+      this.control(Controls.SecurityGroupsIds).clearValidators();
       this.control(Controls.Version).setValue(version.slice(1, version.indexOf('-')));
       this.control(Controls.Version).disable();
       this.control(Controls.Vpc).setValue(this.cluster.spec.eksclusterSpec.vpcConfigRequest.vpcId);
       this.control(Controls.Vpc).disable();
 
       this._externalMachineDeploymentService
-        .getEKSSubnetsForMachineDeployment(this.projectID, this.cluster.id, this.control(Controls.Vpc).value)
+        .getEKSSubnetsForMachineDeployment(this.projectID, this.cluster.id, this.controlValue(Controls.Vpc))
         .subscribe((data: string[]) => {
           this.subnetIds = data;
         });
     } else {
       this._getEKSKubernetesVersions();
+
+      this.form
+        .get(Controls.Vpc)
+        .valueChanges.pipe(debounceTime(this._debounceTime))
+        .pipe(takeUntil(this._unsubscribe))
+        .subscribe(_ => {
+          const vpc = this.controlValue(Controls.Vpc);
+          if (vpc) {
+            this._onVPCSelectionChange(vpc);
+          }
+          this.control(Controls.SubnetIds).setValue([]);
+          this.control(Controls.SecurityGroupsIds).setValue([]);
+        });
     }
 
     this._externalClusterService.presetChanges.pipe(takeUntil(this._unsubscribe)).subscribe(preset => {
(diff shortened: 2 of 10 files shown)


Waseem826:
Why can't we use `km-combobox` component instead of creating new `km-select` component?

ahmadhamzh:
> Why can't we use `km-combobox` component instead of creating new `km-select` component?

km-select it's a simple `mat-select` component for just listing array of string 
km-combobox is more to deal with more complected list that we add some filtering in it and some event emitter to deal with the value selected which it's not the best solution for our case  and to use it need changes like in the reset method and dealing with string array not object array

src/app/shared/components/km-select/template.html
```diff
@@ -0,0 +1,28 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<form [formGroup]="form">
+  <mat-form-field>
+    <mat-label>{{options.length > 0 ? label : 'no '+label+' Available'}}</mat-label>
```
KhizerRehan, line 18: better to use *ngIf as done on other shared components
```suggestion
<mat-label *ngIf="label">{{label}}</mat-label>
```

src/app/shared/components/km-select/component.ts
```diff
@@ -0,0 +1,62 @@
...
+})
+export class KmSelectComponent extends BaseFormValidator implements OnInit {
+  @Input() options: string[];
+  @Input() label: string;
+  @Input() isLoading = false;
+  @Input() hint: string;
+  @Input() required: boolean;
+  @Input() validators: ValidatorFn[] = [];
+  form: FormGroup;
+  readonly controls = KmSelect;
```
KhizerRehan, line 47:
```ts
readonly Controls = Controls;
```

src/app/shared/components/km-select/component.ts
```diff
@@ -0,0 +1,62 @@
...
+      multi: true,
+    },
+    {
+      provide: NG_VALIDATORS,
+      useExisting: forwardRef(() => KmSelectComponent),
+      multi: true,
+    },
+  ],
+})
+export class KmSelectComponent extends BaseFormValidator implements OnInit {
```
KhizerRehan, line 39: Shouldn't it implement following interface?
```suggestion
ControlValueAccessor, Validator
```

see following shared component
- ChipListComponent
- ChipAutocompleteComponent

KhizerRehan (changes requested):
PTAL

src/app/external-cluster-wizard/steps/external-cluster/provider/aks/component.ts
```diff
@@ -251,8 +251,11 @@ export class AKSClusterSettingsComponent
   }
 
   private _updateExternalClusterModel(): void {
-    const indexPosition = 2;
-    const version = this.controlValue(Controls.KubernetesVersion)?.main;
+    let version = this.controlValue(Controls.KubernetesVersion)?.main;
+    const versionSplitArr = version?.split('.');
+    if (versionSplitArr?.length && !(versionSplitArr[2] > 0)) {
```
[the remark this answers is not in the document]
ahmadhamzh: im checking if patch version > 0 then send all the input (e.g if it 1.22.1 -> it will show 1.22.1
and if it 1.22.0 -> it will show 1.22)
so if patch version > 0 then don't go inside the if statement and version will be the same as it is in the field

src/app/external-cluster-wizard/steps/external-cluster/provider/gke/component.ts
```diff
@@ -129,9 +129,9 @@ export class GKEClusterSettingsComponent
   }
 
   ngOnDestroy(): void {
-    this.reset();
     this._unsubscribe.next();
     this._unsubscribe.complete();
+    this.reset();
```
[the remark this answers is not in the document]
ahmadhamzh: when reset. all the fields will be empty and yet not been unsubscribe so the api request for versions will be sent as empty  zones so the popup notification for wrong request will be shown so we have to unsubscribe first and then reset.

is there any  problem that been shown because of this change ?

src/app/external-cluster-wizard/steps/external-cluster/provider/eks/component.ts
```diff
@@ -204,7 +204,12 @@ export class EKSClusterSettingsComponent
         finalize(() => (this.isLoadingVpcs = false))
       )
       .subscribe((vpcs: EKSVpc[]) => {
-        this.vpcs = vpcs.map(vpc => vpc.id);
+        this.vpcs = vpcs.map(vpc => {
```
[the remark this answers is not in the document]
Waseem826: I think this not necessary because `EKSVpc` type is added implicitly, reason being that we've mentioned type of `vpcs` above.

src/app/shared/components/km-select/component.ts
ahmadhamzh: im not using any methods from those interface

src/app/external-cluster-wizard/steps/external-cluster/provider/eks/template.html
```diff
@@ -118,7 +117,7 @@
              [formControlName]="Controls.Vpc" />
     </mat-form-field>
 
-    <ng-container *ngIf="isDialogView() ? true : form.get(Controls.Vpc).value?.main">
+    <ng-container *ngIf="isDialogView() ? true : form.get(Controls.Vpc).value">
```
Talha-Jamil-TJ, line 120:
```suggestion
    <ng-container *ngIf="isDialogView() || form.get(Controls.Vpc).value">
```

Talha-Jamil-TJ (commented):
Changes requested

src/app/external-cluster-wizard/steps/external-cluster/provider/gke/component.ts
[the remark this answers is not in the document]
Talha-Jamil-TJ: All forms related logic should stay above unsubscribe logic

src/app/shared/components/km-select/component.ts
Talha-Jamil-TJ: using `BaseFormValidator` is also fine, and is being used in some components too, we should encourage it's use as it will reduce code in each component.
image

src/app/external-cluster-wizard/steps/external-cluster/provider/eks/template.html
ahmadhamzh: actually i will just make it  `<ng-container *ngIf="form.get(Controls.Vpc).value">` because we already have it in both views and it's only been displayed when VPC have value 
@KhizerRehan right ?

src/app/external-cluster-wizard/steps/external-cluster/provider/aks/component.ts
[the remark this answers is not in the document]
Waseem826: BTW you can also use `patch` method of `semver` package that we have in our `package.json` instead of manual parsing. Something for the future perhaps.

src/app/external-cluster-wizard/steps/external-cluster/provider/eks/template.html
KhizerRehan: right! but i have update in my PR anyways i will resolve issue.

This things is changed in https://github.com/kubermatic/dashboard/pull/4838 PR

src/app/shared/components/km-select/component.ts
KhizerRehan: `BaseFormValidator` is fine but we need to override `registerOnTouched` instead of subscribing to whole form we need to subscribe to individual control so that when value is changed. Value received on parent should be only that control value instead of `obj` e.g km-autocomplete -> we pluck value using `main?.` we don't want that behaviour

Waseem826: `BaseFormValidator` is already implementing `ControlValueAccessor` and `Validator` so if you want to override `registerOnTouched` method you can already do that in this class. (this is not related to later part of the comment)

src/app/shared/components/select/component.ts
```diff
@@ -0,0 +1,84 @@
...
+      multi: true,
+    },
+    {
+      provide: NG_VALIDATORS,
+      useExisting: forwardRef(() => SelectComponent),
+      multi: true,
+    },
+  ],
+})
+export class SelectComponent extends BaseFormValidator implements OnInit, OnDestroy, ControlValueAccessor {
```
Waseem826, line 48: No need to implement `ControlValueAccessor` here because its already implemented by `BaseFormValidator` class

PR #4442, opened 2022-04-29
Migration of km throttle click to km-button component

ahmadhamzh:
What this PR does / why we need it
replace all the kmThrottelClick directives with the km-button component in the code

Which issue(s) this PR fixes
"closes #4379"

```release-note
NONE
```

PR #4808, opened 2022-08-11
Check if plugin is enabled before consuming the event rate limit config

### LINKED ISSUE #4806: Cannot edit existing clusters due to unmarshal failure in v1.EventRateLimitConfigItem
### What happened
On dev, I cannot edit an existing cluster from the UI anymore because changing settings and hitting "Save Changes" produces an error:

> cannot decode patched cluster: json: cannot unmarshal string into Go struct field EventRateLimitConfig.spec.eventRateLimitConfig.namespace of type v1.EventRateLimitConfigItem

Note that I do not have EventRateLimit enabled or configured on this cluster. I suspect #4802 broke it, but just because it's a recent change to EventRateLimit configuration (haven't looked further into this).

### Expected behavior

Editing a cluster works.

### How to reproduce

- Create a cluster via the UI
- Wait for it to start up
- Attempt to edit it via the "Edit Cluster" menu

### Environment
- UI Version: master
- API Version: master
- Domain:
- Others:

### Current workaround

Edit the cluster via `kubectl edit`.

### Affected user persona

Cluster owners / administrators

ahmedwaleedmalik:

With https://github.com/kubermatic/dashboard/pull/4442 we changed the behavior of how form changes are synced with the DTOs. As a side effect, in the edit cluster view we were now setting the `spec.eventRateLimitConfig` unconditionally which was incorrect.

This PR ensures that we only propagate those values when the plugin `EventRateLimit` is enabled.

Fixes #4806

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -134,7 +134,7 @@ export class EditClusterComponent implements OnInit, OnDestroy {
       [Controls.KubernetesDashboardEnabled]: new FormControl(!!this.cluster.spec.kubernetesDashboard?.enabled),
       [Controls.AdmissionPlugins]: new FormControl(this.cluster.spec.admissionPlugins),
       [Controls.PodNodeSelectorAdmissionPluginConfig]: new FormControl(''),
-      [Controls.EventRateLimitConfig]: new FormControl(''),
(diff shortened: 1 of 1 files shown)


Waseem826 (approved):
/approve

PR #4822, opened 2022-08-15
prepare for kubevirt kubeconfig fixes

xrstf:
This PR ensures that the dashboard can deal with an unenoded, cleartext kubeconfig, which is what we want to keep in Vault in the future. The same applies for the GCP ServiceAccount.

/kind cleanup

```release-note
NONE
```


Waseem826 (approved):
/approve

Waseem826:
The failing tests are optional so this PR just needs milestone.

ahmedwaleedmalik:
@Waseem826 if it's okay with you, can you please create a ticket for fixing these failing CI jobs? I get it that they are optional but they are valuable tests.

PR #4835, opened 2022-08-15
Issues with resource quotas

### LINKED ISSUE #4740: Issues with Resource Quotas

1. Alignment issues (*isolated 😁)

Screenshot 2022-07-30 at 9 55 01 AM

2. We should use a typeahead here and project names should be sorted. It was very difficult for me to find my project.

Screenshot 2022-07-30 at 9 45 57 AM (2)

3. We shouldn't send in patch requests when nothing changed. Check if the form/controls changed, `this.control(XYZ).pristine`. 

https://user-images.githubusercontent.com/18264334/181875746-a9ee16ca-e831-4e43-a604-2784a6643248.mov

4. There is something extremely wrong with these input fields. It just keeps flushing the input if it exceeds 3 characters in case of `Memory` and `Disk`. For CPU if I enter `e` then it just removes the whole input, which is wrong. In general either don't handle `e` or ignore it.

https://user-images.githubusercontent.com/18264334/181875853-8516801c-c9a5-4ff2-a6f0-58baeb7e3dc6.mov

5. Maybe I'm not quite understanding this but the quota limitations didn't seem to work for me. This has nothing to do with the front-end IMO. I set a resource quota for 1 CPU and chose a flavor on AWS that has 2 vCPUs. It didn't restrict me in any way and I got a working cluster.

Screenshot 2022-07-30 at 10 05 18 AM
Screenshot 2022-07-30 at 10 05 35 AM
Screenshot 2022-07-30 at 10 08 21 AM

6. We should show units here in the widget.

Screenshot 2022-08-02 at 5 18 36 PM

7. Why are we polling for quotas so so often? This is from the cluster details view.

Screenshot 2022-08-02 at 10 07 09 PM
Screenshot 2022-08-02 at 10 05 20 PM

Talha-Jamil-TJ:

### Point 1 (b) & 3

While in `edit mode`:
- Project name is moved to the top
- submit button will stay disabled when quotas are unchanged
image

### Point 2

Select component is replaced with a typeahead 

image

### Point 4

Issues with number inputs are fixed

### Point 6

Units added to quota widget on hover

image

### Point 8

Fixed background colour of yellow progress bar

image
image

Fixes #4740

**What type of PR is this?**
/kind bug
/kind design

**Documentation**:

```documentation
NONE
```

```release-note
NONE
```

### DIFF
--- a/src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
+++ b/src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
@@ -18,10 +18,10 @@
 //
 // END OF TERMS AND CONDITIONS
 
-import {FormGroup, FormBuilder, Validators} from '@angular/forms';
+import {FormGroup, FormBuilder, Validators, FormControl} from '@angular/forms';
 import {Component, OnInit, OnDestroy, Inject, ChangeDetectorRef} from '@angular/core';
 import {MatDialogRef, MAT_DIALOG_DATA} from '@angular/material/dialog';
-import {takeUntil, filter, tap, distinctUntilChanged} from 'rxjs/operators';
+import {takeUntil, filter, tap, distinctUntilChanged, map} from 'rxjs/operators';
 import {Observable, Subject, of} from 'rxjs';
 import {NotificationService} from '@core/services/notification';
 import {ProjectService} from '@core/services/project';
@@ -30,10 +30,13 @@ import {QuotaVariables, QuotaDetails, Quota} from '@shared/entity/quota';
 import {KmValidators} from '@shared/validators/validators';
 import {ControlsOf} from '@shared/model/shared';
 import {Project} from '@shared/entity/project';
+import _ from 'lodash';
+import {ComboboxControls} from '@shared/components/combobox/component';
 
 enum Error {
   Required = 'required',
   AtLeastOneRequired = 'atLeastOneRequired',
+  IncorrectProject = 'incorrectProject',
 }
 
 @Component({
@@ -52,6 +55,9 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
   projects: Project[] = [];
   selectedProject: Project;
   selectedQuota: QuotaDetails;
+  projectControl: FormControl<Record<ComboboxControls.Select, string>>;
+  projectNameCountMap: Record<string, number>;
+  projectIdAndNameMap: Record<string, string>;
 
   constructor(
     private readonly _dialogRef: MatDialogRef<ProjectQuotaDialogComponent>,
@@ -67,6 +73,14 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
     return this.form?.controls.quota;
   }
 
+  get isQuotaUpdated(): boolean {
+    if (!this.editQuota) {
+      return this.form.controls.quota.dirty;
+    }
+
+    return !_.isEqual(this.editQuota.quota, this.quotaGroup.value);
+  }
+
   ngOnInit(): void {
     this._setSelectedQuota();
     this._initForm();
@@ -79,12 +93,14 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
     this._unsubscribe.complete();
   }
 
-  resetSubjectNameControl(): void {
-    this.form.controls.subjectName.reset();
+  projectDisplayFn(projectId: string): string {
+    return this.projectIdAndNameMap?.[projectId] ?? '';
   }
 
   getObservable(): Observable<Record<string, never>> {
-    if (this.form.invalid) return of(null);
+    if (this.form.invalid || !this.isQuotaUpdated) {
+      return of(null);
+    }
 
     const formValue = this.form.value as Quota;
 
@@ -114,21 +130,35 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
   }
 
   private _getProjects(): void {
-    this._projectService.allProjects.pipe(takeUntil(this._unsubscribe)).subscribe(projects => {
-      this.projects = projects;
+    this._projectService.allProjects
+      .pipe(
+        filter(projects => !_.isEqual(projects, this.projects)),
+        takeUntil(this._unsubscribe)
+      )
+      .subscribe(projects => {
+        this.projects = projects;
+        this.projectNameCountMap = projects.reduce(
+          (prev, curr) => ({...prev, [curr.name]: (prev[curr.name] || 0) + 1}),
+          {}
+        );
 
-      if (this.editQuota) {
-        this.selectedProject = projects.find(({id}) => id === this.editQuota.subjectName);
-      } else {
-        this.form.controls.subjectName.enable();
-      }
-    });
+        this.projectIdAndNameMap = projects.reduce((prev, curr) => ({...prev, [curr.id]: curr.name}), {});
+
+        if (this.editQuota) {
+          this.selectedProject = projects.find(({id}) => id === this.editQuota.subjectName);
+        }
+      });
   }
 
   private _getQuotas(): void {
-    this._quotaService.quotas.pipe(takeUntil(this._unsubscribe)).subscribe(quotas => {
-      this._quotas = quotas;
-    });
+    this._quotaService.quotas
+      .pipe(
+        filter(quotas => !_.isEqual(quotas, this._quotas)),
+        takeUntil(this._unsubscribe)
+      )
+      .subscribe(quotas => {
+        this._quotas = quotas;
+      });
   }
 
   private _initForm(): void {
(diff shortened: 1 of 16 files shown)


src/app/cluster/details/cluster/edit-cluster/template.html
```diff
@@ -96,7 +96,8 @@
                        formControlName="podNodeSelectorAdmissionPluginConfig"
                        fxFlex="100"></km-label-form>
         <div *ngIf="isPluginEnabled(admissionPlugin.EventRateLimit)">
-          <km-wizard-cluster-event-rate-limit [formControl]="form.get(Controls.EventRateLimitConfig)" [eventRateLimitConfig]="eventRateLimitConfig"></km-wizard-cluster-event-rate-limit>
+          <km-wizard-cluster-event-rate-limit [formControl]="form.get(Controls.EventRateLimitConfig)"
```
Waseem826, line 99: This has already been fixed in https://github.com/kubermatic/dashboard/pull/4800 so you can revert this.

Waseem826:
With this updated Project field, can we now enter custom Project name? If not, then why can't we use `km-combobox` which already has search field for filtering? And it also supports custom markup for options.

src/app/dynamic/enterprise/quotas/project-quota-dialog/template.html
```diff
@@ -65,32 +69,45 @@
...
 
-            <div fxLayoutAlign="end"
-                 class="km-select-trigger-button-wrapper">
-              <button matSuffix
-                      mat-icon-button
-                      [disabled]="!!editQuota"
-                      (click)="resetSubjectNameControl()">
-                <i class="km-icon-mask km-icon-remove"></i>
-              </button>
+              <mat-card-subtitle class="km-no-margin"
```
Waseem826, line 95:
```suggestion
              <mat-card-subtitle class="km-no-margin km-no-padding"
```
So that it is aligned with project name.

src/app/shared/components/number-stepper/component.ts
```diff
@@ -116,14 +117,16 @@ export class NumberStepperComponent implements AfterViewInit, OnDestroy, Control
   }
 
   set value(val: number | string) {
-    const parsed = this.type === 'decimal' ? this._decimalPipe.transform(val, '1.0-4') : val?.toString();
-    if (!Number.isNaN(parsed) && parsed !== null && parsed !== undefined) {
+    const parsed =
+      this.type === 'decimal' ? this._decimalPipe.transform(val, '1.0-4')?.replace(/,/g, '') : val?.toString();
```
Waseem826, line 121: What was the case which required to add `replace(/,/g, '')`?

src/app/shared/components/number-stepper/template.html
```diff
@@ -21,6 +21,7 @@
          matInput
          type="number"
          autocomplete="off"
+         onkeydown="return event.keyCode !== 69"
```
Waseem826, line 24: We shouldn't be ignore `e` in this component since its a generic component but instead this should be handled in the parent component where we want to ignore `e`

src/app/cluster/details/cluster/edit-cluster/template.html
Talha-Jamil-TJ: This is happening when I run `npm run fix`, if I don't run it, pipeline will fail.

Waseem826: Pipeline won't fail because of this particular case.

src/app/dynamic/enterprise/quotas/quota-widget/template.html
```diff
@@ -69,13 +69,13 @@
         <div class="km-detail-row"
              *ngIf="quotaDetails.quota.memory">
           <span>Memory</span>
-          <span>{{quotaDetails.status.globalUsage.memory ?? 0}}/{{quotaDetails.quota.memory}}</span>
+          <span>{{quotaDetails.status.globalUsage.memory ?? 0}}/{{quotaDetails.quota.memory}} <strong>GB</strong></span>
```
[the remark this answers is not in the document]
Talha-Jamil-TJ: this part is still being updated, Once I have approval from @cschieder then we can discuss it further.

src/app/shared/components/number-stepper/component.ts
Talha-Jamil-TJ: once we enter upto four digits `ex: 1111`, the method return `1,111`, which not supported by number input and cannot be parsed too, that's why I'm removing the `,` so the value can be parsed.

src/app/shared/components/number-stepper/template.html
Talha-Jamil-TJ: when we enter `e` in the text input it clears the input, this issue is present everywhere this component is being used.
This was done after discussion with @ahmedwaleedmalik

src/app/shared/components/number-stepper/component.ts
Waseem826: But this is only for `decimal` type and we're not using that in Resource quotas. Also, `1.0-4` should not return value with `,`.
In the screenshot below I entered 9 digits without any issue (on dev)

screenshot-localhost_8000-2022 08 16-14_47_58

This is what Angular docs say about the `digitsInfo` parameter: https://angular.io/api/common/DecimalPipe#digitsinfo

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
```diff
@@ -116,12 +131,12 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
   private _getProjects(): void {
     this._projectService.allProjects.pipe(takeUntil(this._unsubscribe)).subscribe(projects => {
       this.projects = projects;
+      this.projectNameCountMap = projects.reduce(
+        (prev, curr) => ({...prev, [curr.id]: projects.filter(({name}) => name === curr.name).length}),
+        {}
+      );
```
Waseem826, line 137: Instead of running `filter` for every project (hence increasing the complexity), can we do the following?
```suggestion
      this.projectNameCountMap = {};
      projects.forEach(project => {
        this.projectNameCountMap[project.id] = (this.projectNameCountMap[project.id] || 0) + 1;
      });
```

src/app/cluster/details/cluster/edit-cluster/template.html
Waseem826, line 99: This has been fixed in `master` so please revert this change and rebase your branch on latest `master`.

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
```diff
@@ -52,6 +56,11 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
   projects: Project[] = [];
   selectedProject: Project;
   selectedQuota: QuotaDetails;
+  projectControl: FormControl<Record<ComboboxControls.Select, string>>;
+  // map of project id and count of its name duplicates
```
Waseem826, line 60: I think there's no need for this comment and the one below because property names are self explanatory.

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
```diff
@@ -79,30 +96,30 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
     this._unsubscribe.complete();
   }
 
-  resetSubjectNameControl(): void {
-    this.form.controls.subjectName.reset();
+  projectDisplayFn(projectId: string): string {
```
Waseem826, line 99: Why can't we use `projectIdAndNameMap?.[projectId]` directly in template instead of calling this method?

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
```diff
@@ -114,21 +131,33 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
...
+      .subscribe(projects => {
+        this.projects = projects;
+        this.projectNameCountMap = projects.reduce(
+          (prev, curr) => ({...prev, [curr.name]: (prev[curr.name] || 0) + 1}),
+          {}
+        );
+
+        this.projectIdAndNameMap = projects.reduce((prev, curr) => ({...prev, [curr.id]: curr.name}), {});
+
+        if (this.editQuota) this.selectedProject = projects.find(({id}) => id === this.editQuota.subjectName);
```
Waseem826, line 148: Please move this statement inside a block

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
```diff
@@ -149,16 +178,31 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
         validators: Validators.required,
         nonNullable: true,
       }),
-      subjectName: this._builder.control(
-        {value: subjectName, disabled: true},
-        {validators: Validators.required, nonNullable: true}
-      ),
+      subjectName: this._builder.control(subjectName ?? '', {
+        validators: Validators.required,
+        asyncValidators: AsyncValidators.ProjectSelection,
```
Waseem826, line 183: I think we don't need this validator anymore since we've now switched to `km-combobox`

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
```diff
@@ -149,16 +178,31 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
...
+        nonNullable: true,
+      }),
+    });
+
+    this.projectControl = this._builder.control(null, {
```
Waseem826, line 188: Why can't we use `subjectName` control instead of this?

Waseem826 (changes requested):

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
Talha-Jamil-TJ: The `km-combobox` component returns a value which is nested inside an object (`{ select: <actual data> }`), which is not representative of the payload I need to send to the API, this is why I'm handling it in a separate formControl.

src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
Talha-Jamil-TJ: we need a function to pass as input to km-combobox

Waseem826 (approved):
/approve

PR #4841, opened 2022-08-16
Loading legacy metering reports 

### LINKED ISSUE #4807: Not able to fetch old metering reports
### What happened?

I was trying to download reports created before 2.21 migration. It is not possible from the dashboard.

### Expected behavior

Available list of reports to download or remove.

Go to `https://<dashboard_url>/settings/metering`.

- KKP version: 2.21

### Additional information

Old reports are stored in the root directory of S3 bucket and we don't download them from there while traversing schedule configurations. Old reports are available to fetch under GET `/api/v1/admin/metering/reports`.

pkprzekwas:
As we introduced new metering mechanism we need to handle presentation of deprecated reports that users may still have after migrating to KKP 2.21.

Fixes #4807 

**What type of PR is this?**
/kind bug

```release-note
NONE
```


src/app/dynamic/enterprise/metering/old-reports/card/template.html
```diff
@@ -0,0 +1,38 @@
...
+-->
+
+<mat-card>
+  <mat-card-header (click)="toggleReports()">
+    <mat-card-title fxFlex
+                    fxLayout="row"
+                    fxLayoutAlign=" center">
+      <div fxFlex>
+        Old Reports <i class="km-icon-mask"
+           [ngClass]="showReports ? 'km-icon-mask km-icon-arrow-up' : 'km-icon-mask km-icon-arrow-down'"></i>
```
Waseem826, line 30:
```suggestion
           [ngClass]="showReports ? 'km-icon-arrow-up' : 'km-icon-arrow-down'"></i>
```

src/app/dynamic/enterprise/metering/old-reports/card/component.ts
```diff
@@ -0,0 +1,36 @@
+//                Kubermatic Enterprise Read-Only License
+//                       Version 1.0 ("KERO-1.0”)
+//                   Copyright © 2020 Kubermatic GmbH
```
Waseem826, line 3:
```suggestion
//                   Copyright © 2022 Kubermatic GmbH
```
Please also update the year in other newly added files.

src/app/dynamic/enterprise/metering/component.ts
```diff
@@ -68,6 +71,13 @@ export class MeteringComponent implements OnInit, OnDestroy {
           this._cdr.detectChanges();
         },
       });
+
+    this._meteringService.oldReports().subscribe(reports => {
```
Waseem826, line 75:
```suggestion
    this._meteringService.oldReports().pipe(takeUntil(this._unsubscribe)).subscribe(reports => {
```

src/app/dynamic/enterprise/metering/old-reports/list/component.ts
```diff
@@ -0,0 +1,132 @@
...
+  size = 'size',
+  lastModified = 'lastModified',
+  actions = 'actions',
+}
+
+@Component({
+  selector: 'km-old-metering-reports-list',
+  templateUrl: './template.html',
+})
+export class MeteringOldReportListComponent implements OnInit {
```
Waseem826, line 45:
```suggestion
export class MeteringOldReportListComponent implements OnInit, OnDestroy {
```

src/app/dynamic/enterprise/metering/old-reports/list/component.ts
```diff
@@ -0,0 +1,132 @@
...
+      },
+    };
+    this._matDialog
+      .open(ConfirmationDialogComponent, dialogConfig)
+      .afterClosed()
+      .pipe(filter(isConfirmed => isConfirmed))
+      .pipe(
+        switchMap(_ => {
+          this._reportsInProgress.add(reportName);
+          this.dataSource.data = this.reports.filter(report => report.name !== reportName);
```
Waseem826, line 110: Shouldn't these 2 lines be inside `subscribe` so that view is updated only when the request finishes?

src/app/dynamic/enterprise/metering/old-reports/list/component.ts
```diff
@@ -0,0 +1,132 @@
...
+        error: _ => this._reportsInProgress.delete(reportName),
+      });
+  }
+
+  isProcessingReport(report: Report): boolean {
+    return this._reportsInProgress.has(report.name);
+  }
+
+  isPaginatorVisible(): boolean {
+    return this.reports && this.reports.length > 0 && this.paginator && this.reports.length > this.paginator.pageSize;
```
Waseem826, line 130:
```suggestion
    return this.reports?.length > this.paginator?.pageSize;
```

src/app/dynamic/enterprise/metering/old-reports/list/template.html
```diff
@@ -0,0 +1,100 @@
...
+        mat-sort-header>Size</th>
+    <td mat-cell
+        *matCellDef="let element">{{element.size | kmSize}}</td>
+  </ng-container>
+
+  <ng-container [matColumnDef]="column.lastModified">
+    <th mat-header-cell
+        *matHeaderCellDef
+        class="km-header-cell p-20"
+        mat-sort-header>Created</th>
```
Waseem826, line 51: Should we label this column `Updated`/`Last Updated` since we're displaying the last modified time?

src/app/dynamic/enterprise/metering/template.html
```diff
@@ -26,6 +26,18 @@
 </ng-container>
 
 <ng-container *ngIf="isLoading()">
```
Waseem826, line 28:
```suggestion
<ng-container *ngIf="isLoading() || !loadOldReportsCard">
```

And then we can remove the spinner below. Wdyt?

Waseem826 (changes requested):
Some feedback. PTAL

src/app/dynamic/enterprise/metering/component.ts
pkprzekwas: This only do one HTTP request under the hood, it does not merge with any timer. But I guess you're right and we still should apply takeUntil.

src/app/dynamic/enterprise/metering/old-reports/list/template.html
pkprzekwas: Hmm, technically if someone touches a file on S3 this value may change, but in 99% of cases this `Last modified` is same as `Created`. It's just that S3 doesn't have created value, whenever you create sth new it sets updated (Last Modified) value with creation timestamp. I think @cschieder has decided in the past that displaying this is more user friendly.

src/app/dynamic/enterprise/metering/template.html
pkprzekwas: Makes sense, I like this idea 👍

src/app/dynamic/enterprise/metering/old-reports/list/template.html
cschieder: I'd be fine with using "Last modified" too, however there's the argument that if it's 99% of the time the creation time then "Created" could be easier to understand, especially in the context of reports.

src/app/dynamic/enterprise/metering/template.html
pkprzekwas: I had to tweak it a bit cos there is a possibility that there are no old reports in which case we do not load the component.

pkprzekwas:
Adding some screenshots for @cschieder 
Screenshot 2022-08-16 at 15 32 39
Screenshot 2022-08-16 at 15 32 45

cschieder:
@pkprzekwas 
Please adjust it to match the event card in the following way
image

- Move the unfold arrow to the left
- Make the spacing below and above the "Old Reports" text match
- Add a counter of how many reports there are on the right 

Additionally, 
should we call it Legacy Reports or add an info icon and a tooltip? Open for suggestions and opinions

> Old Reports were created before Version 2.21 ...

ahmedwaleedmalik:
@pkprzekwas Please remove the ticket number from the PR title.

pkprzekwas:
> @pkprzekwas Please remove the ticket number from the PR title.

Damn, happened again... sorry.

pkprzekwas:
> Member

I love the idea of renaming old to legacy. I had some kind of brain damage and was struggling for quite some time for an appropriate naming.

pkprzekwas:
Updated views for @cschieder (I wasn't sure which icon should be used)
Screenshot 2022-08-17 at 11 31 27
Screenshot 2022-08-17 at 11 31 36

pkprzekwas:
@cschieder please
Screenshot 2022-08-17 at 11 44 25
Screenshot 2022-08-17 at 11 44 32
 check updated views

src/app/dynamic/enterprise/metering/legacy-reports/list/template.html
```diff
@@ -0,0 +1,100 @@
...
+   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+END OF TERMS AND CONDITIONS
+-->
+
+<table class="km-table"
+       mat-table
+       matSort
+       matSortActive="name"
+       matSortDirection="asc"
```
Waseem826, line 27: We should not set these attributes here since we're already setting them in `.ts` file.

src/app/dynamic/enterprise/metering/legacy-reports/list/template.html
```diff
@@ -0,0 +1,100 @@
...
+                kmThrottleClick
+                (throttleClick)="download(element.name)">
+          <i class="km-icon-mask km-icon-download"></i>
+        </button>
+
+        <button mat-icon-button
+                kmThrottleClick
+                (throttleClick)="remove(element.name)">
+          <i class="km-icon-mask km-icon-delete"></i>
+        </button>
```
Waseem826, line 84: I think we should hide action buttons when spinner is displayed.

```suggestion
        <ng-container *ngIf="isProcessingReport(element); else actionButtons">
          <mat-spinner color="accent"
                       class="km-spinner"
                       fxFlexAlign=" center"
                       [diameter]="25"></mat-spinner>
        </ng-container>
        
        <ng-template #actionButtons>
          <button id="km-download-report-button"
                  mat-icon-button
                  kmThrottleClick
                  (throttleClick)="download(element.name)">
            <i class="km-icon-mask km-icon-download"></i>
          </button>

          <button mat-icon-button
                  kmThrottleClick
                  (throttleClick)="remove(element.name)">
            <i class="km-icon-mask km-icon-delete"></i>
          </button>
        </ng-template>
```

Waseem826 (changes requested):
Couple of comments. PTAL

Waseem826 (approved):
/approve

PR #4844, opened 2022-08-16
Rename rockylinux in the correct format

moadqassem:
It seems that rockylinux was renamed into rockyLinux which has messed up couple of properties in the api and in the OSSpec. For example matching the name with operating system name doesn't work, which leads to failures when trying to pick a template for rockylinux in vsphere for example. Same thing will happen to KubeVirt, Openstack and Bare metal provisioning in the future. This PR reverts back rockylinux to it's original name. 

Fixes #

**What type of PR is this?**
/kind bug

```release-note
None
```

**Documentation**:

```documentation
None
```


ahmedwaleedmalik (approved):
/approve ⚡

Waseem826 (approved):
/approve

PR #4854, opened 2022-08-18
Add legacy report prefix to not show raw data csv files

### LINKED ISSUE #4807: Not able to fetch old metering reports

WeirdMachine:

Fixes a bug where raw data was mistakenly shown as reports.

The scheduleName effectively acts as a prefix filter, therefore we can use it and set the prefix for legacy reports.

Fixes #4807

/kind bug

```release-note
NONE
```

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #4856, opened 2022-08-18
Update openstack configuration in edit provider dialog

### LINKED ISSUE #4702: Editing Openstack provider settings is ignoring application credentials
### What happened

When using Openstack application credentials, the "Edit Provider" dialog from cluster details is not correct. It only provides options for username and password, which are unused fields when using Openstack application credentials. This is what it looks like currently:

Screenshot 2022-07-21 at 09 12 27

When creating a cluster, application credentials are offered as option:

Screenshot 2022-07-21 at 09 16 22

Screenshot 2022-07-21 at 09 13 50

After creating a cluster with application credentials, I tried to rototate/update credentials, for example because they are going to expire soon. There's several issues with that:

- I cannot update application credentials as only user credentials (username / password) are offered to be updated.
- It's possible on the backend to switch from app credentials to user credentials (and there's not really a reason to block that), but app credentials don't have a project. That means when switching from app credentials to user credentials, a third entry needs to be offered for the relevant project name or project ID (just like the setup dialog does).
- The current dialog, when used on an app credentials cluster, will render the cluster unusable, because no project will be set (also see https://github.com/kubermatic/kubermatic/issues/8809).

I think it's possible that we need some kind of API endpoint that returns what kind of credentials are used for the cluster (app creds or user creds) so the dashboard can understand what credentials were previously configured. Please open a kubermatic/kubermatic issue with what you need, as I'm not sure about that.

### Expected behavior

The dashboard knows what kind of credentials were used or at least offers the same options as the setup dialog, specifically resolving the bug that going from app credentials to user credentials via the current dialog will not ask for project information. Note that when updating existing user credentials, the project reference shouldn't be editable, since we don't support switching between projects and existing resources will be orphaned in the old project.

### How to reproduce

Create an Openstack cluster and try to rotate/update credentials.

### Environment
- UI Version: master
- API Version: master
- Domain:
- Others:

### Current workaround

Edit cluster credentials in the backend via `kubectl`. It also should work via the KKP API directly since the restriction on what fields can be updated is in the dashboard.

### Affected user persona

KKP users who own Openstack clusters

Waseem826:
Update openstack configuration in edit provider dialog.

screenshot-localhost_8000-2022 08 19-01_38_18

screenshot-localhost_8000-2022 08 19-01_38_39

Since we don't have an API endpoint to determine current credentials type so I've added a warning message when project is updated.

screenshot-localhost_8000-2022 08 19-14_43_04

Fixes #4702 

**What type of PR is this?**

/kind bug

```release-note
Update openstack configuration in edit provider dialog
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/src/app/cluster/details/cluster/edit-provider-settings/openstack-provider-settings/component.ts
+++ b/src/app/cluster/details/cluster/edit-provider-settings/openstack-provider-settings/component.ts
@@ -13,38 +13,81 @@
 // limitations under the License.
 
 import {Component, OnDestroy, OnInit} from '@angular/core';
-import {FormBuilder, FormGroup} from '@angular/forms';
+import {FormBuilder, FormGroup, Validators} from '@angular/forms';
+import {CredentialsType} from '@app/wizard/step/provider-settings/provider/extended/openstack/service';
 import {ClusterService} from '@core/services/cluster';
-import {ProviderSettingsPatch} from '@shared/entity/cluster';
-import {merge, Subject} from 'rxjs';
+import {OpenstackCloudSpecPatch, ProviderSettingsPatch} from '@shared/entity/cluster';
+import {Subject} from 'rxjs';
 import {distinctUntilChanged, takeUntil} from 'rxjs/operators';
 
 enum Control {
+  Default = 'default',
+  Application = 'application',
+  Domain = 'domain',
   Username = 'username',
   Password = 'password',
+  Project = 'project',
+  ProjectID = 'projectID',
+  ApplicationCredentialID = 'applicationCredentialID',
+  ApplicationCredentialSecret = 'applicationCredentialSecret',
 }
 
 @Component({
   selector: 'km-openstack-provider-settings',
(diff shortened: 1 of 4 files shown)


src/app/cluster/details/cluster/edit-provider-settings/openstack-provider-settings/component.ts
```diff
@@ -13,52 +13,134 @@
...
+    this.form.controls[Control.Default]
+      .get(Control.Project)
+      .valueChanges.pipe(distinctUntilChanged())
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(value => {
+        const projectIDControl = this.defaultCredentialsForm.get(Control.ProjectID);
+        value ? projectIDControl.disable() : projectIDControl.enable();
+      });
+
+    this.form.controls[Control.Default]
```
KhizerRehan, line 81: duplicate code as above

embik (approved):

PR #4869, opened 2022-08-19
Fix issue to display empty placeholder in case no nodes exists

### LINKED ISSUE #4744: Display empty placeholder when no nodes exist
Currently, when no worker nodes exist instead of empty placeholder "No nodes available." it currently shows loading icon until any node exist to show table.

- [ ] Validate cluster health from cluster instead of Input from parent
- [ ] Update conditional checks to show loading icon

### How to reproduce
- Navigate to External cluster except "Custom" as UI doesn't display MD list in case of custom
- Navigate to MD details page
- When no nodes exist it was showing "Loading icon" so it should display "no nodes found"

KhizerRehan:
This PR fixes to display "no nodes available" when nodes list is empty but it was showing `loading` icon.

**Before:**
(Cluster was running && Nodes were empty)

before

**After:**

**Case-1:**
(Cluster was running && Nodes were empty)

after

**Case-2:**
(Cluster was running && Nodes list):

https://user-images.githubusercontent.com/17727069/185710814-21bf6813-b9e5-4c28-87e4-ab2668d84cf6.mp4

Fixes #4744 

**What type of PR is this?**
/kind bug

Issue was `isClusterRunning` was not passed in case of **MD details** page. Although cluster was in running state. Instead of passing cluster **state** from parent in details page NOW state is checked inside nodes list as we are already passing external cluster in both usages. which fixes the issue in following pages.

- Details page
- MD Details page

```release-note
Shows placeholder when no nodes exists inside external machine deployment details page
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/src/app/cluster/details/external-cluster/external-node-list/component.ts
+++ b/src/app/cluster/details/external-cluster/external-node-list/component.ts
@@ -17,7 +17,7 @@ import {MatPaginator} from '@angular/material/paginator';
 import {MatSort} from '@angular/material/sort';
 import {MatTableDataSource} from '@angular/material/table';
 import {UserService} from '@core/services/user';
-import {ExternalCluster} from '@shared/entity/external-cluster';
+import {ExternalCluster, ExternalClusterState} from '@shared/entity/external-cluster';
 import {NodeMetrics} from '@shared/entity/metrics';
 import {Node, NodeIPAddress} from '@shared/entity/node';
 import {NodeUtils} from '@shared/utils/node';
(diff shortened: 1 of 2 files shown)


src/app/cluster/details/external-cluster/external-node-list/component.ts
```diff
@@ -72,11 +71,15 @@ export class ExternalNodeListComponent implements OnInit, OnChanges, OnDestroy {
...
-    return (_.isEmpty(this.nodes) && !this.isClusterRunning) || !this.isInitialized;
+    return (_.isEmpty(this.nodes) && !this.isRunning()) || !this.isInitialized;
   }
 
   hasNoData(): boolean {
-    return _.isEmpty(this.nodes) && this.isClusterRunning && this.isInitialized;
+    return _.isEmpty(this.nodes) && this.isRunning() && this.isInitialized;
+  }
+
+  isRunning(): boolean {
```
Waseem826, line 81: This is the same method which was bound with `isClusterRunning` Input so how does this make the difference?

KhizerRehan: Correct! At list details page same logic is used to check state from cluster and pass `isClusterRunning` passing as input where in case of MD details page -> `isClusterRunning` is never passed -> which makes !isClusterRunning -> truthy values and satisfies loader condition therefore it just keep displaying loading state in MD details page where nodes are shown

Waseem826 (approved):
/approve

PR #4897, opened 2022-08-25
Replace quota update PATCH with PUT

pkprzekwas:
From the REST API perspective this should be PUT hence we decided to replace PATCH with PUT in the backend. At the moment PATCH and PUT points to the same logic under the hood so it should be safe to update the Dashboard.

**What type of PR is this?**
/kind bug

Depends on https://github.com/kubermatic/kubermatic/pull/10857 

```release-note
NONE
```


Waseem826 (approved):
/approve

PR #4904, opened 2022-08-29
Display OIDC groups in user settings

### LINKED ISSUE #4711: Present list of groups decoded from bearer token
As a KKP user, on the `/account` page, I would like to see a list of groups that are embedded in my OIDC identity token.

After decoding bearer token contains following payload (you can use jwt.io for decoding):
```json
{
  "iss": "https://dev.kubermatic.io/dex",
  "sub": "xxx",
  "aud": "kubermatic",
  "exp": 1657965131,
  "iat": 1657878731,
  "nonce": "xxx",
  "at_hash": "xxx",
  "email": "pkprzekwas@gmail.com",
  "email_verified": true,
  "groups": [
    "kubermatic:sig-api",
    "kubermatic:development",
    "kubermatic:internal"
  ],
  "name": "pprzekwas",
  "preferred_username": "pkprzekwas"
}
```
We need to get the groups list and present it on `/account` page. Take into account that some tokens doesn't contain a list of groups.

pkprzekwas:
As for today the only way to check OIDC groups assigned to user is to decode JWT token. This PR adds simple list of decoded OIDC groups to user settings view.
Screenshot 2022-08-30 at 09 06 46

Fixes #4711 

**What type of PR is this?**
/kind feature

```release-note
NONE
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/blob/master/content/kubermatic/master/architecture/role-based-access-control/groups-support/_index.en.md
```

### DIFF
--- a/src/app/settings/user/template.html
+++ b/src/app/settings/user/template.html
@@ -38,6 +38,21 @@ limitations under the License.
            fxLayoutGap="10px">
         <div class="user-name">{{user.name}}</div>
         <div class="user-email">{{user.email}}</div>
+        <km-property *ngIf="user.groups; else noGroups">
+          <div key
+               class="user-groups-key">OIDC Groups</div>
+          <div value>
+            <km-labels [labels]="user.groups"></km-labels>
+          </div>
+        </km-property>
+        <ng-template #noGroups>
+          <div>
+            No OIDC groups assigned.
+            <a href="https://docs.kubermatic.com/kubermatic/master/architecture/role-based-access-control/groups-support/"
+               target="_blank"
+               rel="noreferrer noopener">Learn more ></a>
+          </div>
+        </ng-template>
       </div>
     </div>
 
--- a/src/app/settings/user/style.scss
+++ b/src/app/settings/user/style.scss
@@ -32,6 +32,10 @@
     font-size: variables.$font-size-subhead;
   }
 
+  .user-groups-key {
+    margin-bottom: 8px;
+  }
+
   .account-settings {
     font-size: variables.$font-size-subhead;
     min-width: 390px;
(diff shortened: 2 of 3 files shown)


src/app/member/edit-member/component.ts
```diff
@@ -49,6 +49,7 @@ export class EditMemberComponent implements OnInit {
         id: this.member.id,
         name: this.member.name,
         email: this.member.email,
+        groups: this.member.groups,
```
KhizerRehan, line 52: this field is always `undefined` irrespective you have groups or no groups in bearer token.

what is the use case to add here in EditMemberDialog?

src/app/settings/user/template.html
```diff
@@ -38,6 +38,14 @@
            fxLayoutGap="10px">
         <div class="user-name">{{user.name}}</div>
         <div class="user-email">{{user.email}}</div>
+        <km-property *ngIf="user.groups">
```
KhizerRehan, line 41: Inorder to show empty placeholder we need to remove this `*ngIf`

Expected output in case no groups exists.

display empty placeholder

KhizerRehan (changes requested):
Feedback please PTAL

src/app/settings/user/template.html
pkprzekwas: Actually, I'm hesitating if we should show the OIDC groups if not configured? @cschieder WDYT?

src/app/member/edit-member/component.ts
pkprzekwas: You're right. Let me remove it.

pkprzekwas:
> Feedback please PTAL

I've addressed your concerns. Please take a look again.

PR #4962, opened 2022-09-12
Refactor resource quota polling logic

### LINKED ISSUE #4902: Refactor resource quota polling logic
#### Current behaviour
Resource quotas are polled every time we navigate to a page where project quota widget is visible, therefore there is a delay until the widget is visible
#### Required refactoring
The resource quota's data should be saved once retrieved from the API, so it would show instantly once loaded.
Once the resource quota widget is visible, we should still keep updating them from API.

Talha-Jamil-TJ:

A delay was observed in the resource quota widget while navigating between different routes, this PR fixes it.

Fixes #4902 

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/src/app/dynamic/enterprise/quotas/service.ts
+++ b/src/app/dynamic/enterprise/quotas/service.ts
@@ -23,16 +23,14 @@ import {Injectable} from '@angular/core';
 import {environment} from '@environments/environment';
 import {Quota, QuotaDetails, QuotaVariables} from '@shared/entity/quota';
 import {merge, Observable, of, Subject, timer} from 'rxjs';
-import {catchError, map, shareReplay, switchMap} from 'rxjs/operators';
+import {catchError, map, shareReplay, switchMap, startWith, tap} from 'rxjs/operators';
 import {AppConfigService} from '@app/config.service';
 
 @Injectable({
   providedIn: 'root',
 })
 export class QuotaService {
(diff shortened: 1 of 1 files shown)


src/app/dynamic/enterprise/quotas/service.ts
```diff
@@ -70,20 +75,25 @@ export class QuotaService {
...
+            this._previousQuotaMap.set(projectId, quota);
+          }
+        })
       );
 
       this.quotaMap.set(projectId, quota$);
     }
 
-    return this.quotaMap.get(projectId);
+    return this.quotaMap.get(projectId).pipe(startWith(this._previousQuotaMap.get(projectId) ?? null));
```
Waseem826, line 96: Are we handling this `null` value where we call this method?

Talha-Jamil-TJ: Yes, I've put in `filter( Boolean)` where ever I'm. subscribing to quotas
image

Talha-Jamil-TJ: Yes, isLoading is being set to false instantly, I'm catering for this.

Waseem826 (approved):
/approve

PR #5041, opened 2022-09-28
Show all project switcher - Add Spinner and disable toggle while pending

### LINKED ISSUE #4531: Show all project switcher - Add Spinner and disable toggle while pending

**Issue-1**
Fetching all projects took more time as it was an API call which fetches all records and since it was less than 20 projects which took more ~3s and the time while projects were fetching User is clueless is some action is happening or not and this loading time will directly proportional to **No of project** .  It will be great to add **spinner** while some async stuff is going on at the Background to notify user something is going on and be **PATIENT**.

**Issue-2**
- User can be evil and it can trigger multiple API calls based on FAST switching of toggle

- On/Off -> On/Off -> On/Off -> it may be bad from UI point + Backend point that so many request are sent to backend IF (it is not handled from FE) side to cancel previous request but Still we should **disable** until the request response has come back

https://user-images.githubusercontent.com/17727069/171803049-5756790f-1a74-4638-a5ee-154792883d87.mp4

- Add spinner
- Disable toggle  or some better suggestions by chiara

Talha-Jamil-TJ:

Fixes #4531

**What type of PR is this?**
/kind feature

```release-note
[Admin] Clicking Show All Projects toggle:
The loading spinner is shown when all projects are fetched
```

**Documentation**:

```documentation
NONE
```


src/app/project/component.ts
```diff
@@ -171,6 +172,7 @@ export class ProjectComponent implements OnInit, OnChanges, OnDestroy {
       .pipe(takeUntil(this._unsubscribe))
       .pipe(switchMap(() => this._userService.patchCurrentUserSettings(objectDiff(this.settings, this._apiSettings))))
       .subscribe(settings => {
+        this.isProjectsLoading = false;
```
Waseem826, line 175: This should be moved inside subscription of `this._projectService.projects`.

src/app/project/template.html
```diff
@@ -22,7 +22,9 @@
       <div fxFlex></div>
       <mat-slide-toggle *ngIf="isAdmin"
                         class="all-projects-toggle"
+                        [disabled]="isProjectsLoading"
                         labelPosition="before"
+                        [matTooltip]="isProjectsLoading ? 'Please wait while projects are loading' : null"
```
KhizerRehan, line 27: There is something wrong with logic in parent! it appears occasionally.

this section appears on ` *ngIf="!isInitializing ` but it should we need to show when everytime API call is triggered `isProjectsLoading`

https://user-images.githubusercontent.com/17727069/192952980-9a7af28f-3a77-4748-abcd-cb4ee66ac348.mp4

KhizerRehan (changes requested):
Feedback

src/app/project/template.html
KhizerRehan: You can ignore, this comment i think the issue was when i was trying to click on the "Show all projects" and cursor was still on that toggle so **tooltip** was not appearing but if you move your cursor from toggle and than hover again it is showing tooltip

KhizerRehan: cc: @Waseem826

Waseem826 (approved):
/approve

Waseem826:
/hold
Please unhold after all CI jobs have passed.

PR #5055, opened 2022-09-30
External cluster wizard: fixed EKS region changes subscription

Talha-Jamil-TJ:

Fixed issue in external cluster EKS credential step where not selecting `provider preset`, entering `Access Key ID` and `Secret Access Key` and then changing regions was causing no `vpcs` API request.

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/src/app/core/services/external-cluster.ts
+++ b/src/app/core/services/external-cluster.ts
@@ -39,7 +39,7 @@ import {GCPDiskType, GCPMachineSize} from '@app/shared/entity/provider/gcp';
 export class ExternalClusterService {
   providerChanges = new BehaviorSubject<ExternalClusterProvider>(undefined);
   presetChanges = new Subject<string>();
-  regionChanges = new Subject<string>();
+  private _regionChanges = new BehaviorSubject<string>(null);
   presetStatusChanges = new Subject<boolean>();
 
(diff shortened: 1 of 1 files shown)


Waseem826 (approved):
/approve

src/app/core/services/external-cluster.ts
```diff
@@ -86,9 +86,13 @@ export class ExternalClusterService {
     this.presetChanges.next(preset);
   }
 
+  get regionChanges() {
+    return this._regionChanges.asObservable().pipe(filter(Boolean));
```
Waseem826, line 90: What if the region is cleared? I think this event will not be emitted because of `Boolean` filter. Wdyt?

Talha-Jamil-TJ: Once selected, I don't see a way to remove `regions`
image

Waseem826: The UI can change in future. I think we should remove this `filter(Boolean)` operator and handle it where we’re using `regionChanges`. That way we can also do cleanups in those places if region is cleared.

Waseem826 (approved):
/unhold

PR #5060, opened 2022-10-03
External cluster wizard: fixed AKS preset change bug

Talha-Jamil-TJ:

Fixed bug in external cluster AKS wizard, where changing the credentials preset was not updating `Kubernetes version`, `locations` and `resource groups` API calls.

- No Ticket exists

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/src/app/external-cluster-wizard/steps/external-cluster/provider/aks/component.ts
+++ b/src/app/external-cluster-wizard/steps/external-cluster/provider/aks/component.ts
@@ -26,7 +26,7 @@ import {ExternalClusterService} from '@core/services/external-cluster';
 import {NameGeneratorService} from '@core/services/name-generator';
 import {ErrorType} from '@shared/types/error-type';
 import {Observable, of} from 'rxjs';
-import {debounceTime, switchMap, takeUntil, tap} from 'rxjs/operators';
+import {debounceTime, switchMap, takeUntil, tap, filter} from 'rxjs/operators';
 import {
   ExternalCloudSpec,
   ExternalCluster,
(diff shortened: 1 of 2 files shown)


src/app/external-cluster-wizard/steps/external-cluster/provider/aks/component.ts
```diff
@@ -306,9 +306,11 @@ export class AKSClusterSettingsComponent
           this.vmSizeLabel = this.vmSizes?.length ? VMSizeState.Ready : VMSizeState.Empty;
         });
 
-      this._getAKSKubernetesVersions();
-      this._getAKSLocations();
-      this._getAKSResourceGroups();
+      this._externalClusterService.presetChanges.pipe(startWith(null), takeUntil(this._unsubscribe)).subscribe(_ => {
```
KhizerRehan, line 309: There are couple of scenarios:

Case-1: 
- Component gets destroyed when present is cleared from 2nd step via (x) icon -> Reinitialising make it work fine as new component gets initialised

Case-2:
- Component gets initialised on preset -> Navigate to 3rd step -> Navigate back 2nd step -> don't clear preset -> change present -> 3rd component is not destroyed BUT here the logic you have put place is **correct** BUT we will never get `null` until components gets destroyed EVEN if we get it it will be use less because it will be Case-1:

- check to run if value not null then trigger API calls  OR 
- Invoke API calls if preset value is coming in `_` parameter

Waseem826: @Talha-Jamil-TJ why not use the same `BehaviorSubjet` approach that you used for region changes?

Talha-Jamil-TJ: @Waseem826 you are right, I'll change `presetChanges` to a behavior subject too.

Waseem826 (approved):
/approve

Waseem826 (approved):

PR #5163, opened 2022-10-28
Cleanup some unused code in providers

lsviben:

Cleans up some of the unused code that was migrated by accident from `pgk/providers`.

There is still more left, for now I removed the easy pickings.

**What type of PR is this?**
/kind cleanup

A thing to note is that the code for `CreateOrUpdateCredentialSecretForCluster` was removed from the pkg/provider/kubernetes/credentials.go, and instead we take it from the KKP dependency, as it is something used as well in KKP core and I dont think it needs to be duplicated in the API code.

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


pkprzekwas (approved):
lgtm

PR #5167, opened 2022-10-29
Use the operator's native support for a custom dashboard/api image

xrstf:
Now that https://github.com/kubermatic/kubermatic/pull/11229 is merged we can get rid of the hack to inject our Docker image for the API and just rely on the operator for us. :tada: 

I ran one test version first where I purposefully did not push the dashboard image into kind, in order to verify that the operator is indeed picking up our injected version. As expected, the prowjob failed: https://public-prow.loodse.com/view/gs/prow-dev-public-data/pr-logs/pull/kubermatic_dashboard/5167/pre-dashboard-api-e2e/1586437233034924032

I also removed the /scratch hack, which I introduced a long time ago but seemingly more modern kind versions are configured slightly differently. Since we do not use this tmpfs in kubermatic anymore, I figure it's not needed in today's CI cluster and simply removed it. Otherwise the local-path-provisioner crashloops with

> time="2022-10-29T19:32:16Z" level=fatal msg="Error starting daemon: invalid empty flag helper-pod-file and it also does not exist at ConfigMap local-path-storage/local-path-config with err: helperPod.yaml is not exist in local-path-config ConfigMap"

And this then prevents Minio from coming up, as the PVC stays Pending. The full e2e tests still fail, but this is now related to the way they depend on the KKP repo's scripting. Maybe a future PR could refactor them to setup KKP in the same way as the dashboard-api-e2e tests do it (i.e. use the installler from the docker image).

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/.prow/tests-e2e.yaml
+++ b/.prow/tests-e2e.yaml
@@ -41,7 +41,6 @@ presubmits:
       preset-kind-volume-mounts: 'true'
       preset-kubeconfig-ci: 'true'
       preset-minio: 'true'
-      preset-scratch-tmpfs: 'true'
     spec:
       containers:
         - image: quay.io/kubermatic/chrome-headless:v1.6.0
@@ -97,7 +96,6 @@ presubmits:
       preset-kind-volume-mounts: 'true'
       preset-kubeconfig-ci: 'true'
       preset-minio: 'true'
-      preset-scratch-tmpfs: 'true'
     spec:
       containers:
         - image: quay.io/kubermatic/chrome-headless:v1.6.0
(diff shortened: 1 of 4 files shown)


ahmedwaleedmalik (commented):
Looks good but the CI jobs are failing 🕵️‍♂️

xrstf:
I wouldn't know if those failures are related to my PR :/

ahmedwaleedmalik:
> I wouldn't know if those failures are related to my PR :/

I can take a look tomorrow

ahmedwaleedmalik (approved):
/approve

Issue #4819, opened 2022-08-15
Modify External Cluster List Page 


imharshita:
**What would you like to be added? / User Story**:
- Replace Labels on External Cluster list with more relevant and important information
`Kubernetes Version` and `Region`
![ext](https://user-images.githubusercontent.com/78943096/184595283-f130857f-6f97-4cb4-b1c3-6a889153655b.png)

Remove Labels and Add `Region` and `Kubernetes Version`

KhizerRehan:
This change is currently applicable on Normal clusters as it requires Backend API change to return `Kubernetes Version and Region` in external cluster list so won't be manage in this `KKP 2.21` release hence removing milestone.

cc: @Waseem826

KhizerRehan:
**External Clusters:**

This ticker is blocked by backend for `Stopped` clusters from provider side version is not returned so either it requires change at CRD level to store version from backend side.

It also handles edge case what if user tries to update existing version from Version Upgrade picker will it show updated version in list endpoint or old one till the time it is processing or in case of failure any update to notify user if upgrade failed to update.

cc: @imharshita

PR #5209, opened 2022-11-07
Remove Labels from Table Views

### LINKED ISSUE #5086: Remove Labels from Table Views
### Remove labels from
- [ ] Cluster table
- [ ] External Cluster table related #4819
- [ ] Machine Deplyoment table (on cluster page)

### Adjust labels on project table
- [ ] Show labels as counter like we do on card view
- [ ] Decrease width reserved for label column 
image

KhizerRehan:

https://user-images.githubusercontent.com/17727069/200272300-051d6e92-a731-4ef3-8fb6-9ca98750b9df.mp4

**AWS Provider Note: => (Displays Extra Column "Availaibitly Zone")**

aws-provider

**Other Providers:**
provider-other-than-aws

Fixes #5086 #4819 

**What type of PR is this?**
/kind design
/kind chore

In case Kubernetes version is not displayed for External clusters see this https://github.com/kubermatic/dashboard/issues/4819#issuecomment-1220711126

```release-note
NONE
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/src/app/cluster/list/external-cluster/template.html
+++ b/src/app/cluster/list/external-cluster/template.html
@@ -69,7 +69,10 @@ limitations under the License.
             *matCellDef="let element"
             [attr.id]="'km-clusters-' + element.name">
           <div fxLayoutAlign=" center"
-               fxLayoutGap="8px">{{element.name}}
+               fxLayoutGap="8px">
+            <span>{{element.name}}</span>
+            <span *ngIf="(!element?.labels || element?.labels['is-imported'] !== 'true')"
+                  class="km-label-primary">Imported</span>
           </div>
         </td>
       </ng-container>
@@ -77,12 +80,26 @@ limitations under the License.
       <ng-container [matColumnDef]="Column.Provider">
         <th mat-header-cell
             *matHeaderCellDef
-            class="km-header-cell">Provider
+            class="km-header-cell p-20">Provider
         </th>
         <td mat-cell
             *matCellDef="let element">
           <span [ngClass]="getProvider(element.cloud) === Provider.Custom ?
               'km-provider-logo km-provider-logo-custom-external-cluster' : 'km-provider-logo km-provider-logo-'+getProvider(element.cloud)"></span>
+        </td>
+      </ng-container>
+
+
+      <ng-container [matColumnDef]="Column.Region">
+        <th mat-header-cell
+            *matHeaderCellDef
+            class="km-header-cell p-20">Region
+        </th>
+        <td mat-cell
+            *matCellDef="let element">
+          <ng-container *ngIf="element.cloud">
+            {{getRegion(element.cloud)}}
+          </ng-container>
         </td>
       </ng-container>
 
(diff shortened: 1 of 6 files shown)


src/app/cluster/list/external-cluster/template.html
```diff
@@ -69,47 +69,52 @@
             *matCellDef="let element"
             [attr.id]="'km-clusters-' + element.name">
           <div fxLayoutAlign=" center"
-               fxLayoutGap="8px">{{element.name}}
+               fxLayoutGap="8px">
+            <span>{{element.name}}</span>
+            <span *ngIf="(!element?.labels || element?.labels['is-imported'] !== 'true')"
```
ahmadhamzh, line 74:
```suggestion
            <span *ngIf="(!element?.labels || element?.labels['is-imported'] === 'true')"
```

src/app/cluster/list/external-cluster/template.html
```diff
@@ -69,47 +69,52 @@
...
-          <span [ngClass]="getProvider(element.cloud) === Provider.Custom ?
-              'km-provider-logo km-provider-logo-custom-external-cluster' : 'km-provider-logo km-provider-logo-'+getProvider(element.cloud)"></span>
+          <ng-container *ngIf="element.cloud">
+            {{getRegion(element.cloud)}}
+          </ng-container>
         </td>
       </ng-container>
 
-      <ng-container matColumnDef="created">
+      <ng-container [matColumnDef]="Column.Created">
```
ahmadhamzh, line 106: maybe we need to change it to `Column.Imported`

KhizerRehan: Actually label was correct shown in UI but from Code perspective it was created. FYI here `created` was due to the reason because external cluster can be created via wizard or can be imported by dialog but for both still show Import as heading but actually underlying it is `creationTimeStamp` managed by backend but you were right for better CONTEXT it was better to rename column **binding** name

src/app/cluster/list/external-cluster/template.html
KhizerRehan: It is fine! delete button will be shown only for **non-imported** clusters

src/app/cluster/list/external-cluster/template.html
```diff
@@ -69,47 +69,52 @@
...
             *matHeaderCellDef
-            class="km-header-cell p-30">Labels
+            class="km-header-cell p-20">Provider
         </th>
         <td mat-cell
             *matCellDef="let element">
-          <km-labels [labels]="element.labels"
-                     [limit]="2"></km-labels>
+          <span [ngClass]="getProvider(element.cloud) === Provider.Custom ?
+              'km-provider-logo km-provider-logo-custom-external-cluster' : 'km-provider-logo km-provider-logo-'+getProvider(element.cloud)"></span>
```
Talha-Jamil-TJ, line 88: `km-provider-logo` can just be moved into `class` instead of `ngClass`

Waseem826 (approved):

src/app/cluster/list/external-cluster/template.html
KhizerRehan: will be fixed in sep PR due to following reasons

**Adding comment for posterity reasons:**

- it is part of existing code nothing updated diff displaying due to `Label` html is removed so beneath code is moved up
- Changes regarding code improvement where `km-provider-logo` can be moved in class just for duplication will be handled in other issues e.g Code Refactoring ticket

PR #5258, opened 2022-11-12
Store a MachineDeployment instead of NodeDeployment in the initial-md annotation

xrstf:
With https://github.com/kubermatic/kubermatic/pull/11339, KKP removes its own ability to deal with NodeDeployments. Instead it will expect the initial-machine-deployment annotation to contain a `clusterv1alpha1.MachineDeployment` encoded as JSON. KKP will then parse it and apply the current SSH keys from the cluster object, so the API does not have to provide the selected keys as long as the API makes sure to properly bind the keys to the cluster.

This PR therefore changes the API to generate a MachineDeployment.

I guess we might need a step in the KKP installer to check that no clusters exists that still have an initial-md annotation with a NodeDeployment when updating KKP from 2.21 to 2.22?

**What type of PR is this?**
/kind chore

```release-note
The `kubermatic.io/initial-machinedeployment-request` annotation now stores a `clusterv1alpha1.MachineDeployment` instead of an `apiv1.NodeDeployment`.
```

**Documentation**:
```documentation
NONE
```


xrstf:
this needs lgtm, not approve ;)

PR #5266, opened 2022-11-14
Option to enforce custom disk for OpenStack machines

### LINKED ISSUE #5243: Option to enforce custom disk for OpenStack machines
As a user, I'd like the ability to enforce the usage of custom disks for OpenStack machines. 

Currently, when the custom disk option is not checked ephemeral storage is used for the machines and the chosen flavor decides the size of the machine. Although, when this field is checked, volumes are used for storage.

Reference: https://github.com/kubermatic/machine-controller/blob/release/v1.54/examples/openstack-machinedeployment.yaml#L147

It would be nice having the ability to force the root disk to always be created as a volume(cinder volume).

NOTE: This change is required solely for the dashboard.

ahmedwaleedmalik:

This PR adds an option to enforce custom disks on OpenStack. This setting can be toggled from the admin settings.

Fixes #5243

**What type of PR is this?**

/kind feature

```release-note
Introduced an option in admin settings to enforce custom disks for OpenStack Machines.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/defaults/template.html
+++ b/modules/web/src/app/settings/admin/defaults/template.html
@@ -278,3 +278,44 @@ limitations under the License.
     </div>
   </mat-card-content>
 </mat-card>
+
+<mat-card>
+  <mat-card-header>
+    <mat-card-title fxFlex
+                    fxLayout="row"
+                    fxLayoutAlign=" center">
+      <div fxFlex>Provider Defaults</div>
+    </mat-card-title>
+  </mat-card-header>
+  <mat-card-content>
+    <div fxLayout="column"
+         fxLayoutGap="32px"
+         class="admin-settings">
+      <div fxFlex
+           fxLayout="column"
+           fxLayoutGap="16px">
+
+        <div fxLayout="row"
+             fxLayoutAlign=" center">
+          <div fxFlex="16%"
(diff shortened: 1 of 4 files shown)


modules/web/src/app/settings/admin/defaults/template.html
```diff
@@ -278,3 +278,44 @@
...
+                    fxLayout="row"
+                    fxLayoutAlign=" center">
+      <div fxFlex>Provider Defaults</div>
+    </mat-card-title>
+  </mat-card-header>
+  <mat-card-content>
+    <div fxLayout="column"
+         fxLayoutGap="32px"
+         class="admin-settings">
+      <div fxFlex
```
KhizerRehan, line 294:
```suggestion
  <div fxFlex
           fxLayout="column"
           fxLayoutGap="16px">
```
I don't think so we need these inner div with following layout. Since parent does have already attributes to have a column based layout and inner child's will be always row based.,


But since it used above in template so approving it so that when change is done it is reflected al over the file

ahmedwaleedmalik: Just added it for consistency since we are planning on adding more fields in this sub-component.

Waseem826 (approved):

PR #5278, opened 2022-11-15
Use the same version for the KKP installer as is pinned in go.mod

xrstf:
This is to make the dashboard API tests more reliable. Instead of following KKP and breaking everytime a change is made over there, we now deduce the KKP version based on what we have in our go.mod. With a bit of Github hackery we can translate Go's pseudo version into the full Git commit hash.

/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


xrstf:
> [2022-11-15T12:20:10+00:00] KKP Docker image tag: e6c193aeffb075ff115e3b266bbb7ac406b61696

Seems to work just fine :)

PR #5315, opened 2022-11-22
Resource quota calculation

### LINKED ISSUE #5313: KKP API - add Resource Quota endpoint which given a quota and a node calculates and returns used quota
The UI would like to have this endpoint so that it can dynamically show how would the quota usage change for a chosen node.

It should receive the quota subject in question (in our case just project for now), and the machine details which are needed for the quota calculation. 

It should return:
- the resource quota after the calculation (current quota + the machine details)
- bool if it exceeds the quota (so that the UI doesnt have to figure it out)
- message what was exceeded (if something did exceed)

Of course this is flexible but as a first idea on what the UI needs for this function.

lsviben:
Adds endpoint for resource quota calculation, based on the current project quota + the chose node from a provider.

Fixes #5313 

**What type of PR is this?**
/kind feature

```release-note
[EE] Added an resource quota endpoint which given a provider node size and replica count, returns the calculation of what the projects resource quota usage would be and a message if the quota is exceeded.
`POST /api/v2/projects/{project_id}/quotacalculation`
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/modules/api/pkg/ee/resource-quota/handler.go
+++ b/modules/api/pkg/ee/resource-quota/handler.go
@@ -287,7 +287,8 @@ func getResourceDetailsFromRequest(req calculateProjectResourceQuotaUpdate) (*ku
 
 	var err error
 
-	if req.Body.AlibabaInstanceType != nil {
+	switch {
+	case req.Body.AlibabaInstanceType != nil:
 		nc.WithCPUCount(req.Body.AlibabaInstanceType.CPUCoreCount)
 
 		if err = nc.WithMemory(int(req.Body.AlibabaInstanceType.MemorySize), "G"); err != nil {
@@ -296,7 +297,7 @@ func getResourceDetailsFromRequest(req calculateProjectResourceQuotaUpdate) (*ku
 		if err = nc.WithStorage(req.Body.DiskSizeGB, "G"); err != nil {
 			return nil, err
 		}
-	} else if req.Body.AnexiaNodeSpec != nil {
+	case req.Body.AnexiaNodeSpec != nil:
 		nc.WithCPUCount(req.Body.AnexiaNodeSpec.CPUs)
 
 		if err = nc.WithMemory(int(req.Body.AnexiaNodeSpec.Memory), "M"); err != nil {
@@ -310,7 +311,7 @@ func getResourceDetailsFromRequest(req calculateProjectResourceQuotaUpdate) (*ku
 		if err = nc.WithStorage(int(diskSize), "G"); err != nil {
 			return nil, err
 		}
-	} else if req.Body.AWSSize != nil {
+	case req.Body.AWSSize != nil:
 		nc.WithCPUCount(req.Body.AWSSize.VCPUs)
 
 		if err = nc.WithMemory(int(req.Body.AWSSize.Memory), "G"); err != nil {
@@ -319,7 +320,8 @@ func getResourceDetailsFromRequest(req calculateProjectResourceQuotaUpdate) (*ku
 		if err = nc.WithStorage(req.Body.DiskSizeGB, "G"); err != nil {
 			return nil, err
 		}
-	} else if req.Body.AzureSize != nil {
+
+	case req.Body.AzureSize != nil:
 		nc.WithCPUCount(int(req.Body.AzureSize.NumberOfCores))
 
 		if err = nc.WithMemory(int(req.Body.AzureSize.MemoryInMB), "M"); err != nil {
@@ -329,7 +331,7 @@ func getResourceDetailsFromRequest(req calculateProjectResourceQuotaUpdate) (*ku
 		if err = nc.WithStorage(int(req.Body.AzureSize.ResourceDiskSizeInMB+req.Body.AzureSize.OsDiskSizeInMB), "M"); err != nil {
 			return nil, err
 		}
-	} else if req.Body.DOSize != nil {
+	case req.Body.DOSize != nil:
 		nc.WithCPUCount(req.Body.DOSize.VCPUs)
 
 		if err = nc.WithMemory(req.Body.DOSize.Memory, "M"); err != nil {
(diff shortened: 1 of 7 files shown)


modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -172,35 +214,373 @@ func GetResourceQuota(ctx context.Context, request interface{}, provider provide
...
 	quotaProvider provider.ResourceQuotaProvider) (*apiv2.ResourceQuota, error) {
-	req, ok := request.(common.GetProjectRq)
+	projectResourceQuota, projectName, err := getProjectResourceQuota(ctx, request, projectProvider, privilegedProjectProvider, userInfoGetter, quotaProvider)
+	if err != nil {
+		return nil, err
+	}
+
+	if projectResourceQuota == nil {
+		// ResourceQuota not found. Return an empty response.
+		return nil, nil
```
pkprzekwas, line 224: Shouldn't we return 404 not found?

modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -172,35 +214,373 @@ func GetResourceQuota(ctx context.Context, request interface{}, provider provide
...
 	}
+
+	projectResourceQuota, projectName, err := getProjectResourceQuota(ctx, req.GetProjectRq, projectProvider, privilegedProjectProvider, userInfoGetter, quotaProvider)
+	if err != nil {
+		return nil, err
+	}
+
+	if projectResourceQuota == nil {
+		// ResourceQuota not found. Return an empty response.
+		return nil, nil
```
pkprzekwas, line 245: From what I see, this code probably never executes. In case projectResourceQuota is not found kube returns 404 error (in the `getProjectResourceQuota` function)

modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -172,35 +214,373 @@ func GetResourceQuota(ctx context.Context, request interface{}, provider provide
 func GetResourceQuotaForProject(ctx context.Context, request interface{}, projectProvider provider.ProjectProvider,
 	privilegedProjectProvider provider.PrivilegedProjectProvider, userInfoGetter provider.UserInfoGetter,
 	quotaProvider provider.ResourceQuotaProvider) (*apiv2.ResourceQuota, error) {
-	req, ok := request.(common.GetProjectRq)
+	projectResourceQuota, projectName, err := getProjectResourceQuota(ctx, request, projectProvider, privilegedProjectProvider, userInfoGetter, quotaProvider)
+	if err != nil {
+		return nil, err
+	}
+
+	if projectResourceQuota == nil {
```
pkprzekwas, line 222: From what I see, this code probably never executes. In case projectResourceQuota is not found kube returns 404 error (in the getProjectResourceQuota function)

modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -172,35 +214,373 @@ func GetResourceQuota(ctx context.Context, request interface{}, provider provide
...
+		}
+	case req.Body.VMDirectorNodeSpec != nil:
+		if err = getVMCloudDirectorResourceDetails(req, nc); err != nil {
+			return nil, err
+		}
+	default:
+		return nil, fmt.Errorf("provider set in request not supported: %v", req.Body)
+	}
+
+	// These should not happen, unless we have a bug in the setting resource values code above
```
pkprzekwas, line 344: Is it because provider's cpu count is always an `int` and it's defaulted to 0?

modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -172,35 +214,373 @@ func GetResourceQuota(ctx context.Context, request interface{}, provider provide
...
+		return err
+	}
+	if err := nc.WithStorage(req.Body.DOSize.Disk, "G"); err != nil {
+		return err
+	}
+	return nil
+}
+
+func getEquinixResourceDetails(req calculateProjectResourceQuotaUpdate, nc *kubermaticprovider.NodeCapacity) error {
+	var cpuCount int
```
pkprzekwas, line 435: nit: I know it's golang, but I would still prefer to init this variable explicitly with 0.
```suggestion
	cpuCount := 0
```

modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -172,35 +214,373 @@ func GetResourceQuota(ctx context.Context, request interface{}, provider provide
...
+	if err := nc.WithMemory(req.Body.VSphereNodeSpec.Memory, "G"); err != nil {
+		return err
+	}
+
+	if req.Body.VSphereNodeSpec.DiskSizeGB != nil {
+		if err := nc.WithStorage(int(*req.Body.VSphereNodeSpec.DiskSizeGB), "G"); err != nil {
+			return err
+		}
+	} else {
+		nc.Storage = &resource.Quantity{}
```
pkprzekwas, line 536: Consider adding unit test to cover scenario where DiskSize is not set.

modules/api/pkg/ee/resource-quota/handler.go
lsviben: It does execute because of the way the getProjectResourceQuota works, specifically the resource quota provider:
https://github.com/kubermatic/dashboard/blob/03648c694cca330f62ad519cac4338a2bd9af6f5/modules/api/pkg/ee/resource-quota/provider.go#L98-L101

I am returning nil here because it seems to be an agreement between UI and API, as the GetResourceQuotaForProject worked the same for now. Probaby 404 is better, but I guess UI would then need to silence that error for projects which dont have resource quotas set

modules/api/pkg/ee/resource-quota/handler.go
lsviben: Changed it now to default the values. It seems more intuitive then throwing an error, as the implementations for providers can vary

modules/api/pkg/ee/resource-quota/handler.go
pkprzekwas: Damn, you're actually right. In that case we need to leave it.

PR #5319, opened 2022-11-22
limit husky to fe in pre-commit hook

SimonTheLeg:

This PR ensures that husky pre-commit and pre-push hooks exit early if there are no changes made to `modules/web`

**What type of PR is this?**

/kind chore

There is one caveat to the current solution: Let's assume you have branch that has commits in the `modules/web` folder. If you are now making a new commit which does not change the folder, it will still run the pre-commit hook. The reason for this is because it compares your current ref with main.

There is a really weird reason though why we need to compare with main: Running `git diff --quiet` on a branch that has only new files staged, fails because new files cannot be diffed against event if they just have been added
Here's a full example

```sh
❯ touch modules/web/file
❯ git add modules/web/file
❯ git status
# On branch: limit-husky-to-fe  |  [*] => $e*
#
➤ Changes to be committed
#
#       new file:  [1] modules/web/file
#
❯ git diff --quiet -- "$(git rev-parse --show-toplevel)/modules/web" # will succeed, even though it should fail
❯ git diff --quiet main -- "$(git rev-parse --show-toplevel)/modules/web" # fails as expected
```

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/.husky/pre-commit
+++ b/modules/web/.husky/pre-commit
@@ -16,5 +16,12 @@
 . "$(dirname "$0")/_/husky.sh"
 cd modules/web/.husky
 
+# If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
+# Since we cd into husky dir, run git diff based on git root directory
+if git diff --quiet main -- "$(git rev-parse --show-toplevel)/modules/web" ; then
+  echo "No changes detected in modules/web, skipping pre-commit hook"
+  exit 0
+fi
+
 npx lint-staged
 npm run check
--- a/modules/web/.husky/pre-commit
+++ b/modules/web/.husky/pre-commit
@@ -18,10 +18,12 @@ cd modules/web/.husky
 
 # If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
 # Since we cd into husky dir, run git diff based on git root directory
-if git diff --quiet main -- "$(git rev-parse --show-toplevel)/modules/web" ; then
-  echo "No changes detected in modules/web, skipping pre-commit hook"
+web_dir_path="$(git rev-parse --show-toplevel)/modules/web"
+
+if git diff --quiet -- $web_dir_path || git diff --quiet --cached -- $web_dir_path
+then
+  npx lint-staged
+  npm run check
+else
   exit 0
 fi
-
-npx lint-staged
-npm run check
--- a/modules/web/.husky/pre-push
+++ b/modules/web/.husky/pre-push
@@ -16,4 +16,9 @@
 . "$(dirname "$0")/_/husky.sh"
 cd modules/web/.husky
 
+# If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
+if git diff --quiet --cached main -- "$(git rev-parse --show-toplevel)/modules/web"; then
+  exit 0
+fi
+
 npm run test


modules/web/.husky/pre-commit
```diff
@@ -16,5 +16,12 @@
 . "$(dirname "$0")/_/husky.sh"
 cd modules/web/.husky
 
+# If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
+# Since we cd into husky dir, run git diff based on git root directory
+if git diff --quiet main -- "$(git rev-parse --show-toplevel)/modules/web" ; then
+  echo "No changes detected in modules/web, skipping pre-commit hook"
```
ahmedwaleedmalik, line 22: We don't need this message, it's better to just silently exit.

modules/web/.husky/pre-commit
```diff
@@ -16,5 +16,12 @@
 . "$(dirname "$0")/_/husky.sh"
 cd modules/web/.husky
 
+# If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
+# Since we cd into husky dir, run git diff based on git root directory
+if git diff --quiet main -- "$(git rev-parse --show-toplevel)/modules/web" ; then
```
ahmedwaleedmalik, line 21: `main` is not required here.

ahmedwaleedmalik (changes requested):

modules/web/.husky/pre-commit
SimonTheLeg: To quote myself from the description

> There is a really weird reason though why we need to compare with main: Running git diff --quiet on a branch that has only new files staged, fails because new files cannot be diffed against even if they just have been added
Here's a full example:

```sh
❯ touch modules/web/file
❯ git add modules/web/file
❯ git status
# On branch: limit-husky-to-fe  |  [*] => $e*
#
➤ Changes to be committed
#
#       new file:  [1] modules/web/file
#
❯ git diff --quiet -- "$(git rev-parse --show-toplevel)/modules/web" # will succeed, even though it should fail
❯ git diff --quiet main -- "$(git rev-parse --show-toplevel)/modules/web" # fails as expected
```

ahmedwaleedmalik: Yeah but that reasoning is not valid. Those stages are staged so they'll only appear when you do something like `git diff --cached`. TLDR in the `pre-commit` hook you can simply do something like :

```
. "$(dirname "$0")/_/husky.sh"
cd modules/web/.husky

# If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
# Since we cd into husky dir, run git diff based on git root directory
web_dir_path=$(git rev-parse --show-toplevel)/modules/web

if git diff --quiet --name-only -- $web_dir_path || git diff --quiet --name-only --cached -- $web_dir_path
then
  npx lint-staged
  npm run check
else
  exit 0
fi

```

modules/web/.husky/pre-push
```diff
@@ -16,4 +16,10 @@
 . "$(dirname "$0")/_/husky.sh"
 cd modules/web/.husky
 
+# If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
+if ! git diff --name-only main | grep modules/web > /dev/null 2>&1; then
```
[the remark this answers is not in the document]
SimonTheLeg: no particular reason. I think I finished the final version of this script earlier. But you are right `--cached` also does the trick. fixed.

modules/web/.husky/pre-commit
SimonTheLeg: oh nice, this does the trick!

modules/web/.husky/pre-push
```diff
@@ -16,4 +16,9 @@
 . "$(dirname "$0")/_/husky.sh"
 cd modules/web/.husky
 
+# If there are no changes in modules/web, exit the hook early so we don't run on BE-only changes.
+if git diff --quiet --cached main -- "$(git rev-parse --show-toplevel)/modules/web"; then
+  exit 0
+fi
+
 npm run test
```
ahmedwaleedmalik, line 24:
```suggestion
if git diff --quiet --cached -- "$(git rev-parse --show-toplevel)/modules/web"
then
  npm run test
else
  exit 0
fi
```

Basically you need to reverse the condition.

ahmedwaleedmalik (changes requested):

modules/web/.husky/pre-push
SimonTheLeg: If I read the git doc correctly ["it exits with 1 if there were differences and 0 means no differences"](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---exit-code). For our if statement this would mean that if statement was successful (aka exit code 0) there would be no changes, thus implying you can leave early.

Or am I missing something?

ahmedwaleedmalik (approved):
/approve

Sorry for the long delay in the review cycle.

modules/web/.husky/pre-push
ahmedwaleedmalik: I misunderstood it completely, you are absolutely right here.

PR #5330, opened 2022-11-23
Add project-scoped API endpoints for GCP

### LINKED ISSUE #5132: Add project-scoped API endpoints for GCP
As part of kubermatic/kubermatic#7260, we need to provide KKP API endpoints that are "project-aware", e.g. from `/v2/providers/gke/sizes` to `/v2/projects/{project_id}/providers/gke/sizes` because those API endpoints can use `Presets` as credentials, and we need information about the project context to determine whether a project-scoped preset can be used or not.

## Acceptance Criteria

- All provider-specific API endpoints for GCP gain a project-specific equivalent.

shatoboar:
This PR Adds a project-scoped API for GCP as described in the issue #5132 

Fixes #5132

**What type of PR is this?**

/kind api-change

```release-note
Add API endpoints `/api/v2/projects/{project_id}/providers/gcp/vmsizes` `/api/v2/projects/{project_id}/providers/gcp/zones` `/api/v2/projects/{project_id}/clusters/{cluster_id}/providers/gcp/networks` `/api/v2/projects/{project_id}/providers/gcp/subnetworks` `/api/v2/projects/{project_id}/disktypes`
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik:
@shatoboar you need to update the generated API client and swagger spec. Please run the following scripts:

```
hack/gen-api-client.sh
hac/update-swigger.sh
```

modules/api/pkg/handler/v2/provider/gcp.go
```diff
@@ -60,6 +63,61 @@ type gcpSubnetworksNoCredentialReq struct {
 	Network string
 }
 
+type GCPCommonReq struct {
+	ServiceAccount string
+	Credential     string
+}
+
+type GCPProjectCommonReq struct {
+	Request   GCPCommonReq
```
embik, line 72: This isn't exactly how the `Req` structs should look (to be fair it could, but we should be consistent with how things are implemented so far). What you want is to embed the common structures into your structs, so `GCPProjectCommonReq` should look like this:

```go
type GCPProjectCommonReq struct {
   common.ProjectReq // <- embed a project-aware req
   GCPCommonReq // <- embed the common req for GCP
   
   ProjectID string // <- the special field for this request
}
```

Respectively applies to all other structs as well, so `GCPProjectDatacenterReq` could look like this:

```go
type GCPProjectDatacenterReq struct {
	GCPProjectCommonReq
	DC             string
}
```

or

```go
type GCPProjectDatacenterReq struct {
    common.ProjectReq
	GCPCommonReq
	
	DC             string
}
```

shatoboar:
@ahmedwaleedmalik I can't seem to find those files. Are those generation scripts in some other repository?

modules/api/pkg/handler/v2/provider/gcp.go
```diff
@@ -60,6 +63,61 @@ type gcpSubnetworksNoCredentialReq struct {
...
+	ProjectID string
+}
+
+type GCPVMReq struct {
+	Request GCPCommonReq
+	Zone    string
+}
+
+type GCPProjectVMReq struct {
```
embik, line 81: The various `Req` structs need to be commented/annotated for swagger to know which API endpoints use it. Here's an example from Nutanix: https://github.com/kubermatic/dashboard/blob/91bdd42fa485303eda47aa1fefbc9335a9502814/modules/api/pkg/handler/v2/provider/nutanix.go#L102-L107

`listProjectNutanixSubnets` is the identifier for the API endpoint which you assign to it in the "big" swagger comment in `routes_v2.go`.

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -1512,6 +1513,121 @@ func (r Routing) RegisterV2(mux *mux.Router, oidcKubeConfEndpoint bool, oidcCfg
...
+//	Responses:
+//	  default: errorResponse
+//	  200: GCPSubnetworkList
+func (r Routing) listProjectGCPSubnetworks() http.Handler {
+	return httptransport.NewServer(
+		endpoint.Chain(
+			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
+			middleware.UserSaver(r.userProvider),
+		)(provider.ListProjectGCPSubnetworks(r.presetProvider, r.userInfoGetter, r.seedsGetter, true)),
+		provider.DecodeGCPSubnetworksNoCredentialReq,
```
embik, line 1559: All `Req` structs need their own correct `Decode` function. This uses a Decode function that won't include project information, as an example.

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -1512,6 +1513,121 @@ func (r Routing) RegisterV2(mux *mux.Router, oidcKubeConfEndpoint bool, oidcCfg
...
+		handler.EncodeJSON,
+		r.defaultServerOptions()...,
+	)
+}
+
+// swagger:route GET /api/v2/projects/{project_id}/clusters/{cluster_id}/providers/gcp/networks gcp listGCPNetworksNoCredentialsV2
```
embik, line 1565: The identifier for the route (`listGCPNetworksNoCredentialsV2`) needs to be updated here.

modules/api/pkg/handler/v2/provider/gcp.go
```diff
@@ -106,6 +164,278 @@ func DecodeGCPSubnetworksNoCredentialReq(c context.Context, r *http.Request) (in
...
+func ListProjectGCPNetworks(presetProvider provider.PresetProvider, userInfoGetter provider.UserInfoGetter, withProject bool) endpoint.Endpoint {
+	return func(ctx context.Context, request interface{}) (interface{}, error) {
+		var (
+			req       GCPCommonReq
+			projectID string
+			sa        string
+		)
+
+		if !withProject {
+			listReq, ok := request.(GCPProjectCommonReq)
```
embik, line 297: This should probably be `GCPCommonReq` instead of `GCPProjectCommonReq` since you're in the `!withProject` part of the if condition. The `else` branch also needs to be switched over to be project-aware.

modules/api/pkg/handler/v2/routes_v2.go
embik: This applies to all of the `list...()` funcs btw. Here's an example of how one of the decode funcs look like in the Nutanix implementation: https://github.com/kubermatic/dashboard/blob/91bdd42fa485303eda47aa1fefbc9335a9502814/modules/api/pkg/handler/v2/provider/nutanix.go#L193-L210

modules/api/pkg/handler/v2/provider/gcp.go
```diff
@@ -106,6 +164,278 @@ func DecodeGCPSubnetworksNoCredentialReq(c context.Context, r *http.Request) (in
...
+		return "", utilerrors.New(http.StatusInternalServerError, fmt.Sprintf("can not get preset %s for user %s", presetName, userInfo.Email))
+	}
+	credentials := preset.Spec.GKE
+	if credentials == nil {
+		return "", fmt.Errorf("gke credentials not present in the preset %s", presetName)
+	}
+	return credentials.ServiceAccount, nil
+}
+
+func ListProjectGCPDiskTypes(presetProvider provider.PresetProvider, userInfoGetter provider.UserInfoGetter, withProject bool) endpoint.Endpoint {
```
embik, line 196: I know we discussed this with the `withProject` variable so this is my bad, but in this specific case you can simplify your code by dropping the `withProject` boolean and just always assume a project-aware request is passed, simply because the old API endpoints for GCP are in `/v1/` and you needed to copy over the endpoint functions anyway.

Sorry for the confusion, I know I showed you something _with_ this variable.

embik (changes requested):
I found a couple of things to improve, plus what @ahmedwaleedmalik said about updating the generated resources.

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -1512,6 +1513,119 @@ func (r Routing) RegisterV2(mux *mux.Router, oidcKubeConfEndpoint bool, oidcCfg
 		Handler(r.getSeedOverview())
 }
 
+// swagger:route GET /api/v2/projects/{project_id}/disktypes project listProjectGCPDisk
```
embik, line 1516: Should be:

```suggestion
// swagger:route GET /api/v2/projects/{project_id}/providers/gcp/disktypes project listProjectGCPDiskTypes
```

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -1512,6 +1513,119 @@ func (r Routing) RegisterV2(mux *mux.Router, oidcKubeConfEndpoint bool, oidcCfg
...
+			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
+			middleware.UserSaver(r.userProvider),
+		)(provider.ListProjectGCPSubnetworks(r.presetProvider, r.userInfoGetter, r.seedsGetter)),
+		provider.DecodeProjectGCPSubnetworks,
+		handler.EncodeJSON,
+		r.defaultServerOptions()...,
+	)
+}
+
+// swagger:route GET /api/v2/projects/{project_id}/clusters/{cluster_id}/providers/gcp/networks gcp listProjectGCPNetworks
```
embik, line 1563: Should be:

```suggestion
// swagger:route GET /api/v2/projects/{project_id}/providers/gcp/networks gcp listProjectGCPNetworks
```

(without the cluster content).

embik (changes requested):
Two small things to fix and then this should be good to go!

embik:
@shatoboar you'll need to update swagger and the apiclient again via the codegen scripts :)

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -1517,6 +1518,119 @@ func (r Routing) RegisterV2(mux *mux.Router, oidcKubeConfEndpoint bool, oidcCfg
...
+//	Responses:
+//	  default: errorResponse
+//	  200: GCPNetworkList
+func (r Routing) listProjectGCPNetworks() http.Handler {
+	return httptransport.NewServer(
+		endpoint.Chain(
+			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
+			middleware.UserSaver(r.userProvider),
+		)(provider.ListProjectGCPNetworks(r.presetProvider, r.userInfoGetter)),
+		provider.DecodeGCPCommonReq,
```
embik, line 1584: `GCPCommonReq` doesn't include project information, so this probably needs its own request type with project information (don't forget to annotate it with `// swagger:parameters listProjectGCPNetworks`).

embik (changes requested):

embik (approved):
/approve

/cc @ahmedwaleedmalik for sig-api approval

ahmedwaleedmalik (approved):
/approve

PR #5363, opened 2022-11-30
Fix azure disk size display

ahmedwaleedmalik:

The text(hint) displayed for Data and OS Disk Size is wrong. We are always defaulting these values so there is no need for `leave 30 to use default value`. 

This PR also fixes an issue where editing a cluster wouldn't update the `APIServerAllowedIPRanges`. The fix was just a one-liner so I'm not opting for a dedicated PR for that.

Fixes #

**What type of PR is this?**

/kind chore
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik:
/hold
Will be merged after a certain internal session/meeting.

KhizerRehan (approved):

PR #5365, opened 2022-11-30
Bump KKP dependency, remove defaulting package

xrstf:
This bumps KKP to https://github.com/kubermatic/kubermatic/commit/ab124be18d30b7a1a64c2f3386b084fa9dd76106 and removes `pkg/defaulting`, which I could not see a reason for being here. Likewise, `pkg/validation/cluster.go` was removed and instead referenced from upstream.

**What type of PR is this?**
/kind cleanup
/kind chore

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #5377, opened 2022-12-02
Endpoint to retrieve default cluster specifications

### LINKED ISSUE #5375: Introduce an API endpoint to retrieve default cluster specifications
Please check Epic and PR descriptions.

ahmedwaleedmalik:

Endpoint `/providers/{provider_name}/dc/{dc}/defaultcluster` was introduced to retrieve the default cluster spec for the given provider and data center. This will help us lead to our goal of removing/reducing obscurity regarding the cluster object that is created/updated from the dashboard. This endpoint will be consumed at the frontend and will be treated as source of truth for defaulting in the cluster creation wizard, instead of hardcodings at the frontend side.

Fixes #5375

**What type of PR is this?**

/kind feature

```release-note
Endpoint `/providers/{provider_name}/dc/{dc}/cluster` was introduced to retrieve the default cluster spec for the given provider and datacenter
```

**Documentation**:

```documentation
NONE
```


modules/api/pkg/handler/v2/cluster_default/cluster_default.go
```diff
@@ -0,0 +1,284 @@
...
+func (req getDefaultClusterReq) GetSeedCluster() apiv1.SeedCluster {
+	return apiv1.SeedCluster{
+		SeedName: req.seedName,
+	}
+}
+
+// Validate validates getDefaultClusterReq request.
+func (r getDefaultClusterReq) Validate() error {
+	if r.ProviderName == "" {
+		return fmt.Errorf("the provider name cannot be empty")
```
pkprzekwas, line 67: I'm not sure if returning an error here results in a proper http code. As those parameters are parts of path, it probably  doesn't matter, but I think explicitly returning bad request in validation is a good pattern.
```suggestion
		return utilerrors.NewBadRequest("the provider name cannot be empty")
```

modules/api/pkg/handler/v2/cluster_default/cluster_default.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	if !kubermaticv1.IsProviderSupported(r.ProviderName) {
+		return fmt.Errorf("unsupported provider: %q", r.ProviderName)
```
pkprzekwas, line 70: Same as above. I would also consider adding a unit test to check if it indeed returns a proper status code for unsupported provider.

modules/api/pkg/handler/v2/cluster_default/cluster_default.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	if r.DC == "" {
+		return fmt.Errorf("the datacenter cannot be empty")
```
pkprzekwas, line 73: Same as above.

modules/api/pkg/handler/v2/cluster_default/cluster_default.go
ahmedwaleedmalik: It will be transformed to `400` behind the curtains. That's how validation for requests works ;)

modules/api/pkg/handler/v2/cluster_default/cluster_default.go
```diff
@@ -0,0 +1,284 @@
...
+
+func DecodeGetDefaultClusterReq(ctx context.Context, r *http.Request) (interface{}, error) {
+	req := getDefaultClusterReq{
+		ProviderName: mux.Vars(r)["provider_name"],
+		DC:           mux.Vars(r)["dc"],
+	}
+
+	seedName, err := clusterv2.FindSeedNameForDatacenter(ctx, req.DC)
+	if err != nil {
+		return nil, err
```
pkprzekwas, line 86: I suggest wrapping it in some HTTP error. Not sure if this should be 400 or 404. Also it would be nice to cover scenario with non existing seed in units.

ahmedwaleedmalik: Does it really matter? I don't find it to be necessary since this is something that `go-kit` will handle.

PR #5378, opened 2022-12-02
Revert "Fix nil pointer for version"

imharshita:
Reverts kubermatic/dashboard#5373
will revert this pr  https://github.com/kubermatic/dashboard/pull/5367 which caused the issue, so no need for this fix. 
this pr https://github.com/kubermatic/dashboard/pull/5367  needs to be reverted as it doesn't fix the issue of the namespace access in the manager client by using an impersonation client, so the changes seem to be unnecessary.


ahmedwaleedmalik:
/hold

Please add a proper description, of why was the original change made, and why are we reverting it now. Also, add a unit test if possible for this.

PR #5381, opened 2022-12-02
Fix Add Machine Deployment creation

### LINKED ISSUE #5366: Add a MachineDeployment is broken
### What happened
Adding a MachineDeployment is broken for AWS and KubeVirt, probably other providers too (on `dev`)
(maybe linked to https://github.com/kubermatic/dashboard/pull/5258).

### How to reproduce
1- Create an AWS cluster/KubeVirt cluster
2- Add a MachineDeployment

**AWS**
```
method: POST
:scheme: https
:authority: dev.kubermatic.io
:path: /api/v2/projects/f8vwtl5w7s/clusters/q8gxpcn7kx/machinedeployments
```

gives the following error:

```
URL: https://dev.kubermatic.io/api/v2/projects/f8vwtl5w7s/clusters/q8gxpcn7kx/machinedeployments
Status: 500
```
```
{
    "error": {
        "code": 500,
        "message": "failed to create machine deployment from template: unknown operating system"
    }
}
```

**KubeVirt**
Error
```
{
    "error": {
        "code": 500,
        "message": "failed to create machine deployment from template: no machine os was provided"
    }
}
```
 
Both errors related to OS provided.
Probably the same with all providers.

KhizerRehan:

This PR fixes the creation of Machine Deployment creation. In Case of "Add Machine Deployment" default OS is not selected but now based on provider it will select default OS.

[Source of truth for OS Support matrix](https://github.com/kubermatic/machine-controller/blob/main/docs/operating-system.md)

**Anexia:**
anexia

Other Providers "Defaults' Ubuntu:
Add-MD-creation

Fixes #5366 

**What type of PR is this?**

/kind bug
/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/node-data/component.ts
+++ b/modules/web/src/app/node-data/component.ts
@@ -155,10 +155,12 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     if (this.isDialogView()) {
       this.form.addControl(Controls.Kubelet, this._builder.control(''));
       this.dialogEditMode = !!this._nodeDataService.nodeData.name;
-    }
 
-    if (this.dialogEditMode) {
-      this.form.get(Controls.Name).disable();
+      if (this.dialogEditMode) {
+        this.form.get(Controls.Name).disable();
+      } else {
+        this._nodeDataService.operatingSystemSpec = this._getOperatingSystemSpec();
+      }
     }
 
     this._init();
@@ -418,17 +420,16 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     }
   }
 
-  // Ubuntu otherwise
-  private _getDefaultSystemTemplate(provider: NodeProvider): OperatingSystem {
+  private _getDefaultSystemTemplate(provider: NodeProvider): OperatingSystem | null {
     switch (provider) {
       case NodeProvider.VSPHERE: {
         return this._datacenterSpec.spec.vsphere.templates
           ? (Object.keys(this._datacenterSpec.spec.vsphere.templates)[0] as OperatingSystem)
-          : OperatingSystem.Ubuntu;
+          : null;
       }
     }
 
-    return OperatingSystem.Ubuntu;
+    return null;
   }
 
   private _getDefaultOS(): OperatingSystem {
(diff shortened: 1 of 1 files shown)


modules/web/src/app/core/services/node-data/service.ts
```diff
@@ -140,6 +141,18 @@ export class NodeDataService {
     } as NodeData;
   }
 
+  getDefaultOS(): OperatingSystem {
+    if (this.isProvider(NodeProvider.ANEXIA)) {
+      return OperatingSystem.Flatcar;
+    }
+    return OperatingSystem.Ubuntu;
```
ahmedwaleedmalik, line 148: What if the provider doesn't support Ubuntu? Instead of assuming and hardcoding I'd recommend that we should use `isOperatingSystemSupported` to check if it supports ubuntu and only in that case we should return `Ubuntu`. Something that we were already doing in the original method 

https://github.com/kubermatic/dashboard/blob/main/modules/web/src/app/node-data/component.ts#L434

modules/web/src/app/core/services/node-data/service.ts
```diff
@@ -140,6 +141,18 @@ export class NodeDataService {
...
+  }
+
+  isProvider(...providers: NodeProvider[]): boolean {
+    const provider = this._clusterSpecService.provider;
```
ahmedwaleedmalik, line 152: Wouldn't this value be the same as ` this.provider`? Why did we change this?

modules/web/src/app/node-data/component.ts
```diff
@@ -139,7 +139,7 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       ]),
       [Controls.Count]: this._builder.control(this._nodeDataService.nodeData.count),
       [Controls.DynamicConfig]: this._builder.control(this._nodeDataService.nodeData.dynamicConfig),
-      [Controls.OperatingSystem]: this._builder.control(this._getDefaultOS(), [Validators.required]),
+      [Controls.OperatingSystem]: this._builder.control({}, [Validators.required]),
```
ahmedwaleedmalik, line 142: Can we not use an `{}`; declare an object and instead start with a falsy value like null?

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik:
@KhizerRehan is this PR really about `cleanup`? You have mentioned `/kind cleanup` in the description.

KhizerRehan (commented):
Feedback Replies

Waseem826:
I've taken over this PR on behalf of @KhizerRehan and pushed new changes. PTAL

ahmedwaleedmalik (approved):
/approve

Thank you for taking care of this guys 👍

PR #5426, opened 2022-12-13
Integrate API fetch defaults for Clusters

### LINKED ISSUE #5376: Dashboard should rely on API to fetch defaults for Clusters
## What we need

With https://github.com/kubermatic/dashboard/issues/5375 we introduced an API endpoint to fetch a `default cluster` against a provider and data center. We would like to use the data retrieved from that endpoint for defaulting for a cluster instead of relying on multiple sources like hard-coded values in forms, cluster networking endpoints, etc.

## Why it's required

This is essential since right now it's not guaranteed that the cluster is created exactly with the same configuration as set from the dashboard. Since the dashboard does some defaulting, then API, and then our mutating webhooks. 

To avoid such a drift, we can now fetch a default cluster using the new API endpoint, which will already have the defaulting from API and mutating webhooks in place.

Endpoint: `/providers/{provider_name}/dc/{dc}/defaultcluster`

**NOTE:** Please evaluate and remove the usage of `/providers/{provider_name}/dc/{dc}/networkdefaults` from the dashboard. That defaulting is now part of the new endpoint. 

## Side note
@csengerszabo I have set the priority of this ticket to `high` since having this feature available on our dev environment as early as possible would really help with testing this feature thoroughly. I hope that's okay with you. If not, feel free to update it ;)

KhizerRehan:
This PR fetches cluster default values against provider/datacenter to set default value for cluster and can be considered single source of truth. Along with this it removes "networkdefaults" which was used to cater network configuration values based on IP Family selected but now it will be set from same "defaultcluster" endpoint.

https://user-images.githubusercontent.com/17727069/207588186-10258d52-76b4-4791-ba04-99a1e53cbf08.mp4

Fixes #5376 

**What type of PR is this?**
/kind feature
/kind chore

- Tested with local json to see if defaulting was applied on labels, admission plugins

https://user-images.githubusercontent.com/17727069/208856060-59a319be-c011-4539-9dd3-2ab82e034c33.mp4

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -172,76 +174,14 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
 
   // eslint-disable-next-line complexity
```
Waseem826, line 175: We can remove this comment now because code has been moved inside `_initForm` method and we have a comment there.

modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -291,6 +224,21 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(this._setDefaultVersion.bind(this));
 
+    // Cluster defaults in wizard
```
Waseem826, line 227: I think this comment is not required and code is self explanatory.

modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -499,6 +448,167 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
...
+              this._defaultAllowedIPRangeIPv6,
+          });
+        } else {
+          this.form.patchValue({
+            [Controls.IPv4AllowedIPRange]: null,
+            [Controls.IPv6AllowedIPRange]: null,
+          });
+        }
+
+        // Based {emitEvent: false} the valueChanges will not be triggered.
```
Waseem826, line 594: I think this comment doesn't add any value (at least I couldn't understand the context)

modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -499,6 +448,167 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
...
+      [Controls.IPv6AllowedIPRange]: this._builder.control(
+        this.isAllowedIPRangeSupported()
+          ? NetworkRanges.ipv6CIDR(this._getExtraCloudSpecOptions().nodePortsAllowedIPRanges)
+          : '',
+        [IPV6_CIDR_PATTERN_VALIDATOR, this._dualStackRequiredIfValidator(Controls.IPv4AllowedIPRange)]
+      ),
+    });
+  }
+
+  private _loadClusterDefaults(): void {
```
Waseem826, line 523: We're not updating `labels` and `podNodeSelectorAdmissionPluginConfig` in this method, is it intentional?

Waseem826 (commented):
Some feedback. PTAL

modules/web/src/app/wizard/step/cluster/template.html
```diff
@@ -14,6 +14,13 @@
...
 -->
 
+<ng-container *ngIf="loadingClusterDefaults">
+  <mat-spinner [diameter]="25"
+               class="km-spinner"
+               color="accent"></mat-spinner>
+</ng-container>
+
+
 <form [formGroup]="form">
```
Waseem826, line 24: Should we hide the form when cluster defaults are loading?

KhizerRehan: https://user-images.githubusercontent.com/17727069/208647430-f0c79291-1904-4949-94b8-402f9c73ffaa.mp4

Note for reviewer: (Technical note)
- Instead of wrapping form on ngTemplate i have just negated condtion to have smaller changset

Waseem826 (approved):

PR #5450, opened 2022-12-20
Handle default resource quotas

### LINKED ISSUE #5431: Adjust ResourceQuota API for default quotas
When a quota is updated by an user, if it has the Default Quota labels on it, they should be removed to mark the ResourceQuota as an custom quota, instead of a default one. 

This is important to avoid the quota being managed by the default resource quota controller.

lsviben:
Adds handling for default resource quotas:
- info if the quota is default
- on update remove the default label

Fixes #5431

**What type of PR is this?**
/kind feature

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/api/v2/types.go
+++ b/modules/api/pkg/api/v2/types.go
@@ -1713,7 +1713,7 @@ type ResourceQuota struct {
 	SubjectKind string `json:"subjectKind"`
 	// SubjectHumanReadableName contains the human-readable name for the subject(if applicable). Just filled as information in get/list.
 	SubjectHumanReadableName string              `json:"subjectHumanReadableName,omitempty"`
-	Default                  bool                `json:"default"`
+	IsDefault                bool                `json:"isDefault"`
 	Quota                    Quota               `json:"quota"`
 	Status                   ResourceQuotaStatus `json:"status"`
 }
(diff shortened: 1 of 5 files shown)


modules/api/pkg/api/v2/types.go
```diff
@@ -1713,6 +1713,7 @@ type ResourceQuota struct {
 	SubjectKind string `json:"subjectKind"`
 	// SubjectHumanReadableName contains the human-readable name for the subject(if applicable). Just filled as information in get/list.
 	SubjectHumanReadableName string              `json:"subjectHumanReadableName,omitempty"`
+	Default                  bool                `json:"default"`
```
pkprzekwas, line 1716: Do we have any other similar case in KKP codebase? If not I suggest renaming this field to `IsDefault`. Maybe it's just me, but IMO it makes field interpretation much clearer.

modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -668,6 +670,9 @@ func PutResourceQuota(ctx context.Context, request interface{}, provider provide
 	}
 	newResourceQuota := originalResourceQuota.DeepCopy()
 
+	// if a resource quota is updated, its not a default quota anymore. Remove default label if it exists
```
pkprzekwas, line 673:
```suggestion
	// if a resource quota is updated, it's not a default quota anymore. Remove default label if it exists
```

pkprzekwas (changes requested):

pkprzekwas (approved):
/lgtm
/approve

PR #5464, opened 2022-12-22
Cluster details RBAC enhancements.

### LINKED ISSUE #5292: Cluster Service Account UI (RBAC changes)
## Design
https://www.figma.com/file/dJQKgUPzIDND36o2Nq8dct/Loodse-UI?node-id=7906%3A107934&t=2z0lvLUmSxkQWuen-1

### Tables (shown together here, but in live UI only one table should show at a time, changeable via the dropdown on the left) 
image

### View Switcher Dropdown
image

### Dialogs
image

### Service Account Dropdown
image

## Behavior

- Switcher Dropdown switches between table views
- Binding dialog changes depending on which table is active
- When creating a service account binding from the little '+' on the service account row, the same dialog opens as with the normal 'add binding' button, except the service account is prefilled (depending on the row it was clicked on, the respective service account)
- On the Service account dropdown show namespace in grey next to the Service Account name either 1) only if the name is used twice but with different namespaces (because the names are not unique) or 2) always show namespace next to service account in the dropdown 
- download icon: download kubeconfig

KhizerRehan:

This PR support for service account binding (cluster/namespace)

**List Section**

https://user-images.githubusercontent.com/17727069/210199705-671e46e4-1eba-42e8-8d25-bf0cee445d1e.mp4

**Dialogs**

create-sa

sa-binding

**Users/Groups**

user-binding

group-binding

Fixes #5292 

**What type of PR is this?**

/kind feature

```release-note
 Add support for service account creation and (cluster/namespace) binding
```

**Documentation**:

```documentation
TBD
```


modules/web/src/app/cluster/details/cluster/rbac/add-service-account-dialog/template.html
```diff
@@ -0,0 +1,60 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<div class="dialog">
+  <km-dialog-title>Create Service Account</km-dialog-title>
```
Waseem826, line 18: I think dialog title should match the "Add Service Account" button which opens this dialog. Wdyt?
Same for the "Create Service Account" button in this dialog.

modules/web/src/app/cluster/details/cluster/rbac/component.ts
```diff
@@ -37,44 +39,33 @@ import {isClusterRunning} from '@shared/utils/health-status';
...
     private readonly _matDialog: MatDialog,
     private readonly _notificationService: NotificationService,
     private readonly _appConfigService: AppConfigService
   ) {}
 
   ngOnInit(): void {
-    merge(this._refreshTimer$, this._refresh)
+    this._refreshTimer$
       .pipe(
         switchMap(_ => this._clusterService.health(this.projectID, this.cluster.id)),
```
Waseem826, line 64: I think we should take this health value from parent component because adding it here causes 2 health API requests (one from parent component and one here). Wdyt?

modules/web/src/app/cluster/details/cluster/rbac/template.html
```diff
@@ -14,185 +14,119 @@
...
+    </ng-template>
+
+    <div [ngSwitch]="modeControl.value"
+         fxLayout="row"
+         fxLayoutAlign="space-between center"
+         fxLayoutGap="10px">
+      <ng-container *ngSwitchCase="RBACKind.ServiceAccount">
+        <button mat-flat-button
+                color="quaternary"
+                *ngIf="isClusterRunning"
```
Waseem826, line 67: We can remove this `*ngIf` because its already added on parent `ng-container`. Please remove other similar `*ngIf` conditions as well.

modules/web/src/app/cluster/details/cluster/rbac/service-account/component.ts
```diff
@@ -0,0 +1,198 @@
...
+        this.projectID,
+        this.cluster.id,
+        serviceAccount.namespace,
+        serviceAccount.id
+      ),
+      '_blank'
+    );
+  }
+
+  deleteServiceAccount(element: ClusterServiceAccount): void {
```
Waseem826, line 95: We're emitting event to parent for deleting binding so I think we should emit event for this as well and then delete service account in parent component for consistency. This way we also don't need to add `_updateBindings` method here which is duplicate of parent component's method. Wdyt?

modules/web/src/app/cluster/details/cluster/rbac/service-account/component.ts
```diff
@@ -0,0 +1,198 @@
...
+  private _getServiceAccounts(): void {
+    this._clusterServiceAccountService
+      .get(this.projectID, this.cluster.id)
+      .pipe(takeUntil(this._unsubscribe))
+      .subscribe(serviceAccount => {
+        this.dataSource.data = serviceAccount;
+        this.clusterServiceAccountExpansion = serviceAccount.reduce(
+          (prev: Record<string, boolean>, {id}) => ({
+            ...prev,
+            [id]: false,
```
Waseem826, line 140: I think we should not do this because every time service accounts are refreshed it collapses the service account details.


https://user-images.githubusercontent.com/13975988/210236836-3e0fc930-2499-4f61-b763-76e4fb43b540.mov

modules/web/src/app/shared/entity/rbac.ts
```diff
@@ -25,11 +25,21 @@ export class ClusterRole {
 }
 
 export class ClusterBinding {
```
Waseem826, line 27: This class is exactly the same as `Binding` class below but if we want to keep both classes then I think we should rename `Binding` class to `NamespaceBinding` because we're using it in that context. Wdyt?

modules/web/src/app/cluster/details/cluster/rbac/users-or-groups/component.ts
```diff
@@ -0,0 +1,119 @@
...
+export class RBACUsersOrGroupsComponent implements OnInit, OnDestroy {
+  private _unsubscribe = new Subject<void>();
+  readonly Column = Column;
+  readonly RBACType = Kind;
+  columns = [Column.Scope, Column.User, Column.ClusterRole, Column.Namespace, Column.Actions];
+
+  isLoading = true;
+  dataSource = new MatTableDataSource<SimpleClusterBinding>();
+  clusterBindings: ClusterBinding[] = [];
+  namespaceBindings: Binding[] = [];
```
Waseem826, line 50: We're not using these anywhere (other than assigning values to them) so can we remove these?

modules/web/src/app/core/services/cluster-service-account.ts
```diff
@@ -0,0 +1,125 @@
...
+}
+
+@Injectable()
+export class ClusterServiceAccountService {
+  private readonly _restRoot: string = environment.newRestRoot;
+  private readonly _refreshTime = 10;
+  private _refreshTimer$ = timer(0, this._appConfig.getRefreshTimeBase() * this._refreshTime);
+  private _serviceAccountsMap$ = new Map<string, Observable<ClusterServiceAccount[]>>();
+  private _previousServiceAccountsMap = new Map<string, ClusterServiceAccount[]>();
+  private _updateServiceAccount$ = new Subject<void>();
```
Waseem826, line 35:
```suggestion
  private _refreshServiceAccounts$ = new Subject<void>();
```

modules/web/src/app/core/services/cluster-service-account.ts
```diff
@@ -0,0 +1,125 @@
...
+
+      this._serviceAccountsMap$.set(mapKey, serviceAccounts$);
+    }
+    if (this._previousServiceAccountsMap.has(mapKey)) {
+      return this._serviceAccountsMap$.get(mapKey).pipe(startWith(this._previousServiceAccountsMap.get(mapKey)));
+    }
+    return this._serviceAccountsMap$.get(mapKey);
+  }
+
+  update(): void {
```
Waseem826, line 67:
```suggestion
  refresh(): void {
```

modules/web/src/app/core/services/cluster.ts
```diff
@@ -282,8 +286,16 @@ export class ClusterService {
   }
 
   health(projectID: string, clusterID: string): Observable<Health> {
-    const url = `${this._newRestRoot}/projects/${projectID}/clusters/${clusterID}/health`;
-    return this._http.get<Health>(url).pipe(catchError(() => of<Health>({} as Health)));
+    const mapKey = projectID + '-' + clusterID;
```
Waseem826, line 289: I think we should either use `shareReplay` or revert these changes because this method is not called from multiple places on a single view. And there is no benefit in storing previous health value in a map because pending icon is displayed while health is loading which is fine.

modules/web/src/app/core/services/rbac.ts
```diff
@@ -14,33 +14,76 @@
...
+  private _refreshTimer$ = timer(0, this._appConfigService.getRefreshTimeBase() * this._refreshTime);
 
-  constructor(private readonly _http: HttpClient) {}
+  private _refreshClusterBindings$ = new Subject<void>();
+  private _refreshNamespaceBindings$ = new Subject<void>();
+
+  private _clusterNamespacesMap = new Map<string, string[]>();
+  private _namespaceRoleNameMap = new Map<string, RoleName[]>();
+  private _clusterRoleNamesMap = new Map<string, ClusterRoleName[]>();
+  private _clusterBindingMap = new Map<string, ClusterBinding[]>();
```
Waseem826, line 46:
```suggestion
  private _clusterBindingsMap = new Map<string, ClusterBinding[]>();
```

modules/web/src/app/core/services/rbac.ts
```diff
@@ -14,33 +14,76 @@
...
+  private _namespaceBindingsMap = new Map<string, Binding[]>();
+
+  constructor(private readonly _http: HttpClient, private readonly _appConfigService: AppConfigService) {}
 
   getClusterRoleNames(clusterID: string, projectID: string): Observable<ClusterRoleName[]> {
```
Waseem826, line 51: We should use the same approach (with `shareReplay`) as `ClusterServiceAccountService.get` method in all these methods here.

modules/web/src/app/core/services/rbac.ts
```diff
@@ -58,24 +101,51 @@ export class RBACService {
...
+    return this._namespaceBindingsMap.has(mapKey)
+      ? request$.pipe(startWith(this._namespaceBindingsMap.get(mapKey)))
+      : request$;
+  }
+
+  refreshNamespaceBindings(): void {
+    this._refreshNamespaceBindings$.next();
   }
 
   createBinding(
```
Waseem826, line 151:
```suggestion
  createNamespaceBinding(
```
Similarly `deleteNamespaceBinding` below?

modules/web/src/app/wizard/step/cluster/template.html
```diff
@@ -21,7 +21,8 @@
 </ng-container>
 
 
-<form *ngIf="!loadingClusterDefaults" [formGroup]="form">
+<form *ngIf="!loadingClusterDefaults"
+      [formGroup]="form">
```
Waseem826, line 25: This change is already merged in `main` so please rebase this branch otherwise you'll get merge conflict.

Waseem826 (changes requested):
Some feedback. PTAL

KhizerRehan (commented):
Feedback Response

Waseem826 (approved):

Waseem826 (approved):

PR #5470, opened 2022-12-23
Add KubeOneCluster to ExternalCluster API object

imharshita:
Signed-off-by: Harshita sharma <harshita.sharma6174@gmail.com>

Add KubeOneCluster to ExternalCluster API object:
- ProviderName
- Region
- ContainerRuntime

Fixes #

**What type of PR is this?**

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/handler/v2/external_cluster/external_cluster.go
+++ b/modules/api/pkg/handler/v2/external_cluster/external_cluster.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/go-kit/kit/endpoint"
@@ -55,6 +56,7 @@ const (
 	normalType       = "normal"
 	DeleteAction     = "delete"
 	DisconnectAction = "disconnect"
+	Base64           = "^(?:[A-Za-z0-9+\\/]{4})*(?:[A-Za-z0-9+\\/]{2}==|[A-Za-z0-9+\\/]{3}=|[A-Za-z0-9+\\/]{4})$"
 )
 
 // createClusterReq defines HTTP request for createExternalCluster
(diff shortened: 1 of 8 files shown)


modules/api/pkg/handler/v2/external_cluster/external_cluster.go
```diff
@@ -103,12 +105,20 @@ func (req createClusterReq) Validate() error {
 }
 
 func DecodeManifestFromKubeOneReq(encodedManifest string) (*kubeonev1beta2.KubeOneCluster, error) {
+	var err error
+	var manifest []byte
+	rxBase64 := regexp.MustCompile(Base64)
```
hbernardo, line 110: This could be a global var for performance, as `regexp.MustCompile(Base64)` needs to run only once.

modules/api/pkg/kubernetes/helper.go
```diff
@@ -37,18 +42,45 @@ func CheckContainerRuntime(ctx context.Context,
...
+
+	controlPlaneNode, err := ListControlPlaneNode(ctx, clusterClient)
+	if err != nil {
+		return "", fmt.Errorf("failed to fetch container runtime: not able to list control plane nodes %w", err)
+	}
+
+	for _, node := range controlPlaneNode.Items {
+		containerRuntimeVersion := node.Status.NodeInfo.ContainerRuntimeVersion
+		strSlice := strings.Split(containerRuntimeVersion, ":")
+		return strSlice[0], nil
```
hbernardo, line 58: The "for" loop will iterate only the first slice element in that case. Is that intended?

modules/api/pkg/kubernetes/helper.go
```diff
@@ -37,18 +42,45 @@ func CheckContainerRuntime(ctx context.Context,
...
+
+	selector := labels.NewSelector().Add(*nodeReq)
+	if err != nil {
+		return nil, errors.Wrap(err, "error converting node label requirement to selector")
 	}
-	return "", fmt.Errorf("failed to fetch container runtime: no control plane nodes found with label %s", NodeControlPlaneLabel)
+
+	nodes := &corev1.NodeList{}
+	listOpts := &ctrlruntimeclient.ListOptions{
+		Limit:         1,
```
hbernardo, line 77: IMO, this should be a function parameter `limit` instead of constant (1), because the function signature doesn't make it clear that the function will return only one node for the list.

modules/api/pkg/handler/v2/external_cluster/kubeone.go
```diff
@@ -291,3 +298,36 @@ func getKubeOneAPIMachineDeployments(ctx context.Context,
...
+	for _, node := range nodes.Items {
+		region = node.Labels[NodeRegionLabel]
+		containerRuntimeVersion := node.Status.NodeInfo.ContainerRuntimeVersion
+		strSlice := strings.Split(containerRuntimeVersion, ":")
+		containerRuntime = strSlice[0]
+
+		clusterSpec := &apiv2.KubeOneClusterSpec{
+			Region: region,
+		}
+		apiCluster.Spec.ContainerRuntime = containerRuntime
```
hbernardo, line 328: Similar problem here. We have a loop that is filling the same variable attribute. It looks like it shouldn't be a loop.

hbernardo (changes requested):

hbernardo (approved):

PR #5485, opened 2022-12-30
Endpoint for total resource quota availiability/usage

### LINKED ISSUE #5483: Total resource quota/used quota endpoint

Add an endpoint which returns the total resource quota available for all projects and the total usage of the resource quota. For display purposes on the UI

lsviben:

Endpoint for total resource quota availiability/usage. All the existing resource quotas are listed and their total quota amount and used amount are added

Fixes #5483 

**What type of PR is this?**
/kind feature

```release-note
Added an admin API endpoint which shows the total resource quota allocation and usage.
```

**Documentation**:

```documentation
NONE
```


modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -1439,6 +1439,10 @@ func (r Routing) RegisterV2(mux *mux.Router, oidcKubeConfEndpoint bool, oidcCfg
 		Path("/quotas/{quota_name}").
 		Handler(r.getResourceQuota())
 
+	mux.Methods(http.MethodGet).
+		Path("/totalquota").
```
pkprzekwas, line 1443: It seems that GET `quotas` endpoint is still available (you need to double check me on that). If so, how about using it for returning the final sum for all quotas?
```suggestion
		Path("/quotas").
```

pkprzekwas, line 1443: This endpoint's name stands alone a bit. Maybe instead of introducing a new endpoint we could modify existing GET `/quotas` endpoint by introducing additional parameter e.g. `?accumulate=true`. I glanced through the code and it seems possible. With this new approach, using other existing query params for this endpoint, we could get calculation of per project resource quota for free.

pkprzekwas (approved):
/lgtm
/approve

PR #5526, opened 2023-01-10
Default TunnelingAgentIP only when Tunneling expose strategy is set

rastislavs:
Signed-off-by: Rastislav Szabo <rastislav@kubermatic.com>

Dashboard is always sending `TunnelingAgentIP` to the backend no matter which expose strategy is used, which can lead to errors like `TunnelingAgentIP: Forbidden: Tunneling agent IP can be configured only for Tunneling Expose strategy`.
The PR makes sure the default cluster endpoint returns `TunnelingAgentIP` only when the Tunneling expose strategy is set.

Fixes #

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/handler/v2/cluster_default/cluster_default.go
+++ b/modules/api/pkg/handler/v2/cluster_default/cluster_default.go
@@ -156,7 +156,9 @@ func GetDefaultClusterEndpoint(
 
 func mapNetworkDefaultsToCluster(networkDefaults apiv2.NetworkDefaults, cluster *kubermaticv1.Cluster) *kubermaticv1.Cluster {
 	cluster.Spec.ExposeStrategy = networkDefaults.ClusterExposeStrategy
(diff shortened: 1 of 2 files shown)


pkprzekwas (approved):
/lgtm
/approve

pkprzekwas (approved):
/lgtm
/approve

ahmedwaleedmalik:
In general, this should be handled the other way around. The front end should get all the possible defaults and then remove the payload based on the expose strategy. That saves us from sending an API call when we want to change some fields like expose strategy, IP family etc.

rastislavs:
> In general, this should be handled the other way around. Front end should get all the possible defaults and then remove the payload before sending the post/put/patch requests based on the expose strategy. That saves us from sending additional API calls when we want to change some fields like expose strategy, IP family, etc.

@ahmedwaleedmalik ah, okay, I am fine with reverting this if we want to fix this on the frontend side

PR #5544, opened 2023-01-16
Changing Provider after choosing Kubeadm Break the wizard

### LINKED ISSUE #5505: Changing Provider after choosing Kubeadm Break the wizard
**What is the problem?**
after choosing kubeadm in the first step in the wizard and try to chose another provider the wizard still the same wizard for kubeadm (no steps for settings and initial node shown)and an error showed repeatedly in the console 

**Expected behavior **
the user should be able to change the providers in the first step and the wizard's steps will changed according to the provider so when user chose kubeadm provider the wizard step will be 4 (no settings and initial node steps  for kubeadm) and when the user change the provider the steps suppose to be changed to 6 

**How to reproduce the issue?**
1- go to create cluster wizard
2- chose kubeadm the steps will be 4 steps (which is right )
3- chose another provider you will see that the steps number still the same didn't changed (it suppose to be 6)
4-notice the errors in the console

ahmadhamzh:
**What this PR does / why we need it**
fix displaying steps for wizard in case change provider after choosing kubeadm provider

Fixes #5505 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/core/services/wizard/wizard.ts
+++ b/modules/web/src/app/core/services/wizard/wizard.ts
@@ -14,7 +14,7 @@
 
 import {EventEmitter, Injectable} from '@angular/core';
 import {MatStepper} from '@angular/material/stepper';
-import {StepRegistry, WizardStep} from '@app/wizard/config';
+import {StepRegistry, WizardStep, steps} from '@app/wizard/config';
 import {NodeDataService} from '@core/services/node-data/service';
 import {NodeProvider, OperatingSystem} from '@shared/model/NodeProviderConstants';
 import {ClusterSpecService} from '@core/services/cluster-spec';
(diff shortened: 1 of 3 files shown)


modules/web/src/app/core/services/wizard/wizard.ts
```diff
@@ -33,22 +33,15 @@ export class WizardService {
     constructor(private _parent: WizardService) {}
 
     handleProviderChange(provider: NodeProvider): void {
+      this._parent._steps = steps;
```
Waseem826, line 36: We should call the setter method instead.
```suggestion
      this._parent.steps = steps;
```

modules/web/src/app/core/services/wizard/wizard.ts
```diff
@@ -109,6 +102,10 @@ export class WizardService {
     this._clusterSpecService.provider = provider;
   }
 
+  getStepsForEditClusterTemplate(provider: NodeProvider) {
```
Waseem826, line 105: Wdyt?
```suggestion
  forceHandleProviderChange(provider: NodeProvider) {
```

Waseem826 (changes requested):
Some feedback. PTAL

modules/web/src/app/core/services/wizard/wizard.ts
ahmadhamzh: sounds better

Waseem826 (approved):
/approve

PR #5574, opened 2023-01-19
Add missing external cluster error states in `ExternalClusterState`

Waseem826:
Add missing external cluster error states in `ExternalClusterState`.

Fixes #

**What type of PR is this?**

/kind feature

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/shared/entity/external-cluster.ts
+++ b/modules/web/src/app/shared/entity/external-cluster.ts
@@ -94,27 +94,10 @@ export class ExternalCluster {
   }
 
   static getStatusIcon(cluster: ExternalCluster): string {
-    switch (cluster?.status?.state) {
-      case ExternalClusterState.Running:
-        return StatusIcon.Running;
-      case ExternalClusterState.Provisioning:
-      case ExternalClusterState.Reconciling:
-      case ExternalClusterState.ReconcilingUpgrade:
-      case ExternalClusterState.ReconcilingMigrate:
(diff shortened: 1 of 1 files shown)


KhizerRehan (approved):

modules/web/.eslintrc.yml
```diff
@@ -62,7 +62,7 @@ rules:
   prefer-rest-params: error
   complexity:
     - error
-    - 15
+    - 20
```
ahmedwaleedmalik, line 65: @Waseem826 It's better to reduce complexity in your function instead of increasing this in eslint config.

ahmedwaleedmalik (changes requested):

modules/web/.eslintrc.yml
Waseem826: I'll add ts-ignore to mute this error because its not possible to reduce complexity in my scenario

Waseem826: I'll add ts-ignore to mute this error because its not possible to reduce complexity in my scenario

ahmedwaleedmalik: Can't you use a map instead of using switches?

Waseem826: Yea, ended up using map.

ahmedwaleedmalik (approved):
/approve

Issue #5598, opened 2023-01-25
Application Method and Source icons not displayed in cluster wizard summary


Waseem826:
### What happened
Application Method and Source icons are sometimes not displayed on cluster wizard summary step.

![screenshot-localhost_8000-2023 01 25-13_40_00](https://user-images.githubusercontent.com/13975988/214517445-987b0dc8-b7b0-4ad7-9f6c-bb23e45c8cf3.png)

### Expected behavior
Application Method and Source icons should always be displayed on cluster wizard summary step.

![screenshot-localhost_8000-2023 01 25-13_40_44](https://user-images.githubusercontent.com/13975988/214517530-daa77813-203b-40a9-82e3-9cb2b4a2b635.png)

PR #5600, opened 2023-01-25
add method to appdef list endpoint

### LINKED ISSUE #5222: Application method and source information missing in cluster template
### What happened
After recent API changes to applications (https://github.com/kubermatic/dashboard/pull/5184), `method` and `source` information is missing from applications data in cluster template.
Due to this change, `method` and `source` details cannot be displayed for applications in cluster template dialog:

screenshot-localhost_8000-2022 10 06-21_59_16

### Expected behavior
`method` and `source` information is present in applications of cluster template.

SimonTheLeg:

Fixes #5222 

**What type of PR is this?**

/kind bug

To fix what is shown in the screenshots of the original issue, only the field method is required if I understand it correctly. This is added with this PR. Currently this PR purposefully does not include the source field, as it is not required for the versions. Additionally including the source is also a bit more tricky and produces a much larger body, as the source is specified per version of an application.

image
(Screenshot from the create clustertemplate dialogue)

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/cmd/kubermatic-api/swagger.json
+++ b/modules/api/cmd/kubermatic-api/swagger.json
@@ -29108,6 +29108,9 @@
             "type": "string"
           },
           "x-go-name": "Labels"
+        },
+        "method": {
+          "$ref": "#/definitions/TemplateMethod"
         }
       },
       "x-go-package": "k8c.io/dashboard/v2/pkg/api/v2"
--- a/modules/api/pkg/api/v2/types.go
+++ b/modules/api/pkg/api/v2/types.go
@@ -1962,6 +1962,8 @@ type ApplicationDefinitionListItemSpec struct {
 	Labels map[string]string `json:"labels,omitempty"`
 	// Description of the application. what is its purpose
 	Description string `json:"description"`
+	// Method used to install the application
+	Method appskubermaticv1.TemplateMethod `json:"method"`
 }
 
 // ApplicationDefinitionBody is the object representing the POST/PUT payload of an ApplicationDefinition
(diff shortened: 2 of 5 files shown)


Waseem826:
@SimonTheLeg original issue was about Applications in a cluster template and those screenshots (in attached issue) are not from "Create Cluster Template" (which opens the wizard), but those are from "Create Cluster from Template" dialog which is opened from Cluster Templates page. On that dialog we don't load application definitions but only load `applications` from cluster template data. So I think that issue won't be solved by changes in this PR.

https://user-images.githubusercontent.com/13975988/214612359-dec831e9-049d-426f-9d88-7dba0ce58485.mov


That being said, adding `method` in Application Definition list is still helpful since it will help in fixing another [issue](https://github.com/kubermatic/dashboard/issues/5598). But at the same time, since `source` information is not present in Application Definition List so in order to display source icon I might have to cross reference Application Definition from Application Installations and then load details of those Application Definitions.

SimonTheLeg:
@Waseem826 

I see. Luckily it looks like this is fixed as well :)

image

Waseem826:
@SimonTheLeg I think this only fixed the Mode icon but Source icon will still not work. This can be tested if method and source are not the same (e.g. `Helm` and `Git`).

Waseem826:
https://github.com/kubermatic/dashboard/pull/5630 will fix the original issue.

SimonTheLeg:
okay so if I readd the attached PR correctly, then we wanted to display both source and method if they differ. This I did not know about. Okay understood, for this then it makes more sense to directly fetch teh appdef.

If I understand the FE PR correctly this means we do not need this BE PR here anymore, right @Waseem826?

Waseem826:
Yes @SimonTheLeg. We can close this PR I think.

PR #5615, opened 2023-01-30
Fix resource quota calculation - support MD edit

### LINKED ISSUE #5610: Quota calculation - Support MachineDeployment edit
Quota calculation API works like:

- get node details and amount of replicas (in payload)
- based on project, get the resource quota
- calculate node details x amount of replicas 
- return resource quota + calculated quota

This works fine for new clusters and new MachineDeployments, but doesnt cover the case when an user is editing the MD node size or adding more replicas. (https://github.com/kubermatic/dashboard/issues/5584#issuecomment-1404793532)

To support calculating quotas during editing, the API needs to know the existing status. Maybe add an extra field in the payload with:
- existing replicas
- existing node size

then if that field is set, the API can easily calculate the quota

pkprzekwas:
In order to fix resource quota calculation while editing machine deployments, we need to add support for `replacedResources`. This allows to subtract resources of an old type in the final calculation. 

Fixes #5610 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #5631, opened 2023-01-31
Allow User to Go Through the Cluster Template Wizard When Quota Exceeded 

### LINKED ISSUE #5606: Allow to save cluster template in edit and customize mode
### What happened

In case create or edit or customize cluster template, we shouldn't block the user on the initial node step if the quota is exceeded, because the user might **not** want to create a cluster he might want just to create new cluster template so we need in case of quota exceeded to only disable the buttons for creating new cluster

so regarding wizard for cluster template we have three options, in all of them we need to let the user to go through all of the wizard steps:

1- create cluster template :
save template and create cluster button will be disable || save cluster template button will stay enable

2- customize cluster template :
save as new template button will stay enable || create cluster from customized template button will be disable

3- edit cluster template :
save as new template & save changes to template buttons will stay enable

### How to reproduce
**Edit Cluster Template**
- navigate to cluster template
- try to create cluster from template increase replica count or provider specific values that effect quota calculation to make quota exceeded
- Next button will be disabled which it shouldn't as we can save as template we are not creating resource

ahmadhamzh:

Fixes #5606

**What type of PR is this?**
/kind bug

```release-note
NONE
```
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/node-data/component.ts
+++ b/modules/web/src/app/node-data/component.ts
@@ -49,6 +49,8 @@ import {QuotaCalculationService} from '@dynamic/enterprise/quotas/services/quota
 import {ResourceQuotaCalculationPayload, ResourceQuotaCalculation} from '@shared/entity/quota';
 import {END_OF_DYNAMIC_KUBELET_CONFIG_SUPPORT_VERSION} from '@shared/entity/cluster';
 import {KUBERNETES_RESOURCE_NAME_PATTERN_VALIDATOR} from '@app/shared/validators/others';
+import {WizardMode} from '@app/wizard/types/wizard-mode';
+import {ActivatedRoute} from '@angular/router';
 
 enum Controls {
   Name = 'name',
@@ -107,6 +109,7 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
   endOfDynamicKubeletConfigSupportVersion: string = END_OF_DYNAMIC_KUBELET_CONFIG_SUPPORT_VERSION;
   isLoadingOSProfiles: boolean;
   isEnterpriseEdition = DynamicModule.isEnterpriseEdition;
+  wizardMode: WizardMode;
 
   private _enableOperatingSystemManager: boolean;
   private isCusterTemplateEditMode = false;
@@ -124,6 +127,10 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     return this._clusterSpecService.cluster.spec.version < this.endOfDynamicKubeletConfigSupportVersion;
   }
 
+  get displayQuotaInWizard(): boolean {
+    return this.wizardMode === WizardMode.CustomizeClusterTemplate || !this.wizardMode;
+  }
+
   constructor(
     private readonly _builder: FormBuilder,
     private readonly _nameGenerator: NameGeneratorService,
@@ -135,7 +142,8 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     private readonly _projectService: ProjectService,
     private readonly _quotaCalculationService: QuotaCalculationService,
     private readonly _params: ParamsService,
-    private readonly _cdr: ChangeDetectorRef
+    private readonly _cdr: ChangeDetectorRef,
+    private readonly _route: ActivatedRoute
   ) {
     super();
   }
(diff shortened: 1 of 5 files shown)


modules/web/src/app/core/services/cluster-spec.ts
```diff
@@ -138,6 +139,14 @@ export class ClusterSpecService {
     this.clusterChanges.emit(this._cluster);
   }
 
+  get isQuotaExceeded(): boolean {
```
KhizerRehan, line 142: IMHO, this should be part of QuotaCalculation service instead of defining setter/getter on ClusterSpecService. I understand functionality works and buttons disabled based on setter/getter to avoid cluster creation but what i was thinking to deal with this was.

- Define a private new subject in QuotaCalculationService -> quotaExceeding$ 
- Setting value from here [via QuotaCalculationService](https://github.com/kubermatic/dashboard/pull/5631/files#diff-30006b14684ddc4e0caf296926daf3a82160629ea3ca9b2098de99738b022c49R160)
- Define a `refreshQuotaExceed` method which will be a subject and will be emitting a value based on argument.
- Instead of `get` [quotaExceededWarning](https://github.com/kubermatic/dashboard/pull/5631/files#diff-55f7dbeb664a4f25af95aba20dd29bc7d12a57860fde7a0f510ac4588f8eb026R95) initiate subscription so when ever the above part will be executed it will be catched 

**Pros:**
- Makes more sense that quota exceeding is part of quota calculation
- Readability in case when subscription will be catched it will make behave as it is doing currently
- Defining setter does the job but it contionusly checking template .... which IMO is more memory utilization (I am not sure but what i think due to CONTINOUS checking due to CHANGE DETECTION)


thoughts? I think @Waseem826 input will be better in this context

modules/web/src/app/node-data/component.ts
```diff
@@ -124,6 +127,10 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     return this._clusterSpecService.cluster.spec.version < this.endOfDynamicKubeletConfigSupportVersion;
   }
 
+  get displayQuotaInWizard(): boolean {
+    return this.isEnterpriseEdition && (this.wizardMode === WizardMode.CustomizeClusterTemplate || !this.wizardMode);
```
KhizerRehan, line 131: IMO, keeping `isEnterpriseEdition` in template is more explicit + Adding this WHEN to show quota should be dealt in this getter because we only want to show QuotaWidget only

**Show Widget**
- CustomizeClusterTemplate
- Create by wizard

Other than
- Edit / Customize. -> We don't want to show template that will be more clear from CODE
thoughts? this is my opinion

modules/web/src/app/wizard/template.html
```diff
@@ -139,7 +145,7 @@
                      *ngIf="showCreateClusterButton"
                      icon="km-icon-add"
                      [label]="getPrimaryButtonLabel()"
-                     [disabled]="creating"
+                     [disabled]="creating || quotaExceededWarning"
```
KhizerRehan, line 148: I found a edge case where user is allowed to create a cluster even if it exceeds and user will be able to create.

Steps: 
- navigate to cluster templates
- Lets suppose navigate to `Customize Template` option clusterTemplateID=XXXX&mode=customize-cluster-template
- Try to change MODE -> customize-cluster-template`XYZXYZ`
- Try reach end of cluster summary page

Update Initial node to e.g 20 replicase (Assuming: Quota WILL be exceeded for sure) it is not displaying SINCE quota widget is already hidden but it will be exceeded.
 
-  Button will show "Create Cluster" as Text and `ENABLED` to create Cluster.

IMO, this case should be handled and disallowed to create cluster

KhizerRehan (changes requested):
PTAL

modules/web/src/app/wizard/template.html
ahmedwaleedmalik: @KhizerRehan maybe we can open an issue against that? I think the PR in its current state is fine and this is too much of an edge case.

That issue can help us evaluate if we need a fix in KKP 2.23 or afterward.

ahmedwaleedmalik (approved):
/approve
/retest

ahmedwaleedmalik:
/retest
flake

modules/web/src/app/wizard/template.html
KhizerRehan: Yes! it is not related to this PR. So i will create sep ticket for this.

PR #5634, opened 2023-02-01
Fix Restrict Preset to Projects

### LINKED ISSUE #5632: Preset Limitation to Projects
### What happened
regarding restrict preset to projects we need to fix these issues :
* on select the projects the list contain only the projects that showed in the projects view (so if in the select project view all projects is displayed then the list of projects in the create preset dialog will have all project)
* the project list in the payload for create preset is the list of the projects name's
### Expected behavior
- [x] admin should see all the projects in the list always 
- [x] we should send the projects id's instead of name's

### How to reproduce
* disable the option to show all the projects in the projects list view 
* create new preset you will notice that only the projects you own will be in the list 
* after creating the preset check the network (notice the array for projects in the payload have the projects names)

ahmadhamzh:
fix projects in the payload for create preset

Fixes #5632

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/presets/dialog/steps/preset/component.ts
+++ b/modules/web/src/app/settings/admin/presets/dialog/steps/preset/component.ts
@@ -56,7 +56,7 @@ export class PresetStepComponent extends BaseFormValidator implements OnInit {
   readonly domainRegex = '^(?!-)[A-Za-z0-9-]+([\\-.][a-z0-9]+)*\\.[A-Za-z]{2,6}$';
 
   domains: string[] = [];
-  projects: string[] = [];
+  projects: Project[] = [];
   projectLabel = ProjectState.Ready;
 
   constructor(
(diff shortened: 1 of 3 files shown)


modules/web/src/app/settings/admin/presets/dialog/steps/preset/component.ts
```diff
@@ -80,22 +80,31 @@ export class PresetStepComponent extends BaseFormValidator implements OnInit {
...
   }
 
+  projectDisplayFn(projectId: string[]): string[] {
+    if (projectId) {
+      return projectId?.map(id => {
+        return this.projects.find(project => project.id === id)?.name;
+      });
+    }
+    return [];
+  }
```
Waseem826, line 90:
```suggestion
  projectDisplayFn(projectId: string[]): string {
    if (projectId) {
      return projectId.map(id => {
        return this.projects.find(project => project.id === id)?.name;
      }).join(', ');
    }
    return '';
  }
```

Waseem826 (commented):
One suggestion. PTAL

Waseem826 (approved):

PR #5645, opened 2023-02-02
Use router state to communicate `mode` value to cluster wizard

Waseem826:
Use router state to communicate `mode` value to cluster wizard. Previously we were using query params for `mode` which is not ideal since that information is not important for end user.

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
@Waseem826 I think we should merge this before the release. So maybe it should have `code-freeze-approved` label :D

Waseem826:
@ahmedwaleedmalik I'll add it now. I was waiting for the approval 😉

Issue #5286, opened 2022-11-17
Test Admin Panel Seed Configurations - Release 2.22


KhizerRehan:
## Testing Procedure user story

**Display Seeds Configurations**

**Note:**  
- Add new section in admin settings to view seed configurations. 
- User can only **View** seed configuration **ONLY**.
- [Admin Panel Seed UI implementation to display table #5285](https://github.com/kubermatic/dashboard/pull/5285)

#### List Page
- [ ] Navigate to list page and check all seeds general details are shown:
   - Available seeds
   - Available clusters per seed
   - Available providers per seed
   - Available datacenters per seed
   
#### Details Page
- [ ] Navigate to any seed details page by clicking any seed.
- [ ] Validate Seed Panel should display
    - Total Providers
    - Total Datacenters
    - Total Clusters
    - Location
    - Creation Time
- [ ] Display All providers associated to seed and it should display:
    - Total Datacenters associated to each provider.
    - Total Clusters sum for all datacenters.
- [ ] Expanding each provider panel should display:
    - Datacenter name and associated cluster on that datacenter

## Acceptance Criteria
- [ ] List page should display all available seeds.
- [ ] Details page should display seed overview details (providers, datacenters, clusters associated / seed)

ahmedwaleedmalik:
<img width="1517" alt="Screenshot 2023-01-31 at 02 14 57" src="https://user-images.githubusercontent.com/18264334/215596671-46878185-2031-4e0a-8aba-3c35f81c6b76.png">

Sorting doesn't work.

ahmedwaleedmalik:
I have already tested this with https://github.com/kubermatic/dashboard/issues/5711 and it looks good.

PR #5703, opened 2023-02-09
Fix sorting on provider overview table

KhizerRehan:
This PR fixes sorting on Seed Configuration Overview table to sort data by `Provider, Datacenters, Clusters`

https://user-images.githubusercontent.com/17727069/217809327-8423c09e-5fdb-4c16-b334-0ef72c84aa58.mp4

Fixes #
https://github.com/kubermatic/dashboard/issues/5286#issuecomment-1409350941

**What type of PR is this?**
/kind bug

Sorting is applied on overview table where all providers are listed for a particular seed.

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/seed-configurations/seed-configurations-details/component.ts
+++ b/modules/web/src/app/settings/admin/seed-configurations/seed-configurations-details/component.ts
@@ -12,10 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Component, OnInit} from '@angular/core';
+import {Component, OnInit, ViewChild} from '@angular/core';
 import {ActivatedRoute, Router} from '@angular/router';
 import {Subject} from 'rxjs';
 import {finalize, takeUntil} from 'rxjs/operators';
+import {MatSort} from '@angular/material/sort';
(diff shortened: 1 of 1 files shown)


modules/web/src/app/settings/admin/seed-configurations/seed-configurations-details/component.ts
```diff
@@ -80,6 +80,32 @@ export class SeedConfigurationDetailsComponent implements OnInit {
       });
   }
 
+  onSortChange(sort: Sort): void {
```
Waseem826, line 83: Can we use `sortingDataAccessor` instead of manually sorting?
For example,

```
this.dataSource.sortingDataAccessor = (item, property) => {
      switch (property) {
        case Column.Provider:
          return item.provider;
        case Column.Datacenters:
          return item.datacentersCount;
        case Column.Clusters:
          return item.clustersCount;
        default:
          return item[property];
      }
    };

```

For reference take a look at `src/app/dynamic/enterprise/quotas/component.ts` `_setSortConfig` method.

Waseem826 (approved):

PR #5704, opened 2023-02-09
Hotfix: handling for default project quotas

ahmedwaleedmalik:

This PR fixes the patching and retrieval for admin settings.

Due to the mismatch in schema and types for resource quota we have to make some adjustments to the admin settings response from the endpoint to fetch them.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


modules/api/pkg/handler/v1/admin/settings.go
```diff
@@ -140,7 +142,7 @@ func convertAPISettingsToSettingsSpec(settings *apiv2.GlobalSettings) (kubermati
 	}
 
 	if settings.DefaultProjectResourceQuota != nil {
-		crdQuota, err := apiv2.ConvertToCRDQuota(*settings.DefaultProjectResourceQuota)
+		crdQuota, err := apiv2.ConvertToCRDQuota(*&settings.DefaultProjectResourceQuota.Quota)
```
pkprzekwas, line 145:
```suggestion
		crdQuota, err := apiv2.ConvertToCRDQuota(settings.DefaultProjectResourceQuota.Quota)
```

ahmedwaleedmalik: Fixed :D

modules/api/pkg/api/v2/types.go
```diff
@@ -2004,8 +2009,8 @@ type SeedStatus struct {
 	Phase kubermaticv1.SeedPhase `json:"phase"`
 }
 
-// GlobalSettings defines global settings
-// swagger:model GlobalSettings
+// GlobalSettingsV2 defines global settings
+// swagger:model GlobalSettingsV2
```
pkprzekwas, line 2013: Why v2? Do we have same type in `api/v1` and swagger fails?

ahmedwaleedmalik: Yeah, it's failing for everyone, and everyone just `retries` till it passes. Which it eventually does because there is a race condition, sort of, between 2 different specs for the same resource.

ahmedwaleedmalik: But I think we need to change the struct name as well for properly fixing it

ahmedwaleedmalik: Turns out, we don't need the v1 struct anyways.

modules/api/pkg/handler/v1/admin/settings.go
```diff
@@ -140,7 +142,7 @@ func convertAPISettingsToSettingsSpec(settings *apiv2.GlobalSettings) (kubermati
 	}
 
 	if settings.DefaultProjectResourceQuota != nil {
-		crdQuota, err := apiv2.ConvertToCRDQuota(*settings.DefaultProjectResourceQuota)
+		crdQuota, err := apiv2.ConvertToCRDQuota(settings.DefaultProjectResourceQuota.Quota)
```
pkprzekwas, line 145: Do we have any validation to be sure that Quota is not `nil` in this case?

pkprzekwas: It should fail at the marshalling phase in case is no quota, right?

ahmedwaleedmalik: It's a struct so it should work fine, will be an empty object. `Quota` is not a pointer.

ahmedwaleedmalik:
/retest
Just hetzner things.

pkprzekwas (approved):

ahmedwaleedmalik:
@pkprzekwas needs to be code-freeze-approved ;)

PR #5709, opened 2023-02-09
Only take single subscription of project quota API request

Waseem826:
Only take single subscription of project quota API request to prevent continuous polling even after a project's deletion. This issue was caused for non-admin users when they delete a project.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/project/component.ts
+++ b/modules/web/src/app/project/component.ts
@@ -45,7 +45,7 @@ import {MemberUtils, Permission} from '@shared/utils/member';
 import _ from 'lodash';
 import {CookieService} from 'ngx-cookie-service';
 import {Subject} from 'rxjs';
-import {filter, map, switchMap, take, takeUntil, tap} from 'rxjs/operators';
+import {filter, switchMap, take, takeUntil, tap} from 'rxjs/operators';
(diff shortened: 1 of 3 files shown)


modules/web/src/app/project/component.ts
```diff
@@ -232,16 +232,25 @@ export class ProjectComponent implements OnInit, OnChanges, OnDestroy {
...
+    if (this.currentUser?.isAdmin) {
+      this._quotaService.quotas
+        .pipe(take(1))
+        .pipe(filter(quotas => !!quotas?.find(quotaDetails => !_.isEmpty(quotaDetails?.quota))))
+        .subscribe(_ => {
+          this.hasQuota = true;
+          this._cdr.detectChanges();
+        });
+    } else {
+      for (const project of projects) {
```
ahmedwaleedmalik, line 244: We should check if the project is active otherwise there is no point in trying to retrieve the quotas. At least let's do it here since we are already fixing things.

ahmedwaleedmalik (approved):
/approve

PR #5726, opened 2023-02-13
Keep WS connection alive by exchange of messages

KhizerRehan:
This PR keeps WS connection alive by exchange of messages from `be->fe ping` and `fe->be pong` to let backend . In order to avoid session get timeout.

c7a34fb0526c5a24006bcd51f402ebc9

Fixes #
Handle changes for
[PR#5696](https://github.com/kubermatic/dashboard/pull/5696)
[PR#5727](https://github.com/kubermatic/dashboard/pull/5727)

**What type of PR is this?**
/kind bug

Previously WS connection was lost due to middleware proxy layer could be nginx or any reverse proxy so in order to let client/server that communication is still alive we need to keep web-terminal connection to alive

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #5727, opened 2023-02-13
Handle "pong" messages from web terminal client

### LINKED ISSUE #5683: Webterminal time-out after 1 min of inactivity
### What happened
Webterminal time-outs after 1 min of inactivity in some environments.

In https://github.com/kubermatic/kubermatic/pull/11756 we solved it at the KKP ingress level, but the same issue can happen at different levels within the cluster's infrastructure (e.g. load-balancer, proxy, etc).

It would be much better to send periodic (e.g. 30s) ping-pong messages via the websocket connection to keep it up.

### Expected behavior

Webterminal connection should keep up until the session is expired (30 mins)

### Environment
- UI Version: v2.22.0-alpha.0
- API Version: v2.22.0-alpha.0
- Cloud provider: AWS

hbernardo:
This PR adds a missing part for the web terminal bug fix for inactivity timeout: now it is able to handle the "ping-pong" reply from the client.

Fixes #5683

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/handler/websocket/terminal.go
+++ b/modules/api/pkg/handler/websocket/terminal.go
@@ -63,6 +63,7 @@ const (
 	expirationRefreshesKey            = "ExpirationRefreshes"
 	pingInterval                      = 30 * time.Second
 	pingMessage                       = "PING"
+	pongMessage                       = "PONG"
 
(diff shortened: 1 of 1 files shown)


ahmedwaleedmalik (approved):
/approve

PR #5733, opened 2023-02-14
Fix IP ranges pattern error message + labels

rastislavs:

Single IP address cannot be configured in the "Allowed IP Ranges for apiserver". It is correctly guarded by a pattern, but the error message "Input has to be a valid IPv4 or IPv6 address" is wrong - it fact it cannot be an IP address, but must be an IP address CIDR:

image

This PR fixes it + makes "IP Range" plural "IP Ranges"

Fixes #

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #5738, opened 2023-02-14
5728 cluster patch skip skew validation

### LINKED ISSUE #5728: Cluster patch endpoint is dependent on user cluster API server access
### What happened

When editing a cluster with restricted access to apiserver (via `apiServerAllowedIPRanges`), it is not possible to edit a cluster with too restrictive policy, preventing kubermatic-apiserver talking to user cluster's k8s apiserver

e.g. PATCH operation to the endpoint `https://<kkp-domain>/api/v2/projects/<project>/clusters/<cluster>`) fails with:

```
{"error":{"code":500,"message":"failed to check existing nodes' version skew: failed to create a machine client: failed to create restMapper: Get \"https://<redacted>/api?timeout=32s\": dial tcp <redacted>: i/o timeout"}}
```

That can easily happen if a user is editing `apiServerAllowedIPRanges`, adds too restrictive configuration, and later realizes that it needs to be extended - but it may not be possible from the UI anymore.

### Expected behavior

Users should be able to edit the cluster (e.g. via "Edit Cluster" dialog) even if access to the cluster's apiserver is restricted.

### How to reproduce

Configure `apiServerAllowedIPRanges` with a too restrictive policy (e.g. `1.2.3.4/32`) - the cluster will not be editable from the UI anymore.

### Environment
- UI Version: v2.22.0-alpha.1
- API Version: v2.22.0-alpha.1

pkprzekwas:
Checking kubelet versions on user cluster machines requires network connection between kubermatic-api and user cluster api-server. In case where the connection is blocked, we still want to be able to send a patch request. This can be achieved with an additional query param attached to the patch request - "?skip_cluster_version_skew_check=true"

Fixes #5728 

**What type of PR is this?**
/kind bug

```release-note
Add `skip_cluster_version_skew_check` query param to `api/v2/projects/<project>/clusters/<cluster>` PATCH to enable switching off kubelet version validation.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/core/services/cluster.ts
+++ b/modules/web/src/app/core/services/cluster.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {HttpClient, HttpHeaders} from '@angular/common/http';
+import {HttpClient, HttpHeaders, HttpParams} from '@angular/common/http';
 import {Injectable} from '@angular/core';
 import {AppConfigService} from '@app/config.service';
 
@@ -50,6 +50,7 @@ export class ClusterService {
   private _restRoot: string = environment.restRoot;
   private _newRestRoot: string = environment.newRestRoot;
   private _headers: HttpHeaders = new HttpHeaders();
+  private _params: HttpParams = new HttpParams();
   private _clusters$ = new Map<string, Observable<Cluster[]>>();
   private _externalClusters$ = new Map<string, Observable<ExternalCluster[]>>();
   private _cluster$ = new Map<string, Observable<Cluster>>();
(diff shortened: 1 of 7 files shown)


modules/web/src/app/core/services/cluster.ts
```diff
@@ -137,9 +138,12 @@ export class ClusterService {
     return this._http.post<Cluster>(url, createClusterModel);
   }
 
-  patch(projectID: string, clusterID: string, patch: ClusterPatch): Observable<Cluster> {
+  patch(projectID: string, clusterID: string, patch: ClusterPatch, skipKubeletValidation = false): Observable<Cluster> {
     const url = `${this._newRestRoot}/projects/${projectID}/clusters/${clusterID}`;
-    return this._http.patch<Cluster>(url, patch);
+    if (skipKubeletValidation) {
+      this._params = this._params.set('skip_kubelet_version_validation', skipKubeletValidation);
```
Waseem826, line 144: I think `params` should be a local variable in this method, unless we want to save this value for future requests which don't require `skipKubeletValidation`?

pkprzekwas: Yeah, that makes more sense. I was inspired by the way this class manages headers, but keeping it in the method's scope is a good idea.

ahmedwaleedmalik (approved):
/approve

PR #5751, opened 2023-02-16
Fix quota calculation for Kubevirt resources

### LINKED ISSUE #5735: Live Quota calculation issues
### What happened
Live quota calculation on backend side

**Issue 1:**

- Found while testing with kubevirt 

Issue related to irresepective of whatever replica count is set the storage count is not updatd

Request
```
{
    "replicas": 8,
    "kubevirtNodeSize": {
      	...
    },
    "replacedResources": {
        ....
        }
    }
}
```
Response e.g:

```
{
    "cpu": xx,
    "memory": xx, <calculation is not correctly calculated based on replaced resource>
    "storage": yy <remains fixed irrespective replicas count is changed>
}
```

Issue-2: Edit MD dialog

In case of **Kubevirt** when Instance type is not selected and CPU/Memory is provided manually from user input user at the time of creation but in case of edit MD dialog user selects Instance type however calculation in that case to be handled correctly which IMO isn't doing correctly now.

### Environment

UI Version: v2.22.0-alpha.1
API Version: v2.22.0-alpha.1

pkprzekwas:
This PR attempts to resolve two issues:
- missing storage calculations for Kubevirt
- memory mismatch between calculated and requested quota

Solution for missing storage is quite simple. The problem was missing unit, so I added it both on frontend's form as well as extended backend with some defaulting in case a value is missing.

The memory mismatch is a bit more tricky... This is related to the conversion between BinarySI and DecimalSI units - [more info here](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory). 

The problem is that API returns memory for Kubevirt instances in `Gi` unit and we represent our quotas in `G`. It results in a following:
Screenshot 2023-02-16 at 16 11 01

To fix that, I've added some conversion to API and switched UI form to send values in `M` instead of `Mi`.

Fixes #5735 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

I have tested the **API** changes locally and they look good. Great job @pkprzekwas, now we just need to fix the tests on this PR 😄

modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
```diff
@@ -186,7 +186,7 @@ export class KubeVirtBasicNodeDataComponent
       [Controls.Memory]: this._builder.control(this._defaultMemory, Validators.required),
       [Controls.PrimaryDiskOSImage]: this._builder.control('', Validators.required),
       [Controls.PrimaryDiskStorageClassName]: this._builder.control('', Validators.required),
-      [Controls.PrimaryDiskSize]: this._builder.control('10', Validators.required),
+      [Controls.PrimaryDiskSize]: this._builder.control('10G', Validators.required),
```
Waseem826, line 189: We use `km-number-stepper` for this so I think `10G` value might break it. Can you double check that it works?

pkprzekwas: Tested that locally and it seems to work fine.

Waseem826 (approved):

PR #5754, opened 2023-02-16
Bump version to v2.22.0-rc.1

embik:

Just preparing this PR in case we want to cut a second RC, which seems likely at the moment.

Fixes #

**What type of PR is this?**
/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


embik:
/hold

Since it's not fully clear if we want to do this. I just want CI to be on our side already if we go for it.

PR #5759, opened 2023-02-17
Add changelog.json for end user facing changes in KKP 2.22.0

embik:

I've extracted what I considered the most important end user (i.e. someone who uses the dashboard of a KKP instance) facing changes and updated the `changelog.json` with it. I assume it does not cover everything, but unless I've missed something crucial it's probably a solid amount of information for users. The UI links to the full changelog anyway as well.

Fixes #

**What type of PR is this?**
/kind documentation

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/assets/config/changelog.json
+++ b/modules/web/src/assets/config/changelog.json
@@ -57,6 +57,14 @@
         "category": "added",
         "description": "Add Cilium CNI v1.13.0."
     },
+    {
+        "category": "added",
+        "description": "EE only: Administrators can enforce default project quotas.",
+    },
+    {
+        "category": "added",
+        "description": "EE only: If project quotas are enabled, quota changes are live estimated while creating clusters and machine deployments.",
+    },
 
     {
         "category": "changed",
@@ -69,6 +77,10 @@
     {
         "category": "changed",
         "description": "By default, external CCM is turned on for all providers that support it."
+    },
+    {
+        "category": "changed",
+        "description": "Cluster templates can be edited and customized during cluster creation."
     },
     {
         "category": "changed",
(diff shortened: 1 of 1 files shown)


modules/web/src/assets/config/changelog.json
```diff
@@ -1,4 +1,185 @@
...
+    },
+    {
+        "category": "added",
+        "description": "Add Canal CNI v3.24."
+    },
+    {
+        "category": "added",
+        "description": "Add Cilium CNI v1.13.0."
+    },
+
```
pkprzekwas, line 60: Are you sure those empty lines don't affect JSON parsing?

embik: I tried it by running the dashboard locally and it was fine, I was able to see the full changelog via it.

Waseem826:
I think we've missed following features:
- Default Project Quota
- Live estimation of Project Quota while creating cluster/MD.
- vSphere tag categories
- Cluster RBAC changes (maybe)
- Edit/Customize cluster templates
- Default application installation when Cilium is selected as CNI plugin

embik:
> I think we've missed following features:
> 
> * Default Project Quota
> * Live estimation of Project Quota while creating cluster/MD.

Added that, but I suspect we should look into having changelog items only showing up for EE installations.

> * vSphere tag categories

I'm not sure what to write here to be honest since I don't fully understand the changes.

> * Cluster RBAC changes (maybe)

Not sure here either, sorry.

> * Edit/Customize cluster templates

Added.

> * Default application installation when Cilium is selected as CNI plugin

I think this is more of an internal change. It's still on the changelog, but I'm not sure it should be on the abridged version.

Waseem826 (approved):

PR #5766, opened 2023-02-21
Extend swagger doc mentioning API as being internal

pkprzekwas:
What the title says.

Fixes #

**What type of PR is this?**
/kind documentation

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/cmd/kubermatic-api/main.go
+++ b/modules/api/cmd/kubermatic-api/main.go
@@ -17,6 +17,7 @@ limitations under the License.
 // Kubermatic Kubernetes Platform API
 //
 // This spec describes possible operations which can be made against the Kubermatic Kubernetes Platform API.
+// The API may introduce breaking changes and is not meant for external integrations.
 //
 //	Schemes: https
 //	Version: 2.22
--- a/modules/api/cmd/kubermatic-api/main.go
+++ b/modules/api/cmd/kubermatic-api/main.go
@@ -14,10 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Kubermatic Kubernetes Platform API
+// Package docs Kubermatic Kubernetes Platform API
 //
-// This spec describes possible operations which can be made against the Kubermatic Kubernetes Platform API.
-// The API may introduce breaking changes and is not meant for external integrations.
+// This OpenAPI 2.0 specification describes the REST APIs used by the Kubermatic Kubernetes Platform Dashboard.
+//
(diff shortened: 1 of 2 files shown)


modules/api/cmd/kubermatic-api/main.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 // Kubermatic Kubernetes Platform API
 //
 // This spec describes possible operations which can be made against the Kubermatic Kubernetes Platform API.
+// The API may introduce breaking changes and is not meant for external integrations.
```
pkprzekwas, line 20: An alternative wording that came to my mind:
```
The API should be considered private to the Dashboard and may introduce breaking changes.
```

ahmedwaleedmalik, line 20:
```suggestion
// Package docs Kubermatic Kubernetes Platform API
//
// This OpenAPI 2.0 specification describes the REST APIs used by the Kubermatic Kubernetes Platform Dashboard.
//
// KKP API is an **internal API** that is built primarily to complement the KKP dashboard. Our users are strongly advised to use Kubernetes APIs i.e. [KKP CRDs](https://docs.kubermatic.com/kubermatic/v2.22/references/crds/) to interact with and integrate KKP into their own platforms.
//
// Since KKP API is **not a public API**, users can expect breaking changes without any formal announcements.
```

ahmedwaleedmalik (changes requested):

modules/api/cmd/kubermatic-api/main.go
ahmedwaleedmalik: @pkprzekwas I was thinking of something like this. More detailed instead of just a one-liner. WDYT?

pkprzekwas: Now I feel stupid with my oneliner 😅 . But this is exactly what we need! Thanks @ahmedwaleedmalik , let me update swagger.

modules/api/cmd/kubermatic-api/main.go
```diff
@@ -14,9 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Kubermatic Kubernetes Platform API
+// Package docs Kubermatic Kubernetes Platform API
 //
-// This spec describes possible operations which can be made against the Kubermatic Kubernetes Platform API.
+// This OpenAPI 2.0 specification describes the REST APIs used by the Kubermatic Kubernetes Platform Dashboard.
+//
+// KKP API is an **internal API** that is built primarily to complement the KKP dashboard. Our users are strongly advised to use Kubernetes APIs i.e. [KKP CRDs](https://docs.kubermatic.com/kubermatic/v2.22/references/crds/) to interact with and integrate KKP into their own platforms.
```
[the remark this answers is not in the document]
ahmedwaleedmalik: We can update the release procedure here later https://github.com/kubermatic/dashboard/blob/main/.github/ISSUE_TEMPLATE/release.md

PR #5783, opened 2023-02-28
Add a Button to Allow Hide/Show of Sensitive Information

### LINKED ISSUE #5501: Allow unhiding of passwords/sensitive information
from the sig ui discussion: we need to collect all the places it’s relevant, search for locations we already have asterisks/hiding functionality, low priority

ahmadhamzh:
Add a Button to Allow Hide/Show input value in all places where we have Sensitive Information

image

Fixes #5501 

**What type of PR is this?**
/kind design

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/shared/directives/input-password.ts
+++ b/modules/web/src/app/shared/directives/input-password.ts
@@ -1,0 +1,56 @@
+// Copyright 2023 The Kubermatic Kubernetes Platform contributors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import {Directive, ElementRef, OnDestroy, OnInit} from '@angular/core';
+
+@Directive({
+  selector: '[kmInputPassword]',
+})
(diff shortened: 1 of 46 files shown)


ahmedwaleedmalik:
@ahmadhamzh can you please add a gif/screenshot for how this looks like?

Waseem826:
@ahmadhamzh wouldn't it be better to create a separate component for password fields instead of creating a directive and then adding mask button programmatically?

Waseem826:
As far as custom hints and error messages are concerned, we can project them from outside using `ng-content` similar to how we do in `km-autocomplete` and `km-chip-list` components. And I think it would be better to create a component instead of directive because directives are generally used for changing some behaviour of host element but here we're always adding a new element as sibling of host element.

ahmedwaleedmalik:
> directives are generally used for changing some behaviour of host element but here we're always adding a new element as sibling of host element

Which is something completely fine to do with directives. Basically, this use case, show/hide passwords, is mostly solved using directives. For me, I'm divided between using a directive or a dedicated component but I'd lean more towards using a directive.

Waseem826:
@ahmadhamzh upon some rethinking, its fine to use directive approach.

modules/web/src/app/shared/directives/input-password.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  addButton() {
+    const input = this._el.nativeElement;
+    input.type = 'password';
+    const button = document.createElement('button');
+    button.classList.add('mat-icon-password-show-button', 'matSuffix');
+    const icon = document.createElement('i');
+    icon.classList.add('km-icon-mask', this._showPassword ? 'km-icon-show' : 'km-icon-hide');
+    button.appendChild(icon);
+    button.addEventListener('click', event => {
```
Waseem826, line 37: Please make sure to remove this event listener in `ngOnDestroy`

Waseem826 (approved):

PR #5785, opened 2023-03-02
WIP Fix api model changes for Share cluster flag

KhizerRehan:

Fixes #

**What type of PR is this?**

```release-note

```

**Documentation**:

```documentation

```


KhizerRehan:
```
func (s *SettingsProvider) UpdateGlobalSettings(ctx context.Context, userInfo *provider.UserInfo, settings 
 
 // Till here i have debugged the value is correct patched

// But as soon after runnign `s.runtimeClient.Update` the value gets modified by lib and 
// i checked with patryk any  further investigation or help would be appreciated.

	if err := s.runtimeClient.Update(ctx, settings); err != nil {
		return nil, err
	}
	return settings, nil
}

cc: @ahmedwaleedmalik

ahmedwaleedmalik:
I'll create a separate PR with my changes.

PR #5861, opened 2023-03-28
Configure OpenStack ingress cluster settings

### LINKED ISSUE #5856: Newish OpenStack Cluster Resource Fields Should Be Settable in UI

In KKP 2.21 we introduced the fields `enableIngressHostname` and `ingressHostnameSuffix` (see https://docs.kubermatic.com/kubermatic/v2.21/references/crds/#openstackcloudspec). Even though they are being used as a workaround, it makes sense to be able to set them in the UI during cluster creation. Additionally from my understanding these are fields that can also be set after a cluster is already running, so it would make sense to allow for setting these fields in the cluster edit screen as well.

As we introduce these fields in 2.21, I think we should also backport this change to 2.21 and 2.22 of course.

Waseem826:
Allow configuration of `enableIngressHostname` and `ingressHostnameSuffix` settings of OpenStack provider.

screenshot-localhost_8000-2023 03 28-22_22_21

Additionally 2 more issues are fixed:
- Load security group, network and subnet pool dropdown options only when cluster credentials are updated.
- Visually mark Disk Size field (on Initial Nodes step) as required when form control is marked as required.

Fixes #5856 

**What type of PR is this?**

/kind feature

```release-note
Configure Ingress Hostname cluster settings of OpenStack provider.
```

**Documentation**:

```documentation
https://docs.kubermatic.com/kubermatic/main/references/crds/#openstackcloudspec
```

### DIFF
--- a/modules/web/src/app/node-data/basic/provider/openstack/component.ts
+++ b/modules/web/src/app/node-data/basic/provider/openstack/component.ts
@@ -242,6 +242,10 @@ export class OpenstackBasicNodeDataComponent extends BaseFormValidator implement
     return this.form.get(Controls.UseFloatingIP).disabled;
   }
 
+  isDiskSizeRequired(): boolean {
+    return this.form.get(Controls.CustomDiskSize).hasValidator(Validators.required);
+  }
+
   onFlavorChange(flavor: string): void {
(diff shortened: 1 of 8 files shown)


ahmedwaleedmalik (approved):
/approve

PR #5868, opened 2023-03-30
Mutate partial cluster before validating its spec

embik:
This makes use of the new `k8c.io/kubermatic/v2/pkg/mutation/cluster` package and mutates partial clusters generated when creating or updating a user cluster via the Kubermatic API.

These mutations need to be applied because the API also calls validation logic. This validation logic expects mutations have happened (e.g. the externalCCM feature flag missing can fail version validation).

The particular issue this is fixing is that e.g. vSphere clusters cannot be created for 1.25 and higher via the API/dashboard because the API valIdation call fails. Correct version valIdation was added in KKP validation code in https://github.com/kubermatic/kubermatic/pull/11996, so this issue was uncovered by fixing validation logic.

Fixes #

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


pkprzekwas:
/lgtm

Leaving approval for @ahmedwaleedmalik as he was reviewing backend changes.

ahmedwaleedmalik (approved):
/approve

PR #5878, opened 2023-04-03
Increase dependabot open pull requests limit

Waseem826:
We have ~80 web packages and weekly limit of `5` dependabot PRs is too low to cover these number of packages. Also, Angular packages are bumped together so often that limit is filled out by all those updates and it causes other packages to lag behind which sometimes also causes CVEs to miss out.

Fixes #

**What type of PR is this?**

/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,7 +18,7 @@ updates:
     directory: "modules/web/"
     schedule:
       interval: weekly
-    open-pull-requests-limit: 5
+    open-pull-requests-limit: 15
     labels:
       - dependencies
       - docs/none


ahmedwaleedmalik:
`open-pull-requests-limit` is used to determine PR limits for package updates. Even if we set that value to 0, we'll still receive security updates. 

In general, we'd like to rely on manual package updates using tools like [npm-check-updates](https://www.npmjs.com/package/npm-check-updates). They are more manageable and pragmatic for our use cases and team setup.

cc @ahmadhamzh so that you are aware of this.

Waseem826:
@ahmedwaleedmalik I agree and in that sense we can keep the current limit.

PR #5879, opened 2023-04-04
Fix 'v' prefix removal by removing character instead of slicing string

### LINKED ISSUE #5871: Invalid semantic version error after applying KubermaticConfiguration
### What happened
One of our customers reported that after configuring versions in KubermaticConfiguration CRD, it is not able to create a cluster in the dashboard with an error being reported: Invalid Semantic Version

Info from customer
> When I try to create a cluster in the dashboard than I get the following error: Invalid Semantic Version
No Control Plane Version is preselected but I can still choose between the configured versions. 

### Expected behavior
It should not fail on these versions and the cluster should be created.

### How to reproduce

Apply KubermaticConfiguration.yaml based on the defaults: https://github.com/kubermatic/kubermatic/blob/release/v2.22/docs/zz_generated.kubermaticConfiguration.ee.yaml
```bash
...
  versions:
    default: v1.25.6
    updates:
      - # Automatic controls whether this update is executed automatically
        # for the control plane of all matching user clusters.
        automatic: true
        # Automatic controls whether this update is executed automatically
        # for the worker nodes of all matching user clusters.
        automaticNodeUpdate: false
        # From is the version from which an update is allowed. Wildcards are allowed, e.g. "1.18.*".
        from: 1.23.*
        # To is the version to which an update is allowed.
        # Must be a valid version if `automatic` is set to true, e.g. "1.20.13".
        # Can be a wildcard otherwise, e.g. "1.20.*".
        to: 1.24.10
      - from: 1.24.*
        to: 1.24.*
      - automatic: true
        from: '>= 1.24.0, < 1.24.8'
        to: 1.24.8
      - from: 1.24.*
        to: 1.25.*
      - from: 1.25.*
        to: 1.25.*
      - automatic: true
        from: '>= 1.25.0, < 1.25.4'
        to: 1.25.4
      - from: 1.25.*
        to: 1.26.*
      - from: 1.26.*
        to: 1.26.*
    versions:
      - v1.24.3
      - v1.24.6
      - v1.24.8
      - v1.24.9
      - v1.24.10
      - v1.25.2
      - v1.25.4
      - v1.25.5
      - v1.25.6
      - v1.26.1
```
Verify that on the cluster 'v' prefixes got removed from versions
```bash
  versions:
    default: 1.25.6
    updates:
    - automatic: true
      automaticNodeUpdate: false
      from: 1.23.*
      to: 1.24.10
    - from: 1.24.*
      to: 1.24.*
    - automatic: true
      from: '>= 1.24.0, < 1.24.8'
      to: 1.24.8
    - from: 1.24.*
      to: 1.25.*
    - from: 1.25.*
      to: 1.25.*
    - automatic: true
      from: '>= 1.25.0, < 1.25.4'
      to: 1.25.4
    - from: 1.25.*
      to: 1.26.*
    - from: 1.26.*
      to: 1.26.*
    versions:
    - 1.24.3
    - 1.24.6
    - 1.24.8
    - 1.24.9
    - 1.24.10
    - 1.25.2
    - 1.25.4
    - 1.25.5
    - 1.25.6
    - 1.26.1
 ```
 Try to create a cluster via the KKP dashboard

### Environment
- UI Version:  2.22
- API Version: 2.22

### PR DESCRIPTION
**Pull request by embik**
The frontend code made the assumption that there will always be a `v` prefix for versions. That is however not the case, for example when you configure versions on your own in `KubermaticConfiguration`. This PR makes sure we remove a `v` string from the version instead, so it doesn't do anything if the version is not set.

Technically, versions could have a string suffix that could include `v`, but this is not the case for Kubernetes versions and I couldn't get a regex to work here. So if you're from the future stumbling into this PR, I'm sorry I made an incorrect assumption. Otherwise, feel free to discard this comment.

Fixes #5871

**What type of PR is this?**
/kind bug

```release-note
Fix cluster wizard not selecting a default version if custom versions are configured in `KubermaticConfiguration`
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/wizard/step/cluster/component.ts
+++ b/modules/web/src/app/wizard/step/cluster/component.ts
@@ -539,7 +539,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
           [Controls.Name]: this._clusterSpecService?.cluster?.name ?? this.controlValue(Controls.Name),
           // remove the 'v' prefix from the version
           [Controls.Version]:
-            (clusterSpec?.version && clusterSpec?.version.slice(1)) ?? this.controlValue(Controls.Version),
+            (clusterSpec?.version && clusterSpec?.version.replace('v','')) ?? this.controlValue(Controls.Version),
           [Controls.ContainerRuntime]: clusterSpec?.containerRuntime ?? this.controlValue(Controls.ContainerRuntime),
           [Controls.AuditLogging]: clusterSpec?.auditLogging?.enabled ?? this.controlValue(Controls.AuditLogging),
           [Controls.AuditPolicyPreset]:
--- a/modules/web/src/app/wizard/step/cluster/component.ts
+++ b/modules/web/src/app/wizard/step/cluster/component.ts
@@ -539,7 +539,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
           [Controls.Name]: this._clusterSpecService?.cluster?.name ?? this.controlValue(Controls.Name),
           // remove the 'v' prefix from the version
           [Controls.Version]:
-            (clusterSpec?.version && clusterSpec?.version.replace('v','')) ?? this.controlValue(Controls.Version),
+            (clusterSpec?.version && clusterSpec?.version.replace('v', '')) ?? this.controlValue(Controls.Version),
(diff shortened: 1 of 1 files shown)


Waseem826 (approved):
/approve

PR #5920, opened 2023-04-25
Hide enable kubernetes dashboard on cluster creation

### LINKED ISSUE #5904: Kubernetes Dashboard still visible on cluster creation after disable in the admin panel
### What happened

The K8s dashboard is disable in the admin panel

screenshot_63

but still visible during cluster creation and editing 

screenshot_65
screenshot_64

### Expected behavior

The buttons are not visible or not selectable 

### Environment
- UI Version: current Dev

ahmadhamzh:
Disable `Kubernetes Dashboard` option on cluster creation in the cluster step if `Enable Kubernetes Dashboard` is disable in the admin panel.
image

Fixes #5904 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/wizard/step/cluster/component.ts
+++ b/modules/web/src/app/wizard/step/cluster/component.ts
@@ -163,6 +163,10 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
   private readonly _canalDualStackMinimumSupportedVersion = '3.22.0';
   private readonly _cniInitialValuesMinimumSupportedVersion = '1.13.0';
 
+  get isKubernetesDashboardEnable(): boolean {
+    return this._settings.enableDashboard;
+  }
+
   constructor(
     private readonly _builder: FormBuilder,
     private readonly _cdr: ChangeDetectorRef,
@@ -201,6 +205,9 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       this.form.get(Controls.OPAIntegration).setValue(this._settings.opaOptions.enabled);
       if (this._settings.opaOptions.enforced) {
         this.form.get(Controls.OPAIntegration).disable();
+      }
+      if (!settings.enableDashboard) {
+        this.form.get(Controls.KubernetesDashboardEnabled).setValue(false);
       }
       this.form.updateValueAndValidity();
     });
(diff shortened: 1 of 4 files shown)


modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -202,6 +206,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       if (this._settings.opaOptions.enforced) {
         this.form.get(Controls.OPAIntegration).disable();
       }
+      this.form.get(Controls.KubernetesDashboardEnabled).setValue(this._settings.enableDashboard);
```
Waseem826, line 209: I think we shouldn't enable it by default (if its enabled in admin settings). Wdyt?

Waseem826 (changes requested):
Shouldn't we do the same in Edit Cluster dialog as well?

modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -465,7 +469,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       [Controls.OPAIntegration]: this._builder.control(clusterSpec?.opaIntegration?.enabled ?? false),
       [Controls.Konnectivity]: this._builder.control(clusterSpec?.clusterNetwork?.konnectivityEnabled ?? true),
       [Controls.MLALogging]: this._builder.control(clusterSpec?.mla?.loggingEnabled ?? false),
-      [Controls.KubernetesDashboardEnabled]: this._builder.control(clusterSpec?.kubernetesDashboard?.enabled ?? true),
+      [Controls.KubernetesDashboardEnabled]: this._builder.control(clusterSpec?.kubernetesDashboard?.enabled ?? false),
```
Waseem826, line 472: Sorry I might not be clear in my previous comment but what I meant to say was that if dashboard is enabled in admin settings then we shouldn't enable it forcefully. However, we should keep this default `true` value here.
So here are different scenarios:
- Default value should be `true` here like before.
- If dashboard is **disabled** in Admin Settings then we should set `KubernetesDashboardEnabled` control value to `false` (like you did in previous changes but add an `if` condition).
- Inside the `_loadClusterDefaults` method, we should check value of `isKubernetesDashboardEnable` first:
 ```
[Controls.KubernetesDashboardEnabled]:
            (this.isKubernetesDashboardEnable && clusterSpec?.kubernetesDashboard?.enabled) ?? this.controlValue(Controls.KubernetesDashboardEnabled)

modules/web/src/app/wizard/step/cluster/component.ts
Waseem826: @ahmadhamzh please bring back this change but now add an `if` condition here so that we only set this value when dashboard is disabled in admin settings.

Waseem826 (approved):

PR #5927, opened 2023-04-28
Migrate Angular Material Dialog component from legacy to v15

### LINKED ISSUE #5921: Migrate Material Dialog component from legacy to v15
Migrate Angular Material's Dialog component from legacy to v15 by following instructions in https://github.com/angular/components/blob/main/guides/v15-mdc-migration.md#dialog

> - The .mat-dialog-container does not contain a 24px padding anymore. Instead, the inner dialog directives are responsible for adding the right padding. This will be apparent if your dialog does not use any of the directives like <mat-dialog-content>.
> - mat-dialog-content uses the font-settings specified by the Material Design spec, which includes a rather roomy line-height. If you have an information-dense dialog that doesn't look good with these settings, you can avoid using <mat-dialog-content> and just use a div with custom padding, or use custom typography settings that can be applied with the mat.mdc-dialog-typography mixin.
> - The old dialog triggered an extra change detection, which may have masked change detection issues in your application that need to be fixed when migrating.

Waseem826:
This PR migrates the Angular Material Dialog component from legacy to latest version which follows new MDC specs: https://m2.material.io/components/dialogs#specs.

As part of this migration there are some noticeable changes to the dialogs:
- Action buttons (including close dialog button) are moved closer to edges (`8px` distance from sides).
- There is some letter spacing in dialog title and content.
- CSS classes are prefixed with `mat-mdc-` as opposed to just `mat-` previously.
- Some more changes mentioned here: https://material.angular.io/guide/mdc-migration#dialog

**Old Dialog:**

screenshot-dev kubermatic io-2023 05 04-14_29_56

**Updated Dialog:**

screenshot-localhost_8000-2023 05 04-14_28_21

Fixes #5921 

**What type of PR is this?**

/kind-design

```release-note
Update Dialogs to follow latest material design specifications.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/core/components/changelog/template.html
+++ b/modules/web/src/app/core/components/changelog/template.html
@@ -15,15 +15,7 @@ limitations under the License.
 -->
 <div fxFlex
      fxLayout="column">
-  <div class="header-title">
-    <div>What's New</div>
-    <div fxFlex></div>
-    <div>
-      <i class="km-icon-close km-pointer"
-         aria-hidden="true"
-         mat-dialog-close></i>
-    </div>
-  </div>
+  <km-dialog-title>What's New</km-dialog-title>
   <div class="header-info">
     Version {{version}}
     <ng-container *ngIf="released"> - Released {{released | date:'MMM d, y'}}</ng-container>
--- a/modules/web/src/assets/css/theme/_main.scss
+++ b/modules/web/src/assets/css/theme/_main.scss
@@ -26,9 +26,15 @@
 //  If you specify typography styles for the components you use elsewhere, you should delete this line.
 //  If you don't need the default component typographies but still want the hierarchy styles,
 //  you can delete this line and instead use:
-//    `@include mat.legacy-typography-hierarchy(mat.define-legacy-typography-config());`
+//    `@include mat.legacy-typography-hierarchy(mat.define-typography-config());`
+
+/* TODO(mdc-migration): Remove all-legacy-component-typographies once all legacy components are migrated */
 @include mat.all-legacy-component-typographies;
+@include mat.all-component-typographies;
+
+/* TODO(mdc-migration): Remove legacy-core once all legacy components are migrated */
 @include mat.legacy-core;
+@include mat.core;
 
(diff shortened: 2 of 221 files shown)


modules/web/src/app/core/components/changelog/template.html
```diff
@@ -15,15 +15,7 @@
...
-  <div class="header-title">
-    <div>What's New</div>
-    <div fxFlex></div>
-    <div>
-      <i class="km-icon-close km-pointer"
-         aria-hidden="true"
-         mat-dialog-close></i>
-    </div>
-  </div>
+  <km-dialog-title>What's New</km-dialog-title>
```
ahmadhamzh, line 18: maybe we need to remove the `.header-title` from the style file since i think we don't use the class any where

modules/web/src/assets/css/theme/_main.scss
```diff
@@ -26,9 +26,15 @@
 //  If you specify typography styles for the components you use elsewhere, you should delete this line.
 //  If you don't need the default component typographies but still want the hierarchy styles,
 //  you can delete this line and instead use:
-//    `@include mat.legacy-typography-hierarchy(mat.define-legacy-typography-config());`
+//    `@include mat.legacy-typography-hierarchy(mat.define-typography-config());`
+
+/* TODO(mdc-migration): Remove all-legacy-component-typographies once all legacy components are migrated */
 @include mat.all-legacy-component-typographies;
+@include mat.all-component-typographies;
```
ahmadhamzh, line 33: do we need to remove `@include mat.all-legacy-component-typographies;`
and also the `TODO` comment lines ?

Waseem826: Yes, those are added by Angular Material's migration tool and I think these should remain here until we've migrated all components.

modules/web/src/app/core/components/changelog/template.html
Waseem826: You're right, I've removed it now.

PR #5940, opened 2023-05-04
Add option to clear VSphere tags category

### LINKED ISSUE #5815: Cannot clear vSphere "Provider Tags" on machine wizard step
### What happened

When trying to create a vSphere user cluster, the "Initial Nodes" step offers to select "Provider Tags" to be associated with the machines. However, selecting a category cannot be reverted / the field cannot be set to "empty" again. This is particularly problematic when you don't get offered any tags when selecting categories, so you cannot complete the setting to delete it afterwards again.

### Expected behavior

It should be possible to clear an incomplete "Provider Tags" selection.

### Environment
- UI Version: dev
- API Version: dev
- Domain: dev.kubermatic.io
- Others:

### Current workaround

Start creating the cluster from scratch.

Waseem826:
Add an option to clear VSphere tags category so it doesn't get stuck when there are no tags.
Moreover, this PR also fixes an issue which was displaying removed labels even when creating cluster or machine deployment.

screenshot-localhost_8000-2023 05 04-17_44_57

Fixes #5815 

**What type of PR is this?**

/kind bug

```release-note
Add an option to clear VSphere tags category so it doesn't get stuck when there are no tags.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/node-data/basic/provider/kubevirt/topology-spread-contraint-form/component.ts
+++ b/modules/web/src/app/node-data/basic/provider/kubevirt/topology-spread-contraint-form/component.ts
@@ -126,7 +126,9 @@ export class TopologySpreadConstraintFormComponent extends BaseFormValidator imp
   }
 
   deleteConstraint(index: number): void {
-    this.removedConstraints?.push(this.constraints[index]);
+    if (this._dialogModeService.isEditDialog) {
+      this.removedConstraints?.push(this.constraints[index]);
+    }
 
     this.constraintArray.removeAt(index);
     this._updateConstraints();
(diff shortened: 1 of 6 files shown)


ahmedwaleedmalik (approved):
/approve

embik:
@Waseem826 since this fixes a bug, can you maybe add a short release note so people looking for this bug being fixed see that we improved the UI? :)

ahmedwaleedmalik:
> since this fixes a bug

@embik as explained in the corresponding issue; this is not a bug.

embik:
I disagree it's not a bug. It's just a visual bug as you explained and it doesn't have an impact on the ability to create a cluster from this state, but I don't see how that doesn't make it a frontend bug.

ahmedwaleedmalik:
> It's just a visual bug as you explained

It isn't :) It's the way that component works as a default. We just "improved" its UX and functionality with this PR.

ahmedwaleedmalik:
@embik I have added the release note again.

@Waseem826 as discussed earlier, this PR is fine but we need a proper/generic solution at the component level to make sure that this behavior is the same for all usages of the components. Having behavioral drifts for components with similar functionalities is not the way to go. Let's discuss it tomorrow or in the next sig-api-ui meeting.

PR #5956, opened 2023-05-11
Clarify vSphere credentials

### LINKED ISSUE #5952: vSphere credentials split is unclear and needs better description

The step in the cluster creation wizard that asks for cloud provider settings (mostly credentials) for vSphere is super confusing to people not familiar with the implementation details. I had to look at the dashboard code to actually understand what was being set.

In the vSphere provider implementation, you can pass dedicated credentials for "general usage" (which effectively means CCM/CSI) and for infrastructure management (mostly resource provisioning in KKP and machine-controller).

Screenshot 2023-05-11 at 09 14 42

The problem in specific is that it's unclear what the "Use custom VSphere Cloud credentials" checkbox does. When I hear "cloud credentials" (knowing that we have "infrastructure management" credentials), I think those credentials are used to provision cloud resources. However, they are actually the CCM / CSI credentials. Nothing in the wizard step indicates this. 

The wizard behaves very different from the `Cluster` CRD, because there the _infrastructure management credentials_ are optional, not the CCM/CSI credentials. This adds to the confusion for someone knowing the CRD definitions.

Multiple options:

1. Switch things around and make the optional credentials the "infrastructure management" credentials, which means changing the label on the checkbox to e.g. "Use dedicated credentials to provision vSphere resources".
2. Clarify the existing label, either by changing it to e.g. "Use custom credentials for vSphere CCM and CSI" or by adding an info box to the checkbox that gives details on how the credentials are being used.

ahmadhamzh:
Change the checkbox label in Vspher settings(Credentials) step to a better description. 

Screenshot from 2023-05-11 16-56-44

Fixes #5952

**What type of PR is this?**
/kind design

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/wizard/step/provider-settings/provider/basic/vsphere/template.html
+++ b/modules/web/src/app/wizard/step/provider-settings/provider/basic/vsphere/template.html
@@ -44,9 +44,8 @@ limitations under the License.
       </mat-error>
     </mat-form-field>
     <mat-checkbox [formControlName]="Controls.UseCustomCloudCredentials"
-                  [name]="Controls.UseCustomCloudCredentials"
-                  matTooltip="If not selected, above credentials will be used for VSphere Cloud Provider.">
-      Use custom VSphere Cloud credentials
+                  [name]="Controls.UseCustomCloudCredentials">
+      Use dedicated credentials to provision vSphere resources
     </mat-checkbox>
   </div>
 


Waseem826:
@ahmadhamzh we should also check if some related documentation (mentioning this UI) needs to be updated as well.

Waseem826 (approved):

ahmadhamzh:
> @ahmadhamzh we should also check if some related documentation (mentioning this UI) needs to be updated as well.

i checked but i didn't see any documentation regarding this change

PR #5963, opened 2023-05-15
Fix update default backup destination

### LINKED ISSUE #5950: Cannot unset default backup destination from admin UI

So I'm not sure if this is supposed to work or not - Opening this as a feature request, but it might also be a bug. Please categorise appropriately.

Dev currently has a default backup destination configured. But it doesn't have valid credentials, so I wanted to go to the admin UI to disable it as default backup destination. This is what is happening:

https://github.com/kubermatic/dashboard/assets/10295525/cac74275-4328-4a56-9d61-6e281962db70

The message "created backup location" doesn't feel accurate since I'm editing an existing one, but the main problem is that unsetting the checkbox does not actually get applied.

I suspect it should either return an error (then this could be understood as a bug) or just unset the default backup destination, which is possible since https://github.com/kubermatic/kubermatic/pull/11594 (+ https://github.com/kubermatic/kubermatic/pull/11846).

ahmadhamzh:
handle update backup destination when the value is empty string (remove default backup destination)

Fixes #5950

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/bucket-settings/destinations/destination-dialog/component.ts
+++ b/modules/web/src/app/settings/admin/bucket-settings/destinations/destination-dialog/component.ts
@@ -140,7 +140,7 @@ export class DestinationDialog implements OnInit, OnDestroy {
   onNext(): void {
     this._matDialogRef.close();
     this._notificationService.success(
-      `${this.data.mode === Mode.Add ? 'Added' : ' Created'} the ${
+      `${this.data.mode === Mode.Add ? 'Added' : ' Updated'} the ${
         this.form.get(Controls.DestinationName).value
       } destination`
     );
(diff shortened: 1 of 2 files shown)


modules/web/src/app/settings/admin/bucket-settings/destinations/destination-dialog/component.ts
```diff
@@ -140,18 +140,14 @@ export class DestinationDialog implements OnInit, OnDestroy {
   onNext(): void {
     this._matDialogRef.close();
     this._notificationService.success(
-      `${this.data.mode === Mode.Add ? 'Added' : ' Created'} the ${
+      `${this.data.mode === Mode.Add ? 'Added' : ' Update'} the ${
```
Waseem826, line 143:
```suggestion
      `${this.data.mode === Mode.Add ? 'Added' : ' Updated'} the ${
```

ahmedwaleedmalik (approved):
/approve

PR #5987, opened 2023-05-26
Set Azure data disk size default value to 0

### LINKED ISSUE #5973: Azure Data Disk - default should be zero
Hi, Currently, when we define machine deployment for Azure - we provide default OS disk and data disk to be 30GB. This causes troubles to the clusters... 
1. as we end up spawining 30GB disk and never using it. 
2. External CSI driver gets confused as it cannot know about this disk and so it thinks that one more disk can be attached to the machine and causes pod startup issues.

1. Best suggestion is to not provide any possibility to specify this data disk for Azure altogether. This data disk is NEVER usable via Kubernetes automatically.
2. If there is some valid reason to keep this data disk field - default it to `0`.
3. This change is needed in 2 places. a) KKP new cluster wizard, b) New Machine Deployment popup.

Waseem826:
Set Azure data disk size default value to `0`.

screenshot-localhost_8000-2023 05 26-17_03_55

Fixes #5973 

**What type of PR is this?**

/kind feature

```release-note
NONE
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/modules/web/src/app/node-data/basic/provider/azure/component.ts
+++ b/modules/web/src/app/node-data/basic/provider/azure/component.ts
@@ -79,7 +79,8 @@ enum ZoneState {
 })
 export class AzureBasicNodeDataComponent extends BaseFormValidator implements OnInit, OnDestroy {
   private _sizeChanges = new EventEmitter<boolean>();
-  private readonly _defaultDiskSize = 30;
+  private readonly _defaultOSDiskSize = 30;
+  private readonly _defaultDataDiskSize = 0;
 
   readonly Controls = Controls;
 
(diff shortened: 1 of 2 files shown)


ahmadhamzh:
i think  maybe we can remove the required error from Data Disk Size field , im not sure do we need it to be 0 or just empty ,
also but not related to you changes we need to set min limit for the field it seems like you can add numbers less than 0

Waseem826:
@ahmadhamzh PR updated.

embik:
@Waseem826 I think this should have a release note since it's a user-facing change.

PR #6001, opened 2023-05-31
Ignore notification outside click if target element is toggle button

Waseem826:
There was an issue with notification panel where it would immediately close the panel because of outside click handler hence making it look like that click event is ignored. This PR fixes it by ignoring the outside click event if target element is notification toggle button.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #6049, opened 2023-06-16
Prevent calling availability zone api request if assign availability set

ahmadhamzh:

there is an issue that in azure md dialog (add/edit), the availability zones still show `zone loading` label even though the assign availability set is not enabled and it keep like that for more than 30 seconds cause the label wait for the api response, so we shouldn't even do that api request if the field already disabled .

Screenshot from 2023-06-16 13-09-53

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

Waseem826 (approved):

PR #6069, opened 2023-07-06
 Support for configuring multiple networks for vSphere

### LINKED ISSUE #6065: Support for configuring multiple networks for vSphere machines

As a KKP user, I'd like the support to add multiple networks to the VMs on vSphere. This is required when internal services are exposed on a dedicated network etc.

ahmedwaleedmalik:
This PR adds support for configuring multiple networks for vSphere. We don't allow users to update `vmNetName` for the clusters and machine deployments. Similarly presets can't be updated from UI as well. Hence, for new clusters and presets, we'll now simply use `networks` instead of `vmNetName` for vSphere.

Fixes #6065

**What type of PR is this?**

/kind feature

```release-note
Support for configuring multiple networks for vSphere.
```

**Documentation**:

```documentation
NONE
```


modules/web/src/app/settings/admin/presets/dialog/steps/settings/provider/vsphere/component.ts
```diff
@@ -78,11 +80,16 @@ export class VSphereSettingsComponent extends BaseFormValidator implements OnIni
...
+    this.networks = networks;
+    this.form.get(Controls.Networks).updateValueAndValidity();
+  }
+
   private _update(): void {
     this._presetDialogService.preset.spec.vsphere = {
       username: this.form.get(Controls.Username).value,
       password: this.form.get(Controls.Password).value,
-      vmNetName: this.form.get(Controls.VMNetName).value,
+      networks: this.form.get(Controls.Networks).value,
```
Waseem826, line 92:
```suggestion
      networks: this.form.get(Controls.Networks).value?.tags,
```

ahmedwaleedmalik: Do we really need `tags` here?

Waseem826: Yes, because `km-chip-list` has internal form which has `tags` form control so in order to get the actual value we need to do this. Otherwise, we get `tags` object in request payload.

ahmedwaleedmalik: > networks: this.form.get(Controls.Networks).value?.tags,

Yes, we do Waleed 🤦

Waseem826 (approved):

PR #6070, opened 2023-07-06
add the option to hide os from admin settings

ahmadhamzh:

Fixes #

**What type of PR is this?**

```release-note

```

**Documentation**:

```documentation

```


ahmedwaleedmalik:
Can you update the release note to something like:

```
Support for enabling/disabling operating systems for machines of user clusters
```

ahmedwaleedmalik (approved):
The API part looks fine. @Waseem826 can review the UI parts for this one.

modules/web/src/app/node-data/template.html
```diff
@@ -83,37 +83,40 @@
...
         </mat-button-toggle>
         <mat-button-toggle [value]="OperatingSystem.RockyLinux"
-                           *ngIf="isOperatingSystemSupported(OperatingSystem.RockyLinux)">
+                           *ngIf="isOperatingSystemAllowed(OperatingSystem.RockyLinux)">
           <i class="km-os-image-rockylinux"></i>
           Rocky Linux
         </mat-button-toggle>
       </mat-button-toggle-group>
+      <mat-error *ngIf="form.get(Controls.OperatingSystem).hasError('required')">
+        All supported operating system are hiden from the admin.
```
Waseem826, line 118: Wdyt?
```suggestion
        No  operating system supported.
```
Or if we want to keep the same message:
```suggestion
        All supported operating systems are hidden by the admin.
```

ahmadhamzh: i think ` No  operating system supported.` might be misleading right ?

modules/web/src/app/settings/admin/defaults/template.html
```diff
@@ -297,6 +297,56 @@
...
         </div>
 
+        <div fxLayout="row">
+          <div fxFlex="16%"
+               fxLayoutAlign=" center"
+               class="entry-label">
+            <span>Allowed Operating System</span>
+            <div class="km-icon-info km-pointer"
+                 matTooltip="Allowed operating system shows if the operating system is allowed to be use with machinedeployment,
+                  and it is only applicable for the supported operating systems for a cloud provider.">
```
Waseem826, line 307:
```suggestion
                 matTooltip="Allowed operating system shows if the operating system is allowed to be used with machine deployment,
                  and it is only applicable to the supported operating systems for a cloud provider.">
```

modules/web/src/app/settings/admin/defaults/template.html
```diff
@@ -297,6 +297,56 @@
...
+          </div>
+
+          <mat-button-toggle-group group="operatingSystemGroup"
+                                   [(ngModel)]="allowedOperatingSystems"
+                                   (change)="onOperatingSystemChange($event.value)"
+                                   multiple
+                                   fxLayout="row wrap"
+                                   kmValueChangedIndicator>
+            <mat-button-toggle [value]="OperatingSystem.Ubuntu"
+                               [disabled]="isLastEnabledOS(OperatingSystem.Ubuntu)">
```
Waseem826, line 318: Can we put this `[disabled]` on the `mat-button-toggle-group` instead?

modules/web/src/app/node-data/template.html
ahmedwaleedmalik: `All supported operating systems are hidden by the admin` Please let's note add a note like this. Let's reword this into `No  operating system supported. Please contact your administrator`

Waseem826 (commented):
Sorry I accidentally clicked the `Add Single Comment` button but I have some more comments. PTAL.

modules/web/src/app/settings/admin/defaults/template.html
ahmadhamzh: that will disable all the `mat-button-toggle` , what im trying to do here is to prevent the admin from disable all the os so if there is only one os is enabled then the admin can't disable it

modules/web/src/app/node-data/component.ts
```diff
@@ -106,6 +108,8 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
   isLoadingOSProfiles: boolean;
   isEnterpriseEdition = DynamicModule.isEnterpriseEdition;
   wizardMode: WizardMode;
+  currentNodeOS: OperatingSystem;
+  allowedOperatingSystems: AllowedOperatingSystems = {};
```
Waseem826, line 112: Just one problem here is that if settings API request fails or if `settings.allowedOperatingSystems` is empty/undefined then there won't be any allowed operating systems. So IMO we should initialize this with all OS in `ngOnInit`.
For example:
```
{
  ubuntu: true,
  flatcar: true,
  ...
}

Waseem826 (approved):

PR #6106, opened 2023-07-24
[release/v2.23] Bump version to v2.23.1

pkprzekwas:
Bump version for v2.23.1

**What type of PR is this?**
/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


modules/api/go.mod
```diff
@@ -104,8 +104,7 @@ replace (
 
 replace (
 	github.com/ajeddeloh/go-json => github.com/coreos/go-json v0.0.0-20220810161552-7cce03887f34
-	// Pinned to https://github.com/kubermatic/kubermatic/commit/4aa4aa14544e89c8ab85ff5428514585e326b3bf
```
Waseem826, line 107: Can we update this comment instead of removing it? Also, the commit in this link has a TODO note: `TODO: Replace this with v1.57.1 before release` so I guess we're still missing that change? 

https://github.com/kubermatic/kubermatic/commit/4aa4aa14544e89c8ab85ff5428514585e326b3bf#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R42

Waseem826:
I think the title of PR should be `[release/v2.23]`.

modules/api/go.mod
pkprzekwas: I wanted to ask about that on the sig meeting. Do you know why we placed those comments?

pkprzekwas: Updated both MC and OSM

Waseem826 (approved):
/approve

PR #6144, opened 2023-08-02
Fix the redirect url for keycloak logout

### LINKED ISSUE #6127: Logout url does not work with Keycloak 18.0 or higher
### What happened

The oidc logout url feature in the ui configuration is unsuitable because it hardcodes `redirect_uri` to append to the logout url. This does not work with Keycloak 18 or higher.

See [Keycloak OpenID logout documentation](https://www.keycloak.org/docs/latest/upgrading/index.html#openid-connect-logout) for details on what parameters are required.

When you actually try to configure it, you get

> Invalid parameter: redirect_uri

when clicking logout.

### Expected behavior

Keycloak logout flow is supported or to be more precise, ui configuration needs to be flexible enough to allow configuration for the Keycloak logout flow.

### Environment
- UI Version: v2.22.4
- API Version: v2.22.4
- Domain:
- Others: Keycloak 18.0+

### Current workaround

None known.

### Affected user persona

Users with Keycloak as OIDC solution.

ahmadhamzh:
Starting from Keycloak 18, [RP initiated logout](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) was introduced in keycloak. `post_logout_redirect_uri` and `id_token_hint` parameters are required now for logging a user out. 

To support this special case, we are now adding a new configuration named `oidc_provider` which will allow us to manage OIDC provider specific configurations.

Fixes #6127

**What type of PR is this?**

/kind bug

```release-note
Fix support for keycloak OIDC logout. New field `oidc_provider` was introduced to support OIDC provider specific configurations. Configuring `oidc_provider` as `keycloak` will properly configure the logout workflow.
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/modules/web/src/app/core/services/auth/service.ts
+++ b/modules/web/src/app/core/services/auth/service.ts
@@ -23,6 +23,7 @@ import {take, tap} from 'rxjs/operators';
 import {PreviousRouteService} from '../previous-route';
 import {TokenService} from '../token';
 import {UserService} from '../user';
+import {OIDCProviders} from '@app/shared/model/Config';
 
 @Injectable()
 export class Auth {
@@ -130,10 +131,18 @@ export class Auth {
     const config = this._appConfigService.getConfig();
     if (config.oidc_logout_url) {
       const logoutUrl = new URL(config.oidc_logout_url);
-      if (logoutUrl.searchParams.has('redirectUri')) {
-        logoutUrl.searchParams.set('redirectUri', this._redirectUri);
-      } else {
-        logoutUrl.searchParams.set('redirect_uri', this._redirectUri);
+      switch (config.oidc_provider?.toLowerCase()) {
+        case OIDCProviders.Keycloak:
+          logoutUrl.searchParams.set('post_logout_redirect_uri', this._redirectUri);
+          logoutUrl.searchParams.set('id_token_hint', this.getBearerToken());
+          break;
+        default:
+          if (logoutUrl.searchParams.has('redirectUri')) {
+              logoutUrl.searchParams.set('redirectUri', this._redirectUri);
+          } else {
+              logoutUrl.searchParams.set('redirect_uri', this._redirectUri);
+          }
+          break;
       }
       window.location.href = logoutUrl.toString();
     }
--- a/modules/web/src/app/project/template.html
+++ b/modules/web/src/app/project/template.html
@@ -20,7 +20,7 @@ limitations under the License.
          fxLayoutAlign=" center">
       <km-search-field (queryChange)="onSearch($event)"></km-search-field>
       <div fxFlex></div>
-      <ng-container  *ngTemplateOutlet="allProjectToggle"></ng-container>
+      <ng-container *ngTemplateOutlet="allProjectToggle"></ng-container>
       <mat-button-toggle-group class="project-view-switch"
                                group="projectsView"
                                (change)="changeView()">
@@ -335,7 +335,7 @@ limitations under the License.
 </div>
 
 <ng-template #placeholder>
-  <ng-container  *ngTemplateOutlet="allProjectToggle"></ng-container>
+  <ng-container *ngTemplateOutlet="allProjectToggle"></ng-container>
   <div fxLayout="column"
        fxLayoutAlign="center center"
        class="placeholder">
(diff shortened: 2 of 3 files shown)


modules/web/src/app/core/services/auth/service.ts
```diff
@@ -130,10 +130,16 @@ export class Auth {
...
+      if( urlArr.includes('realms')) {
+          logoutUrl.searchParams.set('post_logout_redirect_uri', this._redirectUri);
+          logoutUrl.searchParams.set('id_token_hint', this.getBearerToken());
+      }else {
+        if (logoutUrl.searchParams.has('redirectUri')) {
+          logoutUrl.searchParams.set('redirectUri', this._redirectUri);
+        } else {
+          logoutUrl.searchParams.set('redirect_uri', this._redirectUri);
+        }
       }
```
ahmedwaleedmalik, line 143: Realm name is always there since the keycloak clients are always scoped to a realm. The only case otherwise is when you are using some proxy in front of keycloak to obscure the realm name or the keyloak URL. 

So i'm not sure how we are making decision regarding parameters based on the presence of `realm` within the URL?

ahmedwaleedmalik (changes requested):

modules/web/src/app/core/services/auth/service.ts
```diff
@@ -130,10 +130,16 @@ export class Auth {
...
       const logoutUrl = new URL(config.oidc_logout_url);
-      if (logoutUrl.searchParams.has('redirectUri')) {
-        logoutUrl.searchParams.set('redirectUri', this._redirectUri);
-      } else {
-        logoutUrl.searchParams.set('redirect_uri', this._redirectUri);
+      const urlArr = config.oidc_logout_url.split('/');
+      if( urlArr.includes('realms')) {
+          logoutUrl.searchParams.set('post_logout_redirect_uri', this._redirectUri);
+          logoutUrl.searchParams.set('id_token_hint', this.getBearerToken());
+      }else {
```
ahmedwaleedmalik, line 137: Please fix the indentations here.

modules/web/src/app/project/template.html
```diff
@@ -374,11 +374,11 @@ <h2>Get Started</h2>
...
 <ng-template #allProjectToggle>
   <mat-slide-toggle *ngIf="isAdmin"
-                        class="all-projects-toggle"
-                        [disabled]="isProjectsLoading"
-                        labelPosition="before"
-                        [matTooltip]="isProjectsLoading ? 'Please wait while projects are loading' : null"
-                        [ngModel]="settings.displayAllProjectsForAdmin"
-                        (toggleChange)="changeProjectVisibility()">Show All Projects
+                    class="all-projects-toggle"
+                    [disabled]="isProjectsLoading"
```
ahmedwaleedmalik, line 378: Do we know why these issues with indentations are not being caught by our pipelines? It seems like a common occurrence these days. Because these should have been already fixed in the last PR where these files were modified. 

I noticed this issue a few weeks ago as well.

An example from your own PR would be https://github.com/kubermatic/dashboard/pull/6144/files#diff-1f46f73009b38ea4ecfc48cd53babc9b1f45fb74a441dbab8cd538c1c1c8a190R134, CI should have complained about this.

modules/web/src/app/core/services/auth/service.ts
ahmadhamzh: how about we use a new value in the `config` like `config.oidc_provider` ? or something ?
@Waseem826 @ahmedwaleedmalik

xrstf:
I just have to say it, sorry, but:

It's **keycloak**, not keyclock.

Thank you. ;-)

modules/web/src/app/core/services/auth/service.ts
```diff
@@ -130,10 +130,18 @@ export class Auth {
     const config = this._appConfigService.getConfig();
     if (config.oidc_logout_url) {
       const logoutUrl = new URL(config.oidc_logout_url);
-      if (logoutUrl.searchParams.has('redirectUri')) {
-        logoutUrl.searchParams.set('redirectUri', this._redirectUri);
-      } else {
-        logoutUrl.searchParams.set('redirect_uri', this._redirectUri);
+      switch (config.oidc_provider?.toLowerCase()) {
+        case 'keyclock':
```
ahmedwaleedmalik, line 134: Please create a constant somewhere for this and don't hard-code it here. Also `s/keyclock/keycloak`.

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
@ahmadhamzh You'll have to manually backport this PR to 2.23, 2.22, 2.21. Also please ensure that the documentation makes its way into those versions as well before the next patch releases.

PR #6147, opened 2023-08-04
Update prettier config

Waseem826:
Update prettier config to fix the issue with linter not formatting files. Also run linter again to update all files according to latest prettier version.

**What type of PR is this?**

/kind chore

- `requirePragma` option is removed because in order for it to be enabled we need to add a special comment in all files which we want to format. Reference: https://prettier.io/docs/en/options#require-pragma
- `trailingComma` option is removed because setting it to `all` would add trailing commas in multi-line function parameters and function calls (Ref: https://prettier.io/docs/en/options#trailing-commas). So we keep the `es5` value set in default `gts` config.

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


modules/web/src/app/cluster/details/cluster/edit-provider-settings/alibaba-provider-settings/component.ts
```diff
@@ -33,7 +33,10 @@ export class AlibabaProviderSettingsComponent implements OnInit, OnDestroy {
   readonly Control = Control;
   form: FormGroup;
 
-  constructor(private readonly _clusterService: ClusterService, private readonly _builder: FormBuilder) {}
+  constructor(
+    private readonly _clusterService: ClusterService,
+    private readonly _builder: FormBuilder
+  ) {}
```
Waseem826, line 39: This change (and other similar changes) is due to an update in prettier 3.0 and unfortunately there is no option to disable it as of now. Ref: https://github.com/prettier/prettier/issues/5666#issuecomment-1642084041

ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik (approved):
/lgtm

PR #6153, opened 2023-08-07
[release/v2.23] Fix the redirect URI for keycloak

ahmadhamzh:
manual backporting to v2.23
xref #6144  #6127

```release-note
Fix support for keycloak OIDC logout. New field `oidc_provider` was introduced to support OIDC provider specific configurations. Configuring `oidc_provider` as `keycloak` will properly configure the logout workflow.
```


Waseem826 (approved):
/approve

PR #6156, opened 2023-08-10
openstack client: include TenantID in list options

pkprzekwas:
If openstack user configured in credentials is an admin, our API lists all the networks, security groups and subnet pools that user has access to without filtering by provided ProjectID/ProjectName. This PR introduces additional filtering for networks, security groups and subnet pools listings.

**What type of PR is this?**
/kind bug

```release-note
openstack: take `TenantID` into account while listing networks, security groups and subnet pools.
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #6258, opened 2023-10-06
Set default value for cni plugin version 

### LINKED ISSUE #6233: Dashboard doesn't use `cni.GetDefaultCNIPluginVersion` to determine default Cilium version
### What happened
When Cilium 1.14.1 was added to the list of supported Cilium versions by mistake (https://github.com/kubermatic/kubermatic/issues/12662), the problem was intensified by the fact that it's selected default in the dropdown during cluster creation.

I did a cursory sweep of the code base and it seems that the `modules/api` code never calls `cni.GetDefaultCNIPlugionVersion(kubermaticv1.CNIPluginTypeCilium)` to determine the default version to select for Cilium. Since the `cni` package in `k8c.io/kubermatic/v2` has the explicit "default CNI version" setting that is also used when defaulting clusters, IMHO the dashboard should respect that setting and offer the same version by default for consistency reasons.

### Expected behavior

Cilium 1.13.6 should have been the by default selected CNI version for Cilium because it's configured that way in `k8c.io/kubermatic/v2/pkg/cni`.

### How to reproduce

- Install KKP v2.23.3 (the version with the bug/regression mentioned above).
- Go to dashboard, create new cluster, select Cilium as CNI.
- See that it defaults to selecting 1.14.1.
- Check KKP code at https://github.com/kubermatic/kubermatic/blob/e0c288ef359024bfdaede7e64d06ced46bf68f25/pkg/cni/version.go#L36 to determine that 1.13.6 was meant to be the default version.

### Environment
- UI Version: v2.23.3, main
- API Version: v2.23.3, main
- Domain:
- Others:

### Current workaround

Manually select another version from the CNI dropdown.

ahmadhamzh:
use `cni.GetDefaultCNIPluginVersion` to get the default value for cni Plugin version .

Fixes #6233 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/wizard/step/cluster/component.ts
+++ b/modules/web/src/app/wizard/step/cluster/component.ts
@@ -278,7 +278,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
         this._updateAvailableProxyModes();
         this.form.get(Controls.CNIPluginVersion).setValue('');
         this.cniPluginVersions = cniVersions.versions.sort((a, b) => compare(coerce(a), coerce(b)));
-        this._setDefaultCNIVersion();
+        this._setDefaultCNIVersion(cniVersions.cniDefaultVersion);
       });
 
     this.control(Controls.Konnectivity)
@@ -639,7 +639,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       .pipe(take(1))
       .subscribe(versions => {
         this.cniPluginVersions = versions.versions.sort((a, b) => compare(coerce(a), coerce(b)));
-        this._setDefaultCNIVersion();
+        this._setDefaultCNIVersion(versions.cniDefaultVersion);
       });
   }
 
@@ -701,7 +701,7 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
     }
   }
 
-  private _setDefaultCNIVersion(): void {
+  private _setDefaultCNIVersion(cniDefaultVersion: string): void {
     if (this.cniPluginVersions.length > 0 && !this.form.get(Controls.CNIPluginVersion).value) {
       // Dual-stack not allowed on Canal CNI version lower than 3.22.
       if (this.controlValue(Controls.CNIPlugin) === CNIPlugin.Canal && this.isDualStackIPFamilySelected()) {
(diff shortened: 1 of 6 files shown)


modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -277,8 +277,8 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
       .subscribe(cniVersions => {
         this._updateAvailableProxyModes();
         this.form.get(Controls.CNIPluginVersion).setValue('');
-        this.cniPluginVersions = cniVersions.versions.sort((a, b) => compare(coerce(a), coerce(b)));
```
Waseem826, line 280: Any particular reason for removing the sort?

modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -701,16 +702,19 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
...
     if (this.cniPluginVersions.length > 0 && !this.form.get(Controls.CNIPluginVersion).value) {
       // Dual-stack not allowed on Canal CNI version lower than 3.22.
       if (this.controlValue(Controls.CNIPlugin) === CNIPlugin.Canal && this.isDualStackIPFamilySelected()) {
         this.cniPluginVersions = this.cniPluginVersions.filter(v =>
           gte(coerce(v), this._canalDualStackMinimumSupportedVersion)
         );
       }
-
-      this.form.get(Controls.CNIPluginVersion).setValue(this.cniPluginVersions[this.cniPluginVersions.length - 1]);
+      if (cniDefaultVersion) {
```
Waseem826, line 713: We're filtering `cniPluginVersions` above so here we should also check if the default version exists in filtered list. For example, if `Canal` plugin is selected and the default version is not supported.

Waseem826 (commented):
Some feedback. PTAL

Waseem826 (approved):
/approve

Waseem826:
/assign @ahmedwaleedmalik 
Please `/approve`.

ahmedwaleedmalik:
@ahmadhamzh this needs a release note.

PR #6337, opened 2023-10-26
fix listing events for external clusters

### LINKED ISSUE #6331: Error `unable to list: default because of unknown namespace for the cache` - External Clusters
### What happened
on create or import external cluster  Error `unable to list: default because of unknown namespace for the cache` keep coming from `http://localhost:8000/api/v2/projects/wt8bxznl97/kubernetes/clusters/2g6zd8vwkw/events`. 

image

### How to reproduce
* create or import external cluster 
* navigate to the external cluster detail page

xrstf:
Since controller-runtime introduced the new caching config, we had to explicitly enable caches per namespace. Looking at our dev system, I spotted

> {"level":"error","time":"2023-10-26T06:36:58.837Z","caller":"handler/routing.go:153","msg":"unable to list: default because of unknown namespace for the cache","request":"/api/v2/projects/wt8bxznl97/kubernetes/clusters/2g6zd8vwkw/events"}

Looking deeper, it seemed that we accidentally specified a namespace when ExternalClusters are not namespaced.

Fixes #6331

We can backport this, even though the error should only surface with controller-runtime 0.16, i.e. KKP 2.24.

**What type of PR is this?**
/kind bug

```release-note
Fix listing events for external clusters
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/handler/common/cluster.go
+++ b/modules/api/pkg/handler/common/cluster.go
@@ -655,7 +655,7 @@ func GetClusterEventsEndpoint(ctx context.Context, userInfoGetter provider.UserI
 		eventTypeAPI = corev1.EventTypeNormal
 	}
 
-	events, err := common.GetEvents(ctx, client, cluster, "")
+	events, err := common.GetEvents(ctx, client, cluster, metav1.NamespaceAll)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve

PR #6341, opened 2023-10-26
Fix GKE Credentials validation

### LINKED ISSUE #6333: GKE field validation and cluster Version not updated
### What happened

* the service account credentials have a note that it have to be base64 but actually even if add the service account as base64 the field validation shows an error but if i create a preset and i add the service account ( not in base64) it worked 

image

* cluster version and control plan version are not the same after update.

image

### Expected behavior
* the service account credentials should accept the service account object
* on update control plan version cluster version should show the same version
### How to reproduce
* try to create external cluster on GKE  on second step (settings) 
* create an external Cluster on GKE don't chose the latest k8s version then after the cluster been created update the control plan version cluster version should show the same value

ahmadhamzh:
* send the encode value of service Account instead of the string value

* remove the `cluster version`  property from the external cluster detail since it's control plan version already show the version 

image

Fixes #6333

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/cluster/details/external-cluster/template.html
+++ b/modules/web/src/app/cluster/details/external-cluster/template.html
@@ -117,10 +117,6 @@ limitations under the License.
             <div value>{{cluster?.cloud?.aks?.resourceGroup}}</div>
           </km-property>
           <km-property>
-            <div key>Cluster Version</div>
-            <div value>{{cluster?.spec?.aksclusterSpec?.kubernetesVersion}}</div>
-          </km-property>
-          <km-property>
             <div key>Provisioning State</div>
             <div value>{{cluster?.status?.aks?.provisioningState}}</div>
           </km-property>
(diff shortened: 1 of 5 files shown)


Waseem826:
Is there an issue on the API side which is causing this validation to fail?

ahmadhamzh:
yes i think the issue [here](https://github.com/kubermatic/dashboard/blob/cff3ed4c21691cb4fafb1055c824c28bbe9f24e5/modules/api/pkg/handler/v2/external_cluster/gke.go#L1120) but im not quite sure why the preset work what should be sending im not quite sure

ahmadhamzh:
so it turn out the issue from the UI we were sending the SA value without encoding it.
i thought that we do the encoding in the api but i was wrong.
@ahmedwaleedmalik  @Waseem826

Waseem826 (approved):
/approve

ahmedwaleedmalik (approved):
/approve
/unhold

PR #6345, opened 2023-10-27
Add disableDeletion method in EKS 

### LINKED ISSUE #6335: EKS disable deletion on nodegroup warning
### What happened

* when you try to delete the cluster while we have a nodegroup we get a warning that we need to delete the Nodegroup first, so it would be better if we keep the delete button disabled 

image

### Expected behavior
* as long as the warning shown the delete button should be disable

ahmadhamzh:
disable the deletion button on the delete external cluster dialog for EKS in case there is any nodegroup 

Fixes #6335 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #6346, opened 2023-10-27
Remove the patch version in edit MD for EKS

### LINKED ISSUE #6334: EKS cluster k8s version list and edit value
### What happened
* the list of k8s versions don't have the 1.28 version.

image

* on edit MD the value to update the version must not have the patch release value (e.g 1.27.0 will not work wich is how we do it now, it must be 1.27).

ahmadhamzh:
in eks we shouldn't send the patch version for updating the machinedeployment

Fixes #6334 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/cluster/details/external-cluster/update-external-cluster-machine-deployment-dialog/component.ts
+++ b/modules/web/src/app/cluster/details/external-cluster/update-external-cluster-machine-deployment-dialog/component.ts
@@ -26,6 +26,7 @@ import {BaseFormValidator} from '@shared/validators/base-form.validator';
 import {Observable} from 'rxjs';
 import {filter, take, takeUntil} from 'rxjs/operators';
 import {MasterVersion} from '@shared/entity/cluster';
+import {minor, major} from 'semver';
 
 class UpdateExternalClusterMachineDeploymentDialogData {
   projectID: string;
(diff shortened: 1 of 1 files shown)


modules/web/src/app/cluster/details/external-cluster/update-external-cluster-machine-deployment-dialog/component.ts
```diff
@@ -137,13 +138,17 @@ export class UpdateExternalClusterMachineDeploymentDialogComponent extends BaseF
...
+    const endSliceParameter = 4;
     return {
       spec: {
         replicas: this.form.get(Controls.Replicas).value,
         template: {
           versions: {
-            kubelet: this.form.get(Controls.KubeletVersion).value,
+            // in EKS we shouldn't send the patch with the full version, only the major version.
+            kubelet: isEKSProvider
+              ? this.form.get(Controls.KubeletVersion).value.slice(0, endSliceParameter)
```
Waseem826, line 150: I think we should use the `semver` package for parsing instead of using `slice` just to avoid any possibility of parsing error. For example, if version is `1.23-beta` or `1.100.1` then `slice` would not work as expected.

modules/web/src/app/cluster/details/external-cluster/update-external-cluster-machine-deployment-dialog/component.ts
```diff
@@ -137,16 +139,24 @@ export class UpdateExternalClusterMachineDeploymentDialogComponent extends BaseF
...
         template: {
           versions: {
-            kubelet: this.form.get(Controls.KubeletVersion).value,
+            kubelet: kubeletVersion,
           },
         },
       } as ExternalMachineDeploymentSpecPatch,
     };
+
+    // in EKS we shouldn't send the patch with the full version, only the major version.
```
Waseem826, line 156: Very small nitpick.
```suggestion
    // In EKS we shouldn't send the patch with the full version, only the major and minor versions.
```

Waseem826 (approved):
/approve

PR #6395, opened 2023-11-23
Option to disable CSI drivers

### LINKED ISSUE #6206: Option to disable CSI drivers
## User Story

As a user, I'd like the option to disable CSI drivers for my clusters. This is required in order to leverage external storage providers.

## Solution details

With https://github.com/kubermatic/kubermatic/issues/12008, we'll implement the backend support for this feature. There are a few things to consider when implementing this feature:
1. If CSI drivers are disabled at data center level then the user shouldn't be allowed to override this behavior when creating user clusters.
2. In the `Cluster` step in the cluster creation wizard show an option to enable/disable CSI drivers.
3. If CSI drivers are disabled, hide the CSI driver-related option from the `Settings` step. Mainly applicable to Nutanix and VMware Cloud Director but please investigate if other cloud providers also need such changes.
4. Users should be allowed to disable CSI drivers after cluster creation. At the time of creating this ticket, the implementation details have not been finalized. But the idea is that the Cluster status will have a condition/field that can be used to determine if this operation is permitted or not.

ahmedwaleedmalik:
This PR adds the required changes in API and UI to support disabling CSI drivers for user clusters.

Fixes #6206

**What type of PR is this?**

/kind feature

```release-note
Support for disabling CSI driver for user clusters
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/cmd/kubermatic-api/swagger.json
+++ b/modules/api/cmd/kubermatic-api/swagger.json
@@ -28190,6 +28190,11 @@
           "type": "string",
           "x-go-name": "ContainerRuntime"
         },
+        "disableCsiDriver": {
+          "description": "Optional: DisableCSIDriver disables the installation of CSI driver on the cluster\nIf this is true at the data center then it can't be over-written in the cluster configuration",
+          "type": "boolean",
+          "x-go-name": "DisableCSIDriver"
+        },
         "enableOperatingSystemManager": {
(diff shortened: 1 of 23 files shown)


Waseem826 (approved):
/approve

ahmedwaleedmalik:
/approve

For API 🦀

PR #6422, opened 2023-12-12
Prune `id` value from applications in request payload when editing cluster template

Waseem826:
Missed this change in https://github.com/kubermatic/dashboard/pull/6415. `id` value was still attached in the request payload when editing cluster template.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


Waseem826:
/hold
Waiting for previous PR to be cherrypicked.

PR #6460, opened 2024-01-12
Anexia: Adjust MachineDeployment form to current spec

anx-mschaefer:
Apply MachineDeployment spec changes from https://github.com/kubermatic/machine-controller/pull/1541 and https://github.com/kubermatic/machine-controller/pull/1665 to the dashboard.

This PR adds configuration support for named templates and multiple disks when using the Anexia provider and also fixes a validation bug occurring when listing MachineDeployments which have named templates configured.

**What type of PR is this?**
/kind bug
/kind cleanup
/kind feature
/kind api-change

Haven't done any angular before.. please let me know if any of my changes need improvement :sweat_smile: 

```release-note
Anexia Provider:
- Add configuration support for named templates
- Add configuration support for multiple disks
- diskSize attribute gets automatically migrated to the disks attribute when saved
- Fix error occurring when listing MachineDeployments which have named templates configured
```

**Documentation**:

```documentation
NONE
```


anx-mschaefer:
Hi @embik I've looked into the failed codecov report and noticed that other providers aren't covered (in regards of the changes) either. Should we add some tests anyway?

Waseem826:
@anx-mschaefer you can ignore the codecov report.

Waseem826:
I think you also need to update Project quota calculation logic because I get an error from the `quotacalculation` endpoint when I open create cluster wizard:
```
{
  "code": 400,
  "message": "json: cannot unmarshal object into Go struct field AnexiaDiskConfig.anexiaNodeSpec.disks.performanceType of type string"
}
```
Also, the estimated project quota widget in create cluster wizard is also not working.

Waseem826:
> From what I understand, the `quotacalculation` feature is only available in the enterprise version of KKP. Is it okay for me to build and run the enterprise edition of KKP and the dashboard without a license for testing purposes?

cc @embik

modules/web/src/app/shared/entity/node.ts
```diff
@@ -161,9 +161,16 @@ export class NodeSystemInfo {
 export class AnexiaNodeSpec {
   vlanID: string;
   templateID: string;
+  template: string;
   cpus: number;
   memory: number;
   diskSize: number;
```
Waseem826, line 167: Since we're not using `diskSize` anymore so should we delete this field as well?

modules/web/src/app/shared/entity/node.ts
Waseem826, line 1: Please remove `diskSize` value from `getDefaultNodeProviderSpec` method (for Anexia). That method is called when Add Machinedeployment Dialog is opened and it causes the `diskSize` value being sent in API request payload.

modules/web/src/app/shared/components/cluster-summary/template.html
```diff
@@ -989,6 +989,10 @@ <h4>IPv6</h4>
                 <div key>Template ID</div>
                 <div value>{{machineDeployment.spec.template.cloud?.anexia?.templateID}}</div>
               </km-property>
+              <km-property>
+                <div key>Template Name</div>
+                <div value>{{machineDeployment.spec.template.cloud?.anexia?.template}}</div>
+              </km-property>
```
Waseem826, line 995: We should remove `Disk Size in GB` and display added disks.
Also, we should display only one of Template ID or Template Name (depends which one has value) since both cannot have values together.

modules/web/src/app/node-data/basic/provider/anexia/component.ts
```diff
@@ -239,8 +327,8 @@ export class AnexiaBasicNodeDataComponent extends BaseFormValidator implements O
       anexiaNodeSpec: {
         [Controls.Cpus]: this.form.get(Controls.Cpus).value,
         [Controls.Memory]: this.form.get(Controls.Memory).value,
-        [Controls.DiskSize]: this.form.get(Controls.DiskSize).value,
-        [Controls.TemplateID]: this.form.get(Controls.TemplateID).value?.[AutocompleteControls.Main],
+        [Controls.Disks]: this.form.get(Controls.Disks).value,
```
Waseem826, line 330: As mentioned in earlier comment, we need to send the `disks` property in the same format which we send for creating the cluster.

modules/web/src/app/node-data/basic/provider/anexia/component.ts
```diff
@@ -208,14 +275,36 @@ export class AnexiaBasicNodeDataComponent extends BaseFormValidator implements O
...
+    for (let i = 0; i < this.disks.length; i++) {
+      if (
+        this._nodeDataService.nodeData.spec.cloud.anexia.disks.length > i &&
+        this._nodeDataService.nodeData.spec.cloud.anexia.disks[i].performanceType
+      ) {
+        this.disks
+          .at(i)
+          .get('performanceType')
+          .setValue({main: this._nodeDataService.nodeData.spec.cloud.anexia.disks[i].performanceType});
+      }
```
Waseem826, line 305: I think we don't need this here. `performanceType` is already set for `disks` inside the `_init` method. Or am I missing something?

modules/web/src/app/node-data/basic/provider/anexia/template.html
```diff
@@ -30,18 +30,42 @@
...
+                           min="0"
+                           fxFlex="45">
+        </km-number-stepper>
+
+        <km-autocomplete label="Disk Type"
+                         fxFlex="40"
+                         formControlName="performanceType"
+                         [isLoading]="isLoadingDiskTypes"
+                         [options]="diskTypes">
+        </km-autocomplete>
```
Waseem826, line 52: Just to confirm, is it allowed to enter custom `performanceType` value? i.e. a value which is not inside the available disk types.

Waseem826 (changes requested):
Some feedback. PTAL

modules/web/src/app/node-data/basic/provider/anexia/template.html
anx-mschaefer: Configuring a disk using a performanceType which is not included in the available disk types will fail during VM provisioning in the machine-controller. (Similarly when configuring non-existing VLANs)

modules/web/src/app/shared/entity/node.ts
anx-mschaefer: The `diskSize` attribute is technically still available in the machine-controller, but deprecated. I would expect both specs to be similar.

Waseem826 (approved):
/approve

cc @ahmedwaleedmalik @SimonTheLeg for API review/approval.

ahmedwaleedmalik (approved):
/approve 🤞

PR #6478, opened 2024-01-25
Add backup storage location page

### LINKED ISSUE #6466: Add new page for backup storage location
Add new page to manage backup storage locations  for cluster backups

ahmadhamzh:
add new page to manage backup storage location for the cluster backup feature.

* **The Page**
image

* **Add Dialog**
image

Fixes #6466 

**What type of PR is this?**
/kind feature

```release-note
add new page to manage backup storage location for the cluster backup feature.
```

```documentation
TBD
```


modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -0,0 +1,207 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type BackupStorageProvider struct {
+	// createMasterImpersonatedClient is used as a ground for impersonation
+	// whenever a connection to Seed API server is required
+	createMasterImpersonatedClient kubernetes.ImpersonationClient
```
SimonTheLeg, line 48: more of question: What is the purpose of having the createmasterImpersonatedClient in here? I think it does not get used throughout the code.

I know that we have this in a couple of places like the GroupProjectBindingProvider. But from my understading they are using it directly

https://github.com/kubermatic/dashboard/blob/03648c694cca330f62ad519cac4338a2bd9af6f5/modules/api/pkg/ee/group-project-binding/provider/provider.go#L179

modules/api/pkg/api/v1/types.go
```diff
@@ -1019,6 +1019,12 @@ type ClusterSpec struct {
 	// By default it is disabled.
 	OPAIntegration *kubermaticv1.OPAIntegrationSettings `json:"opaIntegration,omitempty"`
 
+	// A map of optional or early-stage features that can be enabled for the user cluster.
+	// Some feature gates cannot be disabled after being enabled.
+	// The available feature gates vary based on KKP version, Kubernetes version and Seed configuration.
+	// Please consult the KKP documentation for specific feature gates.
+	Features map[string]bool `json:"features,omitempty"`
```
SimonTheLeg, line 1026: If I understand it correctly what this line does is that the features field can now also be set through the api.
If that is the case and you would need to set it through KKP api, then wouldn't there be an equivalent for "ClusterBackUpStorage" in the list from here: https://github.com/kubermatic/kubermatic/blob/main/pkg/apis/kubermatic/v1/cluster.go#L325

But I could be wrong, I think someone from sig-cluster-mgmt needs to take a look at this. Or maybe to ask the other way around: What feature will be set through the api with this change?

modules/api/pkg/ee/clusterbackup/storage-location/handler.go
```diff
@@ -0,0 +1,282 @@
...
+type getCbslReq struct {
+	common.ProjectReq
+	// in: path
+	// required: true
+	ClusterBackupStorageLocationName string `json:"clusterBackupStorageLocation"`
+}
+
+type createCbslReq struct {
+	common.ProjectReq
+	// in: path
```
SimonTheLeg, line 55: If I understand the Go tags correctly, then I think this should be an `// in: body` tag. At least this is what we have done in other places e.g. https://github.com/kubermatic/dashboard/blob/03648c694cca330f62ad519cac4338a2bd9af6f5/modules/api/pkg/handler/v2/application_installation/request.go#L41-L42

modules/api/pkg/ee/clusterbackup/storage-location/handler.go
```diff
@@ -0,0 +1,282 @@
...
+	// required: true
+	ClusterBackupStorageLocationName string `json:"clusterBackupStorageLocation"`
+}
+
+type patchCbslReq struct {
+	common.ProjectReq
+	// in: path
+	// required: true
+	ClusterBackupStorageLocationName string `json:"clusterBackupStorageLocation"`
+	Body                             CbslBody
```
SimonTheLeg, line 79: Same as above with the body tags

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -7830,6 +7853,51 @@ func (r Routing) deleteClusterBackupSchedule() http.Handler {
 	)
 }
 
+func (r Routing) listProjectCBSL() http.Handler {
```
SimonTheLeg, line 7856: From my perspective, swagger route comments are missing. See below for an example. I have to admit, I am not sure why they are missing for the already commited code for ClusterBackUp from above. But from my perspective the comments need to be there, so the routes are correctly displayed in swagger.

If we want to intentionally hide these routes from swagger, then I would like us to have a short comment in code that says so.

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -7830,6 +7853,51 @@ func (r Routing) deleteClusterBackupSchedule() http.Handler {
...
+	return httptransport.NewServer(
+		endpoint.Chain(
+			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
+			middleware.UserSaver(r.userProvider),
+		)(storagelocation.ListCBSLEndpoint(r.userInfoGetter, r.backupStorageProvider, r.projectProvider)), storagelocation.DecodeListProjectCBSLReq, handler.EncodeJSON, r.defaultServerOptions()...,
```
SimonTheLeg, line 7861: So from my understanding of the api, this should not work: 
1. You will need to add the corresponding Decode Function from your handler.go as the httpDecodeRequestFunc  argument to httptransport.NewServer
2. You will need the handler.EncodeJSON as the httpEncodeRequestFunc argument to httptransport.NewServer

For example have a look here https://github.com/kubermatic/dashboard/blob/27df2fc9d1c9873dc07a019eff29e1a4481e21dc/modules/api/pkg/handler/v2/routes_v2.go#L7837-L7838

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/template.html
```diff
@@ -0,0 +1,142 @@
...
+               class="km-table-actions">
+            <button mat-icon-button
+                    matTooltip="Edit"
+                    (click)="editBSL(element); $event.stopPropagation()"
+                    [disabled]="!canAdd">
+              <i class="km-icon-mask km-icon-edit"></i>
+            </button>
+
+            <button mat-icon-button
+                    matTooltip="Delete Backup"
```
Waseem826, line 114:
```suggestion
                    matTooltip="Delete Backup Storage Location"
```

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/add-dialog/component.ts
```diff
@@ -105,6 +105,9 @@ export class AddClustersBackupsDialogComponent implements OnInit, OnDestroy {
     this.form.get(Controls.Destination).disable();
     this.getClusterNamespaces(this.projectID, this.cluster.id);
 
+    // this feild for now will be disable untill we have a list of BSL
+    this.form.get(Controls.Destination).disable();
```
Waseem826, line 109: This field is already disabled above (Line# 105). Also, do we not have list of BSL now?

SimonTheLeg (changes requested):
I think there is something not right in the `routes_v2.go` file specifically. I am not sure if things have changed, since I worked last on API, but does the current code return the correct json responses?

Apart from this only some minor comments/ questions from my side

Waseem826 (changes requested):
Some feedback. PTAL

Waseem826:
/approve
UI changes look good.

modules/api/pkg/api/v1/types.go
ahmadhamzh: Actually, this needs to be removed. It was added before because of another approach we wanted to use, but right now, we don't want it. I'll delete it.

modules/api/pkg/handler/v2/routes_v2.go
ahmadhamzh: Actually, the endpoint is working, and I can list the CBSL from the UI. However, I'm not quite sure what issue you're pointing to because the `storagelocation.DecodeListProjectCBSLReq` and `handler.EncodeJSON` are there.

SimonTheLeg:
/approve

BE looks good to me

Waseem826 (approved):

PR #6489, opened 2024-01-31
Add new option for default volumes file system 

ahmadhamzh:
add new field in the add backup/schedule to chose whether pod volume file system backup should be used for all volumes by default.
also for changing the default value of BSL and removing the destinations.
In an upcoming PR, the list of BSL will be added.

image

```release-note
NONE
```

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #6506, opened 2024-02-09
Set LoadBalancerSKU from Azure preset

embik:
Looks like the preset-to-spec logic never set the load balancer SKU. I'm surprised to see this, but there are no other references to those fields in the dashboard code.

This PR sets the field from the preset. If it's empty, there is still defaulting code in KKP that will put `basic` as value.

Fixes https://github.com/kubermatic/kubermatic/issues/13063

**What type of PR is this?**
/kind bug

```release-note
Set LoadBalancerSKU on Azure clusters if the field is set in the preset
```

**Documentation**:

```documentation
NONE
```


embik:
/retest

flake?

embik:
e2e test is currently broken because commit-based images have been cleaned up, @xrstf is looking into it.

ahmedwaleedmalik (approved):
/approve

PR #6543, opened 2024-02-20
Fix select BSL on cluster creation

ahmadhamzh:
fix an issue with choosing BSL in the cluster creation wizard and show the `displayName` for the user instead of the `name`

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #6551, opened 2024-02-23
Exclude machine controller health check when edge provider is used

moadqassem:
Machine controller health check must be excluded when the edge cloud provider is used. This PR guarantees that no check for machine controller is present when the edge provider used. Another PR will be open in KKP to mitigate that abstract machine controller health check over there.  

Fixes #

**What type of PR is this?**
/kind bug

```release-note
None
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/modules/api/pkg/provider/kubernetes/cluster.go
+++ b/modules/api/pkg/provider/kubernetes/cluster.go
@@ -281,8 +281,17 @@ func (p *ClusterProvider) Get(ctx context.Context, userInfo *provider.UserInfo, 
 		return nil, err
 	}
 	if options.CheckInitStatus {
-		if !cluster.Status.ExtendedHealth.AllHealthy() {
-			return nil, apierrors.NewServiceUnavailable("Cluster components are not ready yet")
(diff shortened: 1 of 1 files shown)


modules/api/pkg/provider/kubernetes/cluster.go
```diff
@@ -281,8 +281,17 @@ func (p *ClusterProvider) Get(ctx context.Context, userInfo *provider.UserInfo,
 		return nil, err
 	}
 	if options.CheckInitStatus {
-		if !cluster.Status.ExtendedHealth.AllHealthy() {
-			return nil, apierrors.NewServiceUnavailable("Cluster components are not ready yet")
+		healthCheck := cluster.Status.ExtendedHealth
+		if cluster.Spec.Cloud.ProviderName == string(kubermaticv1.EdgeCloudProvider) {
```
SimonTheLeg, line 285: I think it would be good if we add a comment here, why edgeprovider is different / why it's fine for edgeprovider if at least one of the components is healthy

PR #6562, opened 2024-02-27
Add valuesBlock and defaultValuesBlock and their conversion

SimonTheLeg:

This PR adds makes the new fields valuesBlock and defaultValuesBlock of AppInstalls and AppDefs available in the api. In addition on POST & PUT requests for AppInstalls/AppDefs the values/defaultValues fields are being converted to their valuesBlock/defaultValuesBlock equivalent.

Fixes https://github.com/kubermatic/kubermatic/issues/12941

**What type of PR is this?**

/kind feature

```release-note
valuesBlock and defaultValuesBlock fields are now available via the api
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/cmd/kubermatic-api/swagger.json
+++ b/modules/api/cmd/kubermatic-api/swagger.json
@@ -27490,6 +27490,11 @@
         },
         "values": {
           "$ref": "#/definitions/RawExtension"
+        },
+        "valuesBlock": {
+          "description": "ValuesBlock specifies values overrides that are passed to helm templating. Comments are preserved.",
+          "type": "string",
+          "x-go-name": "ValuesBlock"
         }
       },
       "x-go-package": "k8c.io/dashboard/v2/pkg/api/v2"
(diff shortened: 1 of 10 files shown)


Waseem826:
Do we still need any UI changes after this?

SimonTheLeg:
I am not 100% sure. If the UI can also handle a ValuesBlock response from the BE then I think we are already fine. 

In case we need to touch the FE code: I would appreciate if we could make all PUT/POST requests also use the new valuesBlock field (but only if it is not too much work).

Waseem826 (approved):
/approve

Waseem826:
@SimonTheLeg I think you'll need to self approve this PR.

Waseem826:
It seems like the API still uses `values` for applications added during cluster creation. It uses `valuesBlock` only for the `applicationinstallation` endpoint.
I added an application during cluster creation and the application installation returned `values` field (no `valuesBlock` field in response).
If I add/update an application installation later on (from clusters page), API returns the new `valuesBlock` field.

SimonTheLeg:
oh I see. This is something I did not consider. I'll have a look at it and will write a fix.

PR #6567, opened 2024-02-29
Fix handling of cloud credentials for vSphere when editting provider

### LINKED ISSUE #6550: vSphere cluster credentials lack the checkbox for dedicated credentials for cloud provider
### What happened

After cluster been created in edit credentials view it asks for both vSphere user/password AND cloud provider credentials, but the checkbox for dedicated credentials for cloud provider is absent. So it ends up to require both sets of credentials always.

image

### Expected behavior

Present the checkbox in like in cluster creation view to enable or disable second set of credentials:
image

### Environment
- UI Version:
- API Version:
- Domain:
- Others:

### Affected user persona

User cluster owner

ahmedwaleedmalik:
Infra management credentials should not be "required" as they are optional credentials. 

Fixes #6550

**What type of PR is this?**

/kind bug

```release-note
vSphere: fix a bug where dedicated credentials were incorrectly being required as mandatory input when editing provider settings for a cluster
```

**Documentation**:

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #6578, opened 2024-03-05
Set replicas to 1 when edge provider is used

moadqassem:
The edge provider is not supported by machine controller by design since it doesn't play any role in the node lifecycle. Thus the replica should be set to 1 and the status should be also updated.

Fixes #

**What type of PR is this?**
/kind bug

```release-note
None
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/modules/api/pkg/resources/machine/machinedeployment.go
+++ b/modules/api/pkg/resources/machine/machinedeployment.go
@@ -24,7 +24,6 @@ import (
 	"strings"
 
 	semverlib "github.com/Masterminds/semver/v3"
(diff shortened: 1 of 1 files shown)


modules/api/pkg/resources/machine/machinedeployment.go
```diff
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	semverlib "github.com/Masterminds/semver/v3"
+	"github.com/aws/smithy-go/ptr"
```
ahmedwaleedmalik, line 27: Please use "k8s.io/utils/ptr" instead

PR #6583, opened 2024-03-05
Fix application status message

### LINKED ISSUE #6582: Extra Space
Screenshot_20240305_142758

There is an extra space between the condition's message and the suffix text that the UI adds, probably because of

```ts
message = `${warning} ${
  warning || !warning.endsWith('.') ? '.' : ''
} Application is in an unknown state.`;
```

Maybe it would make sense to flip the texts around: "Application is an unknown state: [reason]", as in "Application is in an unknown state: application is intalling or updating".

Waseem826:
Fix application status message.

Fixes #6582 

**What type of PR is this?**

/kind design

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #6587, opened 2024-03-06
Disable restore backup button if snabshot/automatic are on deletion

### LINKED ISSUE #6580: Disable restore button when deleting ETCD snapshot/automatic backup
### What happened
in ETCD backups, on deleting a snapshot or automatic backup i can still restore from the backups .

### Expected behavior
user should not be able to restore from a backup if the backup in deletion state.

### How to reproduce
Try creating a snapshot, then deleting it. While the snapshot is being deleted, try to create a restore from a backup

ahmadhamzh:
Disable restore backup button if snabshot/automatic are on deletion status, and fix the status column for the snapshot table.

Fixes #6580 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve
/retest

PR #6589, opened 2024-03-07
Allow non admin users to manage BSL

### LINKED ISSUE #6563: In freshly created project constantly see an error message that I "don't have admin right"
### What happened?

Dashboard constantly complains that I don't have admin right, in a freshly created project.

The request that returns such result is: `/api/v2/projects/<PROJECT_ID>/clusterbackupstoragelocation`

### Expected behavior

No nonsense error messages that I I as as user have nothing to do with.

Login to dashboard.

- KKP version: v2.25.0-beta.1
- Shared or separate master/seed clusters?: both

not applicable

Any of them

Any of them

ahmadhamzh:
Allow non admin users to manage backup storage location.

Fixes #6563 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/ee/clusterbackup/storage-location/handler.go
+++ b/modules/api/pkg/ee/clusterbackup/storage-location/handler.go
@@ -102,7 +102,7 @@ func ListCBSL(ctx context.Context, request interface{}, userInfoGetter provider.
 		return nil, utilerrors.NewBadRequest("invalid request")
 	}
 
-	_, err := userInfoGetter(ctx, req.ProjectID)
+	user, err := userInfoGetter(ctx, req.ProjectID)
 
 	if err != nil {
 		return nil, err
@@ -112,7 +112,7 @@ func ListCBSL(ctx context.Context, request interface{}, userInfoGetter provider.
 		kubermaticv1.ProjectIDLabelKey: req.ProjectID,
 	}
 
-	cbslList, err := provider.ListUnsecured(ctx, labelSet)
+	cbslList, err := provider.List(ctx, user, labelSet)
 	if err != nil {
 		return nil, err
 	}
@@ -135,7 +135,7 @@ func GetCSBL(ctx context.Context, request interface{}, userInfoGetter provider.U
 		return nil, utilerrors.NewBadRequest("invalid request")
 	}
 
-	_, err := userInfoGetter(ctx, req.ProjectID)
+	user, err := userInfoGetter(ctx, req.ProjectID)
 
 	if err != nil {
 		return nil, err
@@ -145,9 +145,13 @@ func GetCSBL(ctx context.Context, request interface{}, userInfoGetter provider.U
 		kubermaticv1.ProjectIDLabelKey: req.ProjectID,
 	}
 
-	cbsl, err := provider.GetUnsecured(ctx, req.ClusterBackupStorageLocationName, labelSet)
+	cbsl, err := provider.Get(ctx, user, req.ClusterBackupStorageLocationName, labelSet)
 	if err != nil {
 		return nil, err
+	}
+
+	if cbsl == nil {
+		return nil, nil
 	}
 
 	return &apiv2.ClusterBackupStorageLocation{
@@ -170,17 +174,13 @@ func CreateCBSL(ctx context.Context, request interface{}, userInfoGetter provide
 		return nil, err
 	}
 
-	if user.Roles.Has("viewers") {
-		return nil, fmt.Errorf("user with a viewer role is not permitted to create storage locations.")
-	}
-
 	cbslName := req.Body.Name
 	cbslSpec := req.Body.CBSLSpec.DeepCopy()
 	creds := req.Body.Credentials
 	cbsl := &kubermaticv1.ClusterBackupStorageLocation{
 		Spec: *cbslSpec,
 	}
-	created, err := provider.CreateUnsecured(ctx, cbslName, req.ProjectID, cbsl, creds)
+	created, err := provider.Create(ctx, user, cbslName, req.ProjectID, cbsl, creds)
 	if err != nil {
 		return nil, err
 	}
@@ -205,11 +205,7 @@ func DeleteCBSL(ctx context.Context, request interface{}, userInfoGetter provide
 		return err
 	}
 
-	if user.Roles.Has("viewers") {
-		return fmt.Errorf("user with a viewer role is not permitted to delete storage locations.")
-	}
-
-	err = provider.DeleteUnsecured(ctx, req.ClusterBackupStorageLocationName)
+	err = provider.Delete(ctx, user, req.ClusterBackupStorageLocationName)
 	if err != nil {
 		return err
 	}
@@ -223,23 +219,23 @@ func PatchCBSL(ctx context.Context, request interface{}, userInfoGetter provider
 		return nil, utilerrors.NewBadRequest("invalid request")
 	}
 
-	cbslSpec := req.Body.CBSLSpec.DeepCopy()
-	cbsl := &kubermaticv1.ClusterBackupStorageLocation{
-		Spec: *cbslSpec,
-	}
-	patched, err := provider.PatchUnsecured(ctx, req.ClusterBackupStorageLocationName, cbsl, req.Body.Credentials)
-	if err != nil {
-		return nil, err
-	}
-
 	user, err := userInfoGetter(ctx, req.ProjectID)
 
 	if err != nil {
 		return nil, err
 	}
 
-	if user.Roles.Has("viewers") {
-		return nil, fmt.Errorf("user with a viewer role is not permitted to edit storage locations.")
+	cbslSpec := req.Body.CBSLSpec.DeepCopy()
+	cbsl := &kubermaticv1.ClusterBackupStorageLocation{
+		Spec: *cbslSpec,
+	}
+	patched, err := provider.Patch(ctx, user, req.ClusterBackupStorageLocationName, cbsl, req.Body.Credentials)
+	if err != nil {
+		return nil, err
+	}
+
+	if err != nil {
+		return nil, err
 	}
 
 	return &apiv2.ClusterBackupStorageLocation{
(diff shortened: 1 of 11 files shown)


modules/api/pkg/ee/clusterbackup/storage-location/handler.go
```diff
@@ -96,11 +96,18 @@ const (
...
 	if !ok {
 		return nil, utilerrors.NewBadRequest("invalid request")
 	}
+
+	_, err := userInfoGetter(ctx, req.ProjectID)
+
+	if err != nil {
+		return nil, err
+	}
+
```
SimonTheLeg, line 110: I think this essentially only queries the user roles and so on, but since you are not using them to check for anything (like in CreateCSBL), I believe it can be removed

SimonTheLeg: same for the `GetCSBL` func

modules/api/pkg/ee/clusterbackup/storage-location/handler.go
```diff
@@ -168,21 +193,31 @@ func CreateCBSL(ctx context.Context, request interface{}, provider provider.Back
...
-	err := provider.DeleteUnsecured(ctx, req.ClusterBackupStorageLocationName)
+	user, err := userInfoGetter(ctx, req.ProjectID)
+
+	if err != nil {
+		return err
+	}
+
+	if user.Roles.Has("viewers") {
+		return fmt.Errorf("user with a viewer role is not permitted to delete storage locations.")
+	}
```
SimonTheLeg, line 210: my understanding from other code is that it is possible for a user to have both "viewer" as well as another role. As a result we also need to check if viewers is the only role the user has.

For example like this is done here
https://github.com/kubermatic/dashboard/blob/140d3a4b773b36c18b9669d86fd7f348cdfd388b/modules/api/pkg/provider/kubernetes/cluster.go#L428

SimonTheLeg: Same for Create and Patch as well

SimonTheLeg (changes requested):

ahmadhamzh:
I changed the approach to using an Impersonated Client, so there is no need to check the role in the API.
@SimonTheLeg

ahmadhamzh:
@SimonTheLeg  sorry but this PR is blocked by https://github.com/kubermatic/kubermatic/pull/13174

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -26,77 +26,108 @@ package storagelocation
...
 	listOpts := &ctrlruntimeclient.ListOptions{
 		LabelSelector: labels.SelectorFromSet(labelSet),
 		Namespace:     resources.KubermaticNamespace,
 	}
 
 	cbslList := &kubermaticv1.ClusterBackupStorageLocationList{}
-	if err := p.privilegedClient.List(ctx, cbslList, listOpts); err != nil {
+	client := p.privilegedClient
+
+	if err := client.List(ctx, cbslList, listOpts); err != nil {
```
SimonTheLeg, line 82: nit: just to keep this uniform with other code. Let's go back to the previous p.privilegedClient.List notation from before

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -26,77 +26,108 @@ package storagelocation
...
+	}
+
+	cbslList, err := p.List(ctx, userInfo, labelSet)
 	if err != nil {
-		return nil, fmt.Errorf("failed to list ClusterBackupStorageLocations: %w", err)
+		return nil, err
 	}
 	for _, cbsl := range cbslList.Items {
 		if cbsl.Name == name {
 			return &cbsl, nil
```
SimonTheLeg, line 99: I know this is not part of this PR, but I would appreciate if we could create a ticket to investigate this. Judging from just the code I think it should be fine to create the impersonated client and run a client.Get. 

See https://github.com/kubermatic/dashboard/blob/03648c694cca330f62ad519cac4338a2bd9af6f5/modules/api/pkg/provider/kubernetes/project.go#L128-L135 for an example

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -200,3 +256,14 @@ func (p *BackupStorageProvider) patchCredentials(ctx context.Context, secretName
 	}
 	return p.privilegedClient.Patch(ctx, updated, ctrlruntimeclient.MergeFrom(existing))
 }
+
+func (p *BackupStorageProvider) getImpersonatedClient(userInfo *provider.UserInfo) (ctrlruntimeclient.Client, error) {
+	if userInfo == nil {
+		return nil, errors.New("a user is missing but required")
+	}
```
SimonTheLeg, line 263: I think since you are checking for userInfo to not be nil always before calling this func, this can be removed.

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -110,57 +141,78 @@ func (p *BackupStorageProvider) CreateUnsecured(ctx context.Context, cbslName, p
 		},
 		Key: "cloud-credentials",
```
SimonTheLeg, line 142: I would recommend to move this into a global const, since we are using this in multiple places in this provider

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -26,77 +26,108 @@ package storagelocation
...
+func NewBackupStorageProvider(createMasterImpersonatedClient kubernetes.ImpersonationClient, privilegedClient ctrlruntimeclient.Client) *BackupStorageProvider {
 	return &BackupStorageProvider{
-		privilegedClient: privilegedClient,
+		createMasterImpersonatedClient: createMasterImpersonatedClient,
+		privilegedClient:               privilegedClient,
 	}
 }
 
-func (p *BackupStorageProvider) ListUnsecured(ctx context.Context, labelSet map[string]string) (*kubermaticv1.ClusterBackupStorageLocationList, error) {
+func (p *BackupStorageProvider) List(ctx context.Context, userInfo *provider.UserInfo, labelSet map[string]string) (*kubermaticv1.ClusterBackupStorageLocationList, error) {
```
SimonTheLeg, line 69: In my opinion this should still be called ListUnsecured, because it uses the priviledged client to access the info

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -26,77 +26,108 @@ package storagelocation
...
 		return nil, fmt.Errorf("failed to list ClusterBackupStorageLocations: %w", err)
 	}
 	return cbslList, nil
 }
 
-// func (p *BackupStorageProvider) Get(ctx context.Context, userInfo *provider.UserInfo, name string) (*kubermaticv1.ClusterBackupStorageLocation, error).
-func (p *BackupStorageProvider) GetUnsecured(ctx context.Context, name string, labelSet map[string]string) (*kubermaticv1.ClusterBackupStorageLocation, error) {
-	// we use List() to filter CBSLs with labels easily, to keep scoped into the owner project.
-	cbslList, err := p.ListUnsecured(ctx, labelSet)
+func (p *BackupStorageProvider) Get(ctx context.Context, userInfo *provider.UserInfo, name string, labelSet map[string]string) (*kubermaticv1.ClusterBackupStorageLocation, error) {
```
SimonTheLeg, line 88: In my opinion this should still be called GetUnsecured, because it calls the List which in turn uses the priviledged client.

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -110,57 +141,78 @@ func (p *BackupStorageProvider) CreateUnsecured(ctx context.Context, cbslName, p
...
+	client := p.privilegedClient
+
 	cbsl := &kubermaticv1.ClusterBackupStorageLocation{}
-	if err := p.privilegedClient.Get(ctx, types.NamespacedName{Name: name, Namespace: resources.KubermaticNamespace}, cbsl); err != nil {
+	if err := client.Get(ctx, types.NamespacedName{Name: name, Namespace: resources.KubermaticNamespace}, cbsl); err != nil {
 		if apierrors.IsNotFound(err) {
 			return nil
 		}
 		return err
 	}
```
SimonTheLeg, line 175: Is there a reason why the priviledged client is needed for getting the ClusterBackupStorageLocation? Couldn't the impersonated client be used for this?

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
```diff
@@ -110,57 +141,78 @@ func (p *BackupStorageProvider) CreateUnsecured(ctx context.Context, cbslName, p
...
+		return nil, errors.New("a user is missing but required")
+	}
+
+	client := p.privilegedClient
+
 	existing := &kubermaticv1.ClusterBackupStorageLocation{}
-	if err := p.privilegedClient.Get(ctx, types.NamespacedName{Name: cbslName, Namespace: resources.KubermaticNamespace}, existing); err != nil {
+	if err := client.Get(ctx, types.NamespacedName{Name: cbslName, Namespace: resources.KubermaticNamespace}, existing); err != nil {
 		return nil, err
 	}
```
SimonTheLeg, line 201: same here. Is the priviledged client required, or could we just use the impersonated client for getting it?

SimonTheLeg (changes requested):
Overall I think this should all work and is in spec what we are doing in other parts of the API. I have some questions regarding the usage of the privileged client in some places. ptal. 
But overall I think we are on the right track.

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
moelsayed: refactored to apply use the same client for both operations based on the user role.

modules/api/pkg/ee/clusterbackup/storage-location/provider.go
moelsayed: done

PR #6621, opened 2024-03-21
[release/v2.23] update changelog.json

xrstf:
**What this PR does / why we need it**:
This updates the link to the changelog.

**Does this PR introduce a user-facing change? Then add your Release Note here**:
```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

PR #6631, opened 2024-03-25
Remove changelog dialog

ahmedwaleedmalik:
Based on the decisions made in the following PRs by sig-release; they want to get rid of changelogs:
- https://github.com/kubermatic/dashboard/pull/6621
- https://github.com/kubermatic/dashboard/pull/6620

We are removing the changelog dialog and all the mechanisms around it in favor of just an external URL that points to relevant release notes. This is being done specifically because from a UI perspective we can't rely on what someone "feels" like when preparing new releases for the dashboard; features on UI need to be consistent. So you either update these changelogs consistently for each release or none at all. Since in the latest releases the decision was made that sig-release is not going to update those changelogs, we are removing this feature.

Fixes #

**What type of PR is this?**

/kind cleanup

```release-note
The dialog for changelog has been removed in favor of an external URL that points to relevant change logs.
```

**Documentation**:

```documentation
TBD
```


Waseem826 (approved):
/approve

PR #6647, opened 2024-04-11
Upgrade golang to 1.22.2

akash-gautam:

Fixes #

**What type of PR is this?**
/kind chore

```release-note
Update to Go 1.22.2
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/package.json
+++ b/modules/web/package.json
@@ -57,7 +57,7 @@
     ]
   },
   "overrides": {
-    "socks": "2.7.4"
+    "socks": "2.7.3"
   },
   "dependencies": {
     "@angular/animations": "17.3.1",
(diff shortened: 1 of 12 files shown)


embik:
Note: You'll probably want to do this for `release/v2.25` and the respective post-submit updates also need this: https://github.com/kubermatic/infra/tree/main/prow/jobs/dashboard.

modules/web/package.json
```diff
@@ -57,7 +57,7 @@
     ]
   },
   "overrides": {
-    "socks": "2.7.3"
+    "socks": "2.7.4"
```
embik, line 60: Why change this?

embik: Okay, sig-api-ui needs to be the judge of that then :D

ahmedwaleedmalik: Please revert this change

ahmedwaleedmalik (approved):
/approve

PR #6651, opened 2024-04-15
api: don't construct upgrade conditions, use KKP logic instead

embik:

It looks like the API code never used the logic provided by KKP to determine conditions that are used in decision-making on possible upgrade paths. This was changed in https://github.com/kubermatic/kubermatic/pull/13046, but the old logic was also not called (its logic seemed to be copy-pasted into the API code though), so the drift wasn't detected.

This PR removes the logic to determine conditions from the API code and instead just re-uses `k8c.io/kubermatic/v2/pkg/version/clusterGetVersionConditions`, as the KKP code does.

Fixes #

**What type of PR is this?**
/kind bug
/kind regression

```release-note
Resolve conflict in determining available Kubernetes versions where upgrades where possible in `Cluster` object but not via the Dashboard
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #6690, opened 2024-05-22
Adjust the preset domain field to accept emails

### LINKED ISSUE #6686: Preset .spec.requiredEmails is not working with full email address
### What happened

When a preset is configured to use `.spec.requiredEmails` with full email address, it is not working as expected. The preset is visible during the cluster creation to other end users too instead of just to the user whose email address was configured.

### Expected behavior

Preset configured with `.spec.requiredEmails` with full email address should be visible to only those end user whose email address is part of the email address list. 

Also documented in the [presetSpec](https://docs.kubermatic.com/kubermatic/v2.25/references/crds/#presetspec) for `requiredEmails` field as follows: 
> RequiredEmails is a list of e-mail addresses that this presets should be restricted to. Each item in the list can be either a full e-mail address or just a domain name. This restriction is only enforced in the KKP API.

### How to reproduce

Configure the preset with `.spec.requiredEmails`  providing some full email address instead of email domain.  
Login to KKP dashboard using another account/email address which is not part of `.spec.requiredEmails` and from Clusters screen the user can see and select that  preset during the cluster creation from the settings tab. 

### Environment
- UI Version:
- API Version:
- Domain:
- Others:

ahmadhamzh:
adjust the domain field to accept emails in the create preset dialog 

* create preset dialog:
image
 
Fixes #6686

**What type of PR is this?**
/kind bug

```release-note
Adjust the preset domain field to accept emails.
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/presets/dialog/steps/preset/component.ts
+++ b/modules/web/src/app/settings/admin/presets/dialog/steps/preset/component.ts
@@ -17,7 +17,7 @@ import {FormBuilder, NG_VALIDATORS, NG_VALUE_ACCESSOR, Validators} from '@angula
 import {PresetDialogService} from '@app/settings/admin/presets/dialog/steps/service';
 import {BaseFormValidator} from '@shared/validators/base-form.validator';
 import {takeUntil} from 'rxjs/operators';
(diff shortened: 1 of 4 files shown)


Waseem826 (approved):
/approve

PR #6697, opened 2024-05-27
bump KKP to eec75550ae15

xrstf:
This makes 1.30 available on dev :)

Since it was removed upstream, this PR removes the `EnableOperatingSystemManager` field entirely from the codebase.

**What type of PR is this?**
/kind chore

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (changes requested):
> The intent was to allow the removal of this flag within the API to happen in conjunction with removal from the UI.

We have already removed the option to enable/disable it from UI. So we should simply drop the mappings from API as well; instead of `nolint` just drop the field/mappins

ahmedwaleedmalik (approved):
/approve

PR #6706, opened 2024-05-31
fix dev deployment

xrstf:
Now that `spec.api.dockerTag` is a thing, it also needs to be set explicitly when deploying to dev.

**What type of PR is this?**
/kind chore

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve

PR #6713, opened 2024-06-05
Add a "Backup All Namespaces" Checkbox

### LINKED ISSUE #6700: Refine backup storage location settings in the user journey
Remarks in current behaviour:

1. When adding cluster backup feature to the cluster which is being created in cluster creation Step 2, currently you have to define backup storage location at cluster creation time.
2. When later trying to create a backup to this cluster, you have a field for cluster storage location, but it is inactive, cannot be changed plus it also has a different name then the storage location I chose in Step 2 for the cluster.
3. It can be frustrating to select all namespaces if you always want to just backup the whole cluster

Acceptance criteria:

1. Backup locations should be able to vary backup by backup, therefore I would not necessarily order one final storage location to the particular cluster at deploying time in cluster creation Step 2, so skip this from the workflow
2. Then I can select a backup location when creating the backup itself and the field won’t be inactive and also should not having that default naming, instead we should list the available backup storage locations there (as they were listed in Step 2 of cluster creation), and the cluster admin should be able to chose it in the backup creation time
3. Add a Select All option to the namespaces when creating a backup

ahmadhamzh:
add a check box to select all namespaces, and show the name of the `CBSL`  that was used during cluster creation.

image

image

image

Fixes #6700 

**What type of PR is this?**
/kind design

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/add-dialog/component.ts
+++ b/modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/add-dialog/component.ts
@@ -25,10 +25,10 @@ import {UserService} from '@core/services/user';
 import {Observable, Subject, takeUntil} from 'rxjs';
 import {RBACService} from '@core/services/rbac';
 import {
+  BackupStorageLocationTempName,
   BackupType,
   ClusterBackup,
   CreateClusterBackupSchedule,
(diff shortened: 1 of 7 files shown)


ahmedwaleedmalik (approved):
/approve

PR #6727, opened 2024-07-05
Adjust chip list component to have multiple values at once

### LINKED ISSUE #6684: Allow entering comma-separated list of IP CIDRs in the network configuration allowlists
Currently, it is only possible to add IP CIDRs one by one during Cluster creation, adding a comma-separated list results in a validation error.
image
Having a big amount of CIDRs this can be a tedious task.

It would be a good UX improvement, to allow adding a comma-separated list of IPs and append them in the allowlist fields.

ahmadhamzh:
Adjust the km-chip-list component so that the user can add multiple values separated by commas or spaces all at once.

Fixes #6684

**What type of PR is this?**
/kind cleanup
/kind design

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/shared/components/chip-list/component.ts
+++ b/modules/web/src/app/shared/components/chip-list/component.ts
@@ -58,7 +58,7 @@ export class ChipListComponent implements OnChanges, OnDestroy, ControlValueAcce
   readonly controls = Controls;
   @Input() title: string;
   @Input() label: string;
-  @Input() description = 'Use comma, enter or space key as the separator.';
+  @Input() description = 'Use comma, space or enter key as the separator.';
   @Input() placeholder: string;
   @Input() disabled: boolean;
   @Input('kmRequired') required: boolean;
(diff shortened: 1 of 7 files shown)


modules/web/src/app/shared/components/chip-list/component.ts
```diff
@@ -112,7 +112,14 @@ export class ChipListComponent implements OnChanges, OnDestroy, ControlValueAcce
     }
 
     event.chipInput?.clear();
-    this.tags.push(value);
+    const delimiters = [',', ' '];
+    const splitRegex = new RegExp(`[${delimiters.join('')}]`);
+    if (splitRegex.test(value)) {
+      const multiTags = value.split(splitRegex);
+      multiTags.forEach(val => this.tags.push(val.trim()));
```
Waseem826, line 119: If both `,` and ` ` appear together in the string then it `split` will give an extra empty value . For example, `123, 456` would return `['123', '', '456']`. So we should add a filter on splitted values:
```suggestion
      const multiTags = value.split(splitRegex);
      multiTags.filter(val => !!val.trim()).forEach(val => this.tags.push(val.trim()));
```

ahmedwaleedmalik (approved):
Other than waseem's comment, looks good 👍 
/hold 
to resolve the feedback

PR #6728, opened 2024-07-05
remove scary deprecation warning

xrstf:
While technically correct, the warning about the API stability is causing uncertainty with integrators. In practice the KKP REST API is not going away anytime soon, especially not in KKP 2.x, given KKP's architecture and inherent limitations.

CRDs are still the way to go for "simple" setups or setups where a singular entity is managing KKP (like when you have a GitOps pipeline and only scripts modify KKP directly, instead of all members in a team individually connecting to KKP). But I can see how my strong stance on deprecating and de-emphasizing the REST API might have caused more friction than I intended.

/cc @moadqassem 

**What type of PR is this?**
/kind cleanup

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


moadqassem (approved):
/approve
/lgtm

PR #6735, opened 2024-07-10
Add static labels option to the admin panel 

### LINKED ISSUE #6724: Static labels for clusters

As a KKP admin, I'd like to configure a static list of labels somewhere that can then be used for the clusters.

This list will consist of keys and multiple values(options) against them which can then be configured by the KKP admin/user during cluster creation. When static labels are configured, the input for Labels should be drop downs for key and value.

Screenshot 2024-07-01 at 12 11 22

It'd be also important to configure default and protected labels:
1. Default: Applied on the cluster by default
2. Protected: Once applied they cannot be deleted

ahmadhamzh:
add static labels in the admin defaults page, these labels are only for clusters and can be used in the cluster creation wizard 
**admin settings:**

image

**create cluster:**
image

**edit cluster:**

image

Fixes #6724

**What type of PR is this?**
/kind feature

```release-note
add new static labels option in admin settings.
```

```documentation
TBD
```


modules/web/src/app/settings/admin/defaults/static-labels-form/template.html
```diff
@@ -0,0 +1,62 @@
...
+      <km-chip-list class="static-labels-values"
+                    label="Values"
+                    placeholder="Values..."
+                    isStatickLabel="true"
+                    [formControlName]="Controls.Values">
+      </km-chip-list>
+      <mat-checkbox fxFlex="13"
+                    [formControlName]="Controls.Default"
+                    kmValueChangedIndicator>Default
+      </mat-checkbox>
```
Waseem826, line 48: Should we check and disable this when "Protected" is selected?

modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts
```diff
@@ -0,0 +1,156 @@
...
+      [Controls.StaticLabels]: this._formBuilder.array([]),
+    });
+  }
+
+  private _addStaticLabel(
+    key: string = '',
+    values: string[] = [],
+    isProtected: boolean = false,
+    isDefault: boolean = false
+  ): void {
```
Waseem826, line 109: Could be simplified to:
```suggestion
  private _addStaticLabel(label?: StaticLabel): void {
```

modules/web/src/app/settings/admin/defaults/component.ts
```diff
@@ -156,13 +156,29 @@ export class DefaultsComponent implements OnInit, OnDestroy {
...
   }
 
+  private _checkLabels(staticLabels: StaticLabel[]): boolean {
+    let correct = true;
+    staticLabels.forEach(label => {
+      if (!label.key?.length || !label.values?.length) {
+        correct = false;
+      }
+    });
+    return correct;
```
Waseem826, line 180: Could be simplified to:
```suggestion
    return staticLabels.every(label => label.key?.length &&  label.values?.length));
```

modules/web/src/app/shared/components/label-form/template.html
```diff
@@ -70,14 +79,22 @@
...
+            </mat-option>
+          </mat-autocomplete>
           <mat-error *ngIf="label.get('value').errors?.validLabelValuePattern"> {{valueName}} not allowed. </mat-error>
           <mat-error *ngIf="label.get('value').errors?.validLabelValueLength"> {{valueName}} is too long. </mat-error>
         </mat-form-field>
         <button mat-icon-button
                 type="button"
                 class="delete-button"
-                [disabled]="!isRemovable(i)"
+                [disabled]="!isRemovable(i) || isProtectedKey(label.get('key').value, i)"
```
Waseem826, line 97: IMO we should move `isProtectedKey(label.get('key').value, i)` inside the `isRemovable` method.

modules/web/src/assets/css/material/_main.scss
```diff
@@ -447,6 +447,10 @@ mat-form-field {
   }
 }
 
+.static-labels-values mat-form-field.km-node-tags .mat-mdc-text-field-wrapper .mdc-floating-label--float-above{
+  top: 30px !important;
+}
```
Waseem826, line 452: Why do we need this? Can you please share a screenshot?

Waseem826 (changes requested):
Some feedback. PTAL

modules/web/src/assets/css/material/_main.scss
ahmadhamzh: this is a pic without this
image

but i check it again and maybe 28px would look better
image

modules/web/src/app/shared/components/label-form/template.html
```diff
@@ -70,14 +79,22 @@
...
                 (click)="deleteLabel(i)">
```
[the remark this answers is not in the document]
ahmadhamzh: i thought it would be already disabled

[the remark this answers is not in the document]
Waseem826: Yes but someone can remove these HTML class/attributes.

modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts
```diff
@@ -0,0 +1,159 @@
...
+export class StaticLabelsFormComponent implements OnInit, OnDestroy {
+  private _unsubscribe = new Subject<void>();
+  readonly Controls = Controls;
+
+  form: FormGroup;
+  @Input() staticLabels: StaticLabel[] = [];
+  asyncLabelValidators = [AsyncValidators.RestrictedLabelKeyName(ResourceType.Cluster)];
+  @Output() staticLabelsChange = new EventEmitter<StaticLabel[]>();
+
+  get statickLabelArray(): FormArray {
```
Waseem826, line 45: Sorry I didn't notice this previously:
```suggestion
  get staticLabelArray(): FormArray {
```

modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  constructor(private readonly _formBuilder: FormBuilder) {}
+
+  ngOnInit(): void {
+    this._initForm();
+
+    this.staticLabels?.forEach(label => {
+      this._addStaticLabel(label);
+    });
+    this._addStaticLabel(EMPTY_STATIC_LABELS);
```
Waseem826, line 57: Why do we need `EMPTY_STATIC_LABELS` instead of defaulting the values in `_addStaticLabel` method using `||`? For example,

```
[Controls.Values]: [label.values || []],
[Controls.Protected]: [label.protected || false],
[Controls.Default]: [label.default || false],

File: modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts [hunk 9]
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  isProtected(val: boolean, index: number): boolean {
+    const defaultControl = this.statickLabelArray.at(index).get(Controls.Default);
+    if (val && defaultControl.value) {
+      this.statickLabelArray.at(index).get(Controls.Default).setValue(false);
+    }
+    return val;
+  }
```
Waseem826, line 97: This method is semantically wrong, its returning the same value it got as input. If you want to modify the value of "Default" control, then please add a `change` event on the "Protected" checkbox e.g. `(change)="onProtectedValueChange($event, i)"`.
Also, shouldn't we set the value of "Default" to `true` when "Protected" is `true`?

modules/web/src/app/shared/components/chip-list/component.ts
```diff
@@ -64,6 +65,7 @@ export class ChipListComponent implements OnChanges, OnDestroy, ControlValueAcce
   @Input('kmRequired') required: boolean;
   @Input('kmPatternError') patternError = 'Invalid pattern';
   @Input('kmPattern') pattern: string;
+  @Input() isStatickLabel = false;
```
Waseem826, line 68:
```suggestion
  @Input() isStaticLabel = false;
```

modules/web/src/app/shared/components/label-form/component.ts
```diff
@@ -221,16 +233,38 @@ export class LabelFormComponent implements OnChanges, OnInit, OnDestroy, Control
...
     // Add initial label for the user.
     this._addLabel();
+    this._updateLabelsObject();
   }
 
   isRemovable(index: number): boolean {
-    return index < this.labelArray.length - 1 && !this._isInherited(Object.keys(this.labels)[index]);
+    return (
+      (index < this.labelArray.length - 1 && !this._isInherited(Object.keys(this.labels)[index])) ||
+      this.isProtectedKey(this.labelArray.controls[index].get('key').value, index)
```
Waseem826, line 244: Shouldn't this be the other way around? i.e. label is removable when its _not_ protected?
```suggestion
      !this.isProtectedKey(this.labelArray.controls[index].get('key').value, index)
```

Waseem826 (changes requested):
Some feedback after latest changes. PTAL

Waseem826 (approved):

Waseem826:
Approved from my side. @ahmedwaleedmalik do you have some feedback? If not please take a look at API changes and approve.

ahmedwaleedmalik (approved):
/approve
/retest

Looks good overall, please create a release testing ticket for this. You can assign it to me.

PR #6782, opened 2024-08-28
Update baremetal provider config

mohamed-rafraf:
This PR will update  the MC baremetal driverspec in the API baremetal provider config

Fixes #

**What type of PR is this?**

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


modules/api/go.mod
```diff
@@ -70,7 +70,7 @@ require (
 	google.golang.org/api v0.187.0
 	gopkg.in/yaml.v3 v3.0.1
 	k8c.io/kubeone v1.7.3
-	k8c.io/kubermatic/v2 v2.25.1-0.20240823163015-db36dbe1818f
+	k8c.io/kubermatic/v2 v2.25.1-0.20240828091320-045f90ef8db4
```
ahmedwaleedmalik, line 73: This PR doesn't "This PR will update the MC package" so please remove that from title and PR description.

modules/api/pkg/resources/machine/common.go
```diff
@@ -273,30 +274,48 @@ func getVSphereProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc
...
 }
 
-func getBaremetalProviderSpec(cluster *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*runtime.RawExtension, error) {
-	// Check if Tinkerbell configuration is provided
+func GetBaremetalProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*baremetal.RawConfig, error) {
+	// Ensure Tinkerbell configuration is provided
 	if nodeSpec.Cloud.Baremetal.Tinkerbell == nil {
-		return nil, errors.New("tinkerbell provisioner is required")
+		return nil, errors.New("tinkerbell provisioner is required in baremetal configuration")
 	}
```
ahmedwaleedmalik, line 281: What if `nodeSpec.Cloud.Baremetal` is nil? This would result in a null exception. Please update this condition.

modules/api/pkg/resources/machine/common.go
```diff
@@ -273,30 +274,48 @@ func getVSphereProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc
...
 
-	// Create TinkerbellNodeSpec from nodeSpec
-	tinkerbellSpec := apiv1.TinkerbellNodeSpec{
-		OsImageUrl:  nodeSpec.Cloud.Baremetal.Tinkerbell.OsImageUrl,
-		HardwareRef: nodeSpec.Cloud.Baremetal.Tinkerbell.HardwareRef,
+	tinkerbellConfig := nodeSpec.Cloud.Baremetal.Tinkerbell
+
+	// Validate required fields in Tinkerbell configuration
+	if tinkerbellConfig.HardwareRef.String() == "/" || tinkerbellConfig.OsImageUrl == "" {
```
ahmedwaleedmalik, line 286: > tinkerbellConfig.HardwareRef.String() == "/"

Is "/" the default value for this field? Could there be a case where this value is nil/empty?

ahmedwaleedmalik, line 286: Would be nice if we can separate the `If` conditions out. For instance, if someone doesn't provide OsImageUrl but does provider HardwareRef, they'll still get an error stating "HardwareRef and OsImageUrl must be provided".

modules/api/pkg/resources/machine/common.go
```diff
@@ -273,30 +274,48 @@ func getVSphereProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc
 	return EncodeAsRawExtension(config)
 }
 
-func getBaremetalProviderSpec(cluster *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*runtime.RawExtension, error) {
-	// Check if Tinkerbell configuration is provided
+func GetBaremetalProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*baremetal.RawConfig, error) {
```
ahmedwaleedmalik, line 277: `dc *kubermaticv1.Datacenter` is not being used anywhere can you please either drop that argument(preferred) or update it to `_ *kubermaticv1.Datacenter`.

modules/api/pkg/resources/machine/common.go
```diff
@@ -273,30 +274,48 @@ func getVSphereProviderSpec(c *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc
...
+	tinkerbellData, err := json.Marshal(tinkerbellSpec)
 	if err != nil {
-		return nil, fmt.Errorf("failed to marshal TinkerbellNodeSpec: %w", err)
+		return nil, fmt.Errorf("failed to marshal TinkerbellPluginSpec: %w", err)
 	}
 
 	// Create a new baremetal.RawConfig and set the DriverSpec
 	config := &baremetal.RawConfig{
 		DriverSpec: runtime.RawExtension{Raw: tinkerbellData},
+		Driver:     providerconfig.ConfigVarString{Value: "tinkerbell"},
```
ahmedwaleedmalik, line 307: Why is the value being hardcoded here? Can't we use a constant from machine-controller for this?

modules/api/pkg/resources/machine/common.go
ahmedwaleedmalik, line 277:
```suggestion
func getBaremetalProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.NodeSpec, dc *kubermaticv1.Datacenter) (*baremetal.RawConfig, error) {
```

Do we need to expose this method? Seems like it's not being used anywhere out of this file?

ahmedwaleedmalik (changes requested):

modules/api/pkg/resources/machine/common.go
Waseem826: I've marked both Hardware Ref Name and Hardware Ref Namespace fields as mandatory on the UI. Are these optional?

ahmedwaleedmalik:
For `pre-dashboard-api-e2e` the failing test `TestCreateUpdateHetznerCluster/create_cluster_on_Hetzner` is most likely a flake. I re-ran the e2e tests on your PR. If the test still fails there and the problem persists please share this in development channel in slack; we are most likely hitting our quotas on Hetzner.

modules/api/pkg/resources/machine/common.go
mohamed-rafraf: "/" is the default value for HardwareRef.String()

mohamed-rafraf: they are not optional. I am checking if `HardwareRef` is empty to return an error. 

> `HardwareRef` is empty means: HardwareRef.String() == "/"

modules/api/pkg/resources/machine/common.go
mohamed-rafraf: we are checking this field before calling the function.

ahmedwaleedmalik (approved):
/approve

PR #6783, opened 2024-08-29
Enable editing allowed IP ranges for NodePorts

### LINKED ISSUE #6772: Enable editing allowed IP ranges for NodePorts

As a KKP User cluster owner
I would like to edit allowed IP ranges for NodePorts in the UI
in order to have flexible IP ranges.

### Solution details

- Add an editing option in the Edit Cluster menu
- OpenStack is the prioritized provider

### Use cases

- IP range was enlarges, or a second IP range was added, because the original IP range was becoming too small. 
- IPs could just change for many reasons - and maybe only a single host (/32) was granted access.

### Additional information

- Currently IP range setting reconciles on OpenStack and Azure at least, which makes it impossible to customize them from the provider without having an editing option on the KKP side

ahmadhamzh:
add new option to edit the allowed ip ranges in the edit cluster dialog for the supported providers.
**edit cluster dialog**
image

Fixes #6772

**What type of PR is this?**
/kind feature

```release-note
Enable editing allowed IP ranges for NodePorts 
```
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -18,6 +18,7 @@ import {MatDialogRef} from '@angular/material/dialog';
 import {ClusterBackupService} from '@app/core/services/cluster-backup';
 import {DynamicModule} from '@app/dynamic/module-registry';
 import {BackupStorageLocation} from '@app/shared/entity/backup';
+import {NODEPORTS_IPRANGES_SUPPORTED_PROVIDERS} from '@app/shared/model/NodeProviderConstants';
 import {BSLListState} from '@app/wizard/step/cluster/component';
 import {ClusterService} from '@core/services/cluster';
 import {DatacenterService} from '@core/services/datacenter';
@@ -70,6 +71,7 @@ enum Controls {
   DisableCSIDriver = 'disableCSIDriver',
   ClusterBackup = 'clusterBackup',
   BackupStorageLocation = 'backupStorageLocation',
+  NodePortsAllowedIPRanges = 'nodePortsAllowedIPRanges',
 }
 
 @Component({
@@ -102,6 +104,7 @@ export class EditClusterComponent implements OnInit, OnDestroy {
   isCSIDriverDisabled = false;
   backupStorageLocationsList: BackupStorageLocation[];
   backupStorageLocationLabel: BSLListState = BSLListState.Ready;
+  isAllowedIPRangeSupported: boolean;
   readonly isEnterpriseEdition = DynamicModule.isEnterpriseEdition;
   readonly CLUSTER_DEFAULT_NODE_SELECTOR_NAMESPACE = CLUSTER_DEFAULT_NODE_SELECTOR_NAMESPACE;
   readonly CLUSTER_DEFAULT_NODE_SELECTOR_TOOLTIP = CLUSTER_DEFAULT_NODE_SELECTOR_TOOLTIP;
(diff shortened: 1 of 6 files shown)


modules/web/src/app/shared/model/NodeProviderConstants.ts
```diff
@@ -35,7 +35,12 @@ export enum NodeProvider {
   EDGE = 'edge',
   NONE = '',
 }
-
+export const PROVIDERS_SUPPORT_IPRANGES_NODEPORTS = [
```
Waseem826, line 38: Perhaps this is better:
```suggestion
export const NODEPORTS_IPRANGES_SUPPORTED_PROVIDERS = [
```

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -363,6 +374,11 @@ export class EditClusterComponent implements OnInit, OnDestroy {
     }
   }
 
+  isAllowedIPRangeSupported(): boolean {
```
Waseem826, line 377: This can also be just a field instead of method. It will provide minor performance benefit since we won't be searching through the array on every change detection.

Waseem826 (commented):
Some feedback. PTAL

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -222,12 +226,19 @@ export class EditClusterComponent implements OnInit, OnDestroy {
...
           this.isKubeLBEnforced = !!datacenter.spec.kubelb?.enforced;
           this.isCSIDriverDisabled = datacenter.spec.disableCsiDriver;
+          this._provider = datacenter.spec.provider;
         })
       )
       .pipe(switchMap(_ => this._datacenterService.seedSettings(this.datacenter.spec.seed)))
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(seedSettings => (this._seedSettings = seedSettings));
 
+    if (this.cluster.spec.cloud?.[this._provider]?.nodePortsAllowedIPRanges?.cidrBlocks?.length) {
```
Waseem826, line 236: We initialize `this._provider` inside `getDatacenter` subscription and this code block will run before it, hence `this._provider` should be undefined. Are you sure this works as expected?

ahmadhamzh: well it is working as expected but yes it should be inside tab pipe

modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -167,6 +167,9 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
   backupStorageLocationsList: BackupStorageLocation[];
   backupStorageLocationLabel: BSLListState = BSLListState.Ready;
   provider: NodeProvider;
+  isAllowedIPRangeSupported: boolean = NODEPORTS_IPRANGES_SUPPORTED_PROVIDERS.includes(
+    this._clusterSpecService.provider
+  );
```
Waseem826, line 172: We should keep the method in this component because user can go back in the wizard and change the provider.

Waseem826: My previous comment was on the edit-cluster component.

ahmadhamzh: oops :sweat_smile:  sorry for that

Waseem826 (approved):
/approve

Issue #6458, opened 2024-01-10
KKP admin setting for making applications enforced or default


csengerszabo:
KKP Admin can select the set of applications in the Defaults tab which will

- either be defaulted- settings can be changed or app can be removed during UC creation
- or enforced if the checkbox is filled - settings cannot be changed and app cannot be removed during UC creation).

https://github.com/kubermatic/dashboard/assets/18055937/3a92c655-1646-4f7c-8293-59d4f7798f56

PR #6787, opened 2024-08-30
Admin panel settings for applications

ahmedwaleedmalik:
I've broken down the implementation of default/enforced applications further into two PRs. This is the first of the series; it takes care of the "admin's perspective" and adds a dedicated view in the UI to manage applications. Since right now the use cases are limited to defaulting and enforcement, I've intentionally not expanded the options to include edit, delete, etc.

* Admins can now manage applications using the admin panel
* Application can be marked as default or enforced using the dashboard

Towards #6458

**What type of PR is this?**

```release-note
* Admins can now manage applications using admin panel
* Application can be marked as default or enforced using dashboard
```

**Documentation**:

```documentation
TBD
```


modules/web/src/app/settings/admin/nav/template.html
```diff
@@ -106,6 +106,15 @@
         </a>
       </div>
 
+      <div class="sub-menu-item"
```
Waseem826, line 109: We should also add this entry for collapsed side nav.

ahmedwaleedmalik: Good point, sorry I completely overlooked it.

Waseem826 (approved):
/approve

Issue #6776, opened 2024-08-27
Static labels for clusters - Test Release 2.26


ahmadhamzh:
#6724

ahmedwaleedmalik:
Few issues that I found:

## 1. If I refresh the admin panel, the static labels are not displayed

<img width="1484" alt="Screenshot 2024-09-18 at 11 38 16" src="https://github.com/user-attachments/assets/d498077a-2079-4e1a-8201-f1ccdc68d117">

## 2. When we add multiple values against a key. There is no dropdown displayed indicating that there are multiple options available for this field, the UI needs to be improved here

<img width="1484" alt="Screenshot 2024-09-18 at 11 39 13" src="https://github.com/user-attachments/assets/b255eb61-977d-4145-8e4c-46612d90c733">

## 3. The tooltip needs to be more descriptive in explaining what default and protected labels are

<img width="247" alt="Screenshot 2024-09-18 at 11 43 24" src="https://github.com/user-attachments/assets/71de16cc-8e83-4fc4-8a85-5dddc91a76e7">


cc @ahmadhamzh

ahmedwaleedmalik:
The enhancements by @ahmadhamzh have fixed most of the issues https://github.com/kubermatic/dashboard/pull/6891. I already tested this at my end.

PR #6891, opened 2024-09-27
static labels enhancement

ahmadhamzh:

[Screencast from 27 أيلول, 2024 +03 10:38:50 ص.webm](https://github.com/user-attachments/assets/3205d4d3-9b4d-43a7-84bf-1537a6c34c67)

image

[Screencast from 27 أيلول, 2024 +03 10:44:53 ص.webm](https://github.com/user-attachments/assets/1d28d6aa-210b-4c79-8e91-73a5abb4c77b)

ref #6776 

**What type of PR is this?**
/kind bug

```release-note
NONE
```

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts
+++ b/modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Component, EventEmitter, Input, OnDestroy, OnInit, Output} from '@angular/core';
+import {Component, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges} from '@angular/core';
 import {FormArray, FormBuilder, FormGroup, Validators} from '@angular/forms';
 import {ResourceType} from '@app/shared/entity/common';
 import {StaticLabel} from '@app/shared/entity/settings';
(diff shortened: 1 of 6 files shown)


modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts
```diff
@@ -51,11 +52,13 @@ export class StaticLabelsFormComponent implements OnInit, OnDestroy {
   ngOnInit(): void {
     this._initForm();
 
-    this.staticLabels?.forEach(label => {
-      this._addStaticLabel(label);
-    });
-    this._addStaticLabel();
-    this._updateLabelsObject();
+    setTimeout(() => {
```
Waseem826, line 55: Why do we need `setTimeout`? Can we handle this in `ngOnChanges` if we want to wait for the value of `staticLabels`?

ahmadhamzh: well i wasn't sure what to use here `ngOnChanges` or just `setTimeout`, but i thought `setTimeout` is simpler and i only need the static labels just for one time.
you think `ngOnChanges` is better ?

ahmedwaleedmalik (approved):
/approve
/hold so that @Waseem826's feedback can be resolved.

Apart from the feedback by Waseem. I've tested these changes locally and they look good, as always. 👍

modules/web/src/app/settings/admin/defaults/static-labels-form/component.ts
Waseem826: `ngOnChanges` is better because we can be sure that `staticLabels` is not empty. With `setTimeout` we're just guessing that `staticLabels` will have a value in 500ms.

Waseem826 (approved):

ahmedwaleedmalik (approved):

PR #6907, opened 2024-10-02
Prevent system applications from re-appearing after an application is updated

Waseem826:
Prevent system applications from re-appearing after an application is updated.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


modules/web/src/app/settings/admin/applications/component.ts
```diff
@@ -150,7 +150,6 @@ export class ApplicationsComponent implements OnInit, OnChanges {
           const index = this.applications.findIndex(app => app.name === updatedApplication.name);
           if (index !== -1) {
             this.applications[index] = updatedApplication;
-            this.dataSource.data = [...this.applications];
```
Waseem826, line 153: Removing this altogether because updating it causes the whole table to re-render and flickers the toggles.

ahmedwaleedmalik (approved):
/approve

PR #6915, opened 2024-10-07
Support KubeVirt Provider Network

moadqassem:
With some specific CNI, KubeVirt can use the networking model to support different network fabric features. One of these features are VPC and Subnet where we can run VMs in a dedicated isolated VPCs in addition to dedicate a subnet for VMs, where the IPs are handed out of a dedicated subnet. 

This PR adds the support in the API where users can list the VPCs and choose one. In addition to that, assign a subnet of that VPC to the cluster, so all created VMs will have an IP out of that subnet. 

Fixes #

**What type of PR is this?**

/kind feature

```release-note
Support Kube-OVN provider networks for VPCs and Subnets
```

**Documentation**:

```documentation
NONE
```


ahmedwaleedmalik (approved):
/approve
/hold for confirmation regarding endpoints

PR #6949, opened 2024-10-23
Read KubeVirt infra namespace from datacenter

moadqassem:
In KKP KubeVirt provider, there are two modes used to provision VMs in the infra cluster. Namespaced one where all VMs from all clusters are created in one namespace. and multi-tenant one where clusters VMs will be created in a separate namespace called after the cluster name. This PR adjust the api code to fetch the right resources in respect to these two modes, instead of always reading the resources from the cluster namespace.   

Fixes #

**What type of PR is this?**
/kind bug

```release-note
None
```

**Documentation**:

```documentation
None
```


ahmedwaleedmalik (approved):
/approve

PR #6962, opened 2024-11-04
Display only the admin-allowed OS on project add/edit dialog

### LINKED ISSUE #6952: Project owners can allow/disallow operating systems that are actually disallowed globally, resulting in confusion

Project owners can choose which OSes to allow, but OSes that are not allowed globally (admin settings) are shown. Allowing an OS on the project level doesn't overwrite the global settings, but it's confusing to project owners why allowed OSes aren't available.

Only globally-allowed OSes should be shown in project settings

**Screenshots**
Project settings:
Image

Admins settings:
Image

ahmadhamzh:
Only operating systems allowed by the admin will be displayed in the add/edit dialog.

**Admin panel:**
image

**Edit project dialog:**
image

**Add project dialog:**
image

Fixes #6952 
/kind design

```release-note
Display only the admin-allowed OS's on project add/edit dialog.
```

**Documentation**:
```documentation
NONE
```


Waseem826 (approved):
/approve

PR #6972, opened 2024-11-06
Bump KKP API version

archups:
This PR is to bump the KKP API version in the swagger.json. 

Fixes #

**What type of PR is this?**
/kind documentation

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/cmd/kubermatic-api/swagger.json
+++ b/modules/api/cmd/kubermatic-api/swagger.json
@@ -12,7 +12,7 @@
   "info": {
     "description": "This OpenAPI 2.0 specification describes the REST APIs used by the Kubermatic Kubernetes Platform Dashboard.",
     "title": "Kubermatic Kubernetes Platform API",
-    "version": "2.23"
+    "version": "2.26"
   },
   "paths": {
     "/api/projects/{project_id}/clusters/{cluster_id}/sshkeys/{key_id}": {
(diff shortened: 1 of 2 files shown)


Waseem826 (approved):
/approve

embik:
Shouldn't this be 2.27 for the `main` branch? We already released 2.26, so the next release cut from `main` will be 2.27.

modules/api/cmd/kubermatic-api/swagger.json
```diff
@@ -12,7 +12,7 @@
   "info": {
     "description": "This OpenAPI 2.0 specification describes the REST APIs used by the Kubermatic Kubernetes Platform Dashboard.",
     "title": "Kubermatic Kubernetes Platform API",
-    "version": "2.23"
+    "version": "2.26"
```
embik, line 15: `swagger.json` is generated, we can't change it manually. The source is this line: https://github.com/kubermatic/dashboard/blob/main/modules/api/cmd/kubermatic-api/main.go#L22.

embik (changes requested):

modules/api/cmd/kubermatic-api/swagger.json
archups: Bit confused since I reference this [commit ](https://github.com/kubermatic/dashboard/commit/b6d97439f61eae5218f0bf42e8a09af51327922c) now and there is manual change to swagger.json. So should I keep it or revert if there is any change in process later.

embik: You can keep it I suppose, the `swagger.json` _should_ be changed, but it should be because you ran codegen scripts, not because you updated it yourself.

Waseem826 (approved):

PR #6980, opened 2024-11-08
Allow adding a postfix to the page title

### LINKED ISSUE #6917: Page titles should include KKP (or similar)

Currently the page titles (i.e. the `<title>` tag) do not include the platform's name. If I have a tab open that just reads `Projects`, that's not very helpful. I suggest you either add "KKP" oder "Kubermatic Kubernetes Platform" as a postfix. Ideally it would be configurable (at least for the EE edition) so we can set it to our product's name instead.

Add `| Kubermatic Kubernetes Platform` to all `<title>` tags. Make the string configurable through the KubermaticConfiguration (or KubermaticSettings). If preferred, the default could also be an empty string, to keep the current UI/UX, and just allowing us to define a postfix for the title tag.

ahmadhamzh:
Added a new `postfix_page_title` property in the UI config to allow setting a custom postfix for the page title.

Fixes #6917 

**What type of PR is this?**
/kind feature

```release-note
Added a new `postfix_page_title` property in the UI config to allow setting a custom postfix for the page title.
```

**Documentation**:
```documentation
TBD
```

### DIFF
--- a/modules/web/src/app/shared/entity/common.ts
+++ b/modules/web/src/app/shared/entity/common.ts
@@ -63,6 +63,8 @@ export enum View {
   ClusterRestore = 'clusterrestores',
   BackupStorageLocation = 'backupstoragelocations',
   Access = 'access',
+  RestAPI = 'rest-api',
+  SignIn = 'signIn',
 }
 
 // IMPORTANT: names in 'ViewDisplayName' have to be in sync with names in 'View'
@@ -89,6 +91,8 @@ export enum ViewDisplayName {
   ClusterRestore = 'Cluster Restores',
   BackupStorageLocation = 'Backup Storage Locations',
   Access = 'Access',
+  RestAPI = 'Rest API',
+  SignIn = 'SignIn',
 }
 
 export enum ProjectSidenavSection {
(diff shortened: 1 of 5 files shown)


Waseem826:
I think it should be called postfix. Also, the format was something like `| <postfix> or Kubermatic Kubernetes Platform`.

> Add | Kubermatic Kubernetes Platform to all <title> tags. Make the string configurable through the KubermaticConfiguration (or KubermaticSettings). If preferred, the default could also be an empty string, to keep the current UI/UX, and just allowing us to define a postfix for the title tag.

ahmadhamzh:
> If I have a tab open that just reads Projects, that's not very helpful. I suggest you either add "KKP" oder "Kubermatic Kubernetes Platform" as a postfix.

yes he mention postfix but since the issue with open tab i thought maybe it's better to be a prefix :thinking:

Waseem826:
> > If I have a tab open that just reads Projects, that's not very helpful. I suggest you either add "KKP" oder "Kubermatic Kubernetes Platform" as a postfix.
> 
> Yes, they mentioned a postfix, but since the main concern was open tabs, I thought using a prefix might be better. 🤔

No, primary information about the page (title in this case) should come first. Secondary information like organization/platform name should come last.

modules/web/src/app/shared/entity/common.ts
```diff
@@ -89,6 +91,8 @@ export enum ViewDisplayName {
   ClusterRestore = 'Cluster Restores',
   BackupStorageLocation = 'Backup Storage Locations',
   Access = 'Access',
+  RestAPI = 'Rest api',
```
Waseem826, line 94:
```suggestion
  RestAPI = 'Rest API',
```

Waseem826 (approved):
/approve

PR #7026, opened 2024-11-30
Remove CentOS

julioc-p:

Centos has reached EOL, and we are removing it from KKP. 

ref https://github.com/kubermatic/kubermatic/issues/13580

**What type of PR is this?**

```release-note
Remove CentOS as a supported operating system since it has reached EOL.
```

**Documentation**:

```documentation
NONE
```


modules/web/src/assets/css/_images.scss
```diff
@@ -570,10 +570,6 @@
   @include mixins.provider-image('/assets/images/clouds/kubeone.svg', 130px, 30px);
 }
 
-.km-os-image-centos {
```
Waseem826, line 573: IMO we should keep the CentOS icon for now otherwise if someone has older cluster with CentOS then it will cause a visual problem. Or is there no way for a CentOS cluster to exist after 2.27?

julioc-p: CentOS was not working before as well. So there should be no possibility of having a cluster running CentOS.

ahmedwaleedmalik:
@julioc-p The changes in the PR look great. 

Please improve the release note from `Remove Centos` to something more succinct. Also, reference the original ticket for these PRs.

ahmedwaleedmalik (approved):
/approve

PR #7034, opened 2024-12-03
Remove the `includedNamespaces` property when select all namespaces

### LINKED ISSUE #7032: UC Backups: "backup all namespaces" does not actually backup all namespaces
### What happened?

When creating a backup or backup schedule, it's possible to choose "all namespaces" (the default). However, KKP will then configure the backup to instead backup all namespaces that existed when the backup was created in the UI. In other words: the velero backup configuration will contain an explicit list of namespaces.

There are two problems with this:
1. Namespaces that are created later, will not be included in the backup.
2. Cluster scoped resources are not backed up. By default, Velero does include cluster scoped resources only if truly all namespaces (i.e. without filter) are backed up, and not include them if a list of namespaces is provided. (By default, because  it's possible to overwrite this behavior to always/never include cluster scoped resources in Velero).

### Expected behavior

If "backup all namespaces" is selected, all namespaces, including those created in the future (for a backup schedule) are backed up. And by implication, all cluster scoped resources are backed up as well.

- Enable the UC Backup feature in the admin settings
- configure a backup location
- enable the UC backup for a UC
- create a backup or backup schedule where you enable the "backup all namespaces" checkbox
- download UC kubeconfig
- `velero backup get <name> -o json | jq .spec.includedNamespaces`

This should not be set all all (or possibly an empty list / nil would be acceptable).

After the backup was executed, you'll also find such log lines:

```
$ k -n velero logs -l app=velero --tail=1000 | grep cluster-scoped | head
Defaulted container "velero" out of: velero, velero-velero-plugin-for-aws (init)
time="2024-11-19T20:00:50Z" level=info msg="Skipping resource certificatesigningrequests.certificates.k8s.io, because it's cluster-scoped and only specific namespaces or namespace scope types are included in the backup." backup=velero/debug-test-schedule-20241119200049 logSource="pkg/util/collections/includes_excludes.go:155"
time="2024-11-19T20:00:50Z" level=info msg="Skipping resource ingressclasses.networking.k8s.io, because it's cluster-scoped and only specific namespaces or namespace scope types are included in the backup." backup=velero/debug-test-schedule-20241119200049 logSource="pkg/util/collections/includes_excludes.go:155"
```

- KKP version: 2.26.0
- Shared or separate master/seed clusters?: shared

N/A

N/A

N/A

ahmadhamzh:
Remove the `includedNamespaces` property when select all namespaces in the create backup/schedule 

Fixes #7032 

**What type of PR is this?**
/kind bug

```release-note
Fix backing up all namespaces.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/add-dialog/component.ts
+++ b/modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/add-dialog/component.ts
@@ -196,7 +196,7 @@ export class AddClustersBackupsDialogComponent implements OnInit, OnDestroy {
     };
 
     if (this.form.get(Controls.AllNamespaces).value) {
-      backup.spec.includedNamespaces = this.nameSpaces;
+      delete backup.spec?.includedNamespaces;
     } else {
       backup.spec.includedNamespaces = this.form.get(Controls.Namespaces).value;
     }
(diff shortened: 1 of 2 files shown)


ahmedwaleedmalik (approved):
/approve

PR #7044, opened 2024-12-13
Make `Domain` field optional when using application credentials for Openstack provider

### LINKED ISSUE #7033: OpenStack: "domain" is wrongfully required when specifying "application credentials" for authorization
### What happened?

When creating a cluster and specifying "application credentials", a `Domain` is required 
in the UI (despite the CRD not requiring it). However, I think a domain should only be required when user credentials are used.

Additionally, when creating a preset, the UI doesn't enforce specifying a domain, but this time the CRD does. If not specified in the UI, the domain is simply set it `""` and it's unclear whether that's identical with the domain not being defined at all.

### Expected behavior

Only require a domain when user credentials are used, but not when application credentials are used.

In the cluster creation wizard, when creating a cluster in OpenStack, try providing application credentials but no domain.

- KKP version: 2.25.7
- Shared or separate master/seed clusters?: shared

N/A

OpenStack

Ubuntu 22.04

### Additional information

See `domain-id` in the OS CCM docs: https://github.com/kubernetes/cloud-provider-openstack/blob/master/docs/openstack-cloud-controller-manager/using-openstack-cloud-controller-manager.md#global

Waseem826:
Make `Domain` field optional when using application credentials for Openstack provider in cluster creation wizard:

https://github.com/user-attachments/assets/addd19d1-2bc4-4648-933c-996563a1a62f

Fixes #7033 

**What type of PR is this?**

/kind bug

```release-note
Make `Domain` field optional when using application credentials for Openstack provider.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/presets/dialog/steps/settings/provider/openstack/component.ts
+++ b/modules/web/src/app/settings/admin/presets/dialog/steps/settings/provider/openstack/component.ts
@@ -115,7 +115,7 @@ export class OpenstackSettingsComponent extends BaseFormValidator implements OnI
   private _update(): void {
     this._presetDialogService.preset.spec.openstack = {
       ...this._presetDialogService.preset.spec.openstack,
-      domain: this.form.get(Controls.Domain).value,
+      domain: this.form.get(Controls.Domain).value || null,
       network: this.form.get(Controls.Network).value,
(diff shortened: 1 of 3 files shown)


ahmadhamzh (approved):
/approve

PR #7069, opened 2025-01-09
Add search field in autocomplete options dropdown

### LINKED ISSUE #7038: Select OSP dropdown is confusing
The km-autocomplete component's dropdown has a filter functionality that filters options based on the value the user types. When a default value is pre-set, users get confused because, upon opening the dropdown, they only see the filtered value instead of all available options.

Image

Avoid using the km-autocomplete component when a default value is set to ensure all options are visible when the dropdown is opened.

Waseem826:
Add a new search field in options dropdown of km-autocomplete instead of using the value for filtering.

https://github.com/user-attachments/assets/1392fc94-5d60-4ad2-b68a-a62c13e79010

Fixes #7038

**What type of PR is this?**

/kind design

```release-note
Add search field in options dropdown of autocomplete element.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/shared/components/autocomplete/component.ts
+++ b/modules/web/src/app/shared/components/autocomplete/component.ts
@@ -50,6 +50,7 @@ export class AutocompleteComponent extends BaseFormValidator implements OnInit {
   @Input() disabled: boolean;
   controls = AutocompleteControls;
   isDropdownOpen: boolean;
+  search: string = '';
 
   constructor(private readonly _builder: FormBuilder) {
     super();
@@ -81,6 +82,7 @@ export class AutocompleteComponent extends BaseFormValidator implements OnInit {
 
   onDropdownOpened(): void {
     this.isDropdownOpen = true;
+    this.search = '';
   }
 
   onDropdownClosed(): void {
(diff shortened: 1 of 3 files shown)


ahmedwaleedmalik (changes requested):
This doesn't actually solve the UX issue. The requirement would be to show all the OSPs unfiltered when an OSP is already selected and to perform the filtering/search functionality when some user types something in the search bar/field.

Waseem826:
> This doesn't actually solve the UX issue. The requirement would be to show all the OSPs unfiltered when an OSP is already selected and to perform the filtering/search functionality when some user types something in the search bar/field.

Yes, that's exactly how it works now. Currently, there are only 2 options in the dropdown with similar names so that might've caused the confusion. As an example, this is how it worked previously:

https://github.com/user-attachments/assets/77cc6bcb-6c82-46a9-9e2d-0f2ad572ebad

So now regardless of what's selected, filtering will work on the basis of this new search field inside dropdown.

Waseem826:
I've updated the video in description so hopefully it better showcases the fix.

ahmedwaleedmalik:
> I've updated the video in description so hopefully it better showcases the fix.

Got slightly confused, but after looking at the code and your updated video all is good. Thanks, the change looks good 👍

ahmedwaleedmalik (approved):
/approve

PR #7144, opened 2025-02-06
Filter KubeVirt Storage Casses based on volume Provisioner

moadqassem:
For Kubevirt there are some storage classes that can be used for VM disk images and PVs in the tenant cluster, in addition there are other storage classes that can be used only for PVs in the tenant cluster, This PR make sure that the api would only retrieve storage classes that can be used by CDI to create VM image disks instead of fetching all the storage classes and show them in the UI. 

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Filter out KubeVirt storage classes in the API/UI based on the the volume provisioner field. 
```

**Documentation**:

```documentation
None
```


Waseem826 (approved):
/approve

Waseem826 (approved):

PR #7146, opened 2025-02-07
Handle default namespace for applications

### LINKED ISSUE #6858: Support for specifying Application Installation namespace
As a KKP user, I'd like the ability to explicitly specify the namespace in which my `ApplicationInstallation` resource will be created. Right now, we are only allowed to specify a single namespace that then is used for both the ApplicationInstallation and `application.spec.namespace.name` which determines where the resources for application will be created/installed.

This makes it a bit ambiguous as it should be possible to create all the ApplicationInstallations in a single namespace irrespective of where the resources are created since they have a separate dedicated field.

### LINKED ISSUE #7084: Default Namespace for the Application

Frontend changes for:
- https://github.com/kubermatic/kubermatic/issues/12578
- https://github.com/kubermatic/kubermatic/pull/13932

Waseem826:
This PR adds functionality to handle default namespaces for application installation and application resources.
A new field has also been added to customize both namespace values.

**Add Application Dialog:**
Screenshot 2025-02-07 at 7 54 41 PM

Fixes #6858
Fixes #7084 

**What type of PR is this?**

/kind feature

```release-note
Add functionality to handle default namespaces for application installation and application resources.
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/1890
```

### DIFF
--- a/modules/web/src/app/shared/components/application-list/add-application-dialog/template.html
+++ b/modules/web/src/app/shared/components/application-list/add-application-dialog/template.html
@@ -116,17 +116,17 @@ limitations under the License.
               </div>
             </div>
             <mat-form-field>
-              <mat-label>Namespace</mat-label>
+              <mat-label>Application Installation Namespace</mat-label>
               <input matInput
                      required
-                     [formControlName]="Controls.Namespace"
+                     [formControlName]="Controls.AppInstallationNamespace"
                      type="text"
                      autocomplete="off">
-              <mat-hint>Namespace where application resources will be deployed.</mat-hint>
-              <mat-error *ngIf="form.get(Controls.Namespace).hasError('required')">
+              <mat-hint>Namespace where application installation will be created.</mat-hint>
+              <mat-error *ngIf="form.get(Controls.AppInstallationNamespace).hasError('required')">
                 <strong>Required</strong>
               </mat-error>
-              <mat-error *ngIf="form.get(Controls.Namespace).hasError('pattern')">
+              <mat-error *ngIf="form.get(Controls.AppInstallationNamespace).hasError('pattern')">
                 Namespace can only contain alphanumeric characters and dashes (only between characters).
               </mat-error>
             </mat-form-field>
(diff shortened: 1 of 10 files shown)


modules/web/src/app/shared/components/application-list/add-application-dialog/template.html
```diff
@@ -116,17 +116,17 @@
               </div>
             </div>
             <mat-form-field>
-              <mat-label>Namespace</mat-label>
+              <mat-label>Application Installation Namespace</mat-label>
```
ahmadhamzh, line 119: isn't better to have the name of the application first ?

modules/web/src/app/shared/components/application-list/edit-application-dialog/component.ts
```diff
@@ -171,8 +172,18 @@ export class EditApplicationDialogComponent implements OnInit, OnDestroy {
...
         )
       );
+      this.form.addControl(
+        Controls.AppResourcesNamespace,
+        this._builder.control(
+          {
+            value: this.application.spec.namespace?.name,
+            disabled: this.isEnforcedApplication(),
+          },
+          [KUBERNETES_RESOURCE_NAME_PATTERN_VALIDATOR]
```
ahmadhamzh, line 182: Two questions:

1- Can we edit the namespaces and the app name?
2- If so, shouldn't we add Validators.required?

ahmadhamzh (changes requested):
small notes
and if you can check edit application i got some error's regarding the `appResourcesNamespace` control

image

modules/web/src/app/shared/components/application-list/add-application-dialog/template.html
Waseem826: Since Application Installation Namespace has higher scope than other fields so I thought it might be better to put it first. Also, I didn't want the confusion between both namespace fields by putting them together.

modules/web/src/app/shared/components/application-list/edit-application-dialog/component.ts
Waseem826: No, we can only edit application `values` field.

Waseem826:
> small notes and if you can check edit application i got some error's regarding the `appResourcesNamespace` control
> 
> image

Fixed.

ahmedwaleedmalik (approved):
/approve

PR #7155, opened 2025-02-11
Fix swagger client for Kyverno

ahmedwaleedmalik:
Kyverno causes conflicts with Kubernetes Admission APIs, thus resulting in inconsistent swagger spec generation on each run. Ignoring Kyverno APIs is not possible as it would break the generation of the Kyverno API client. So instead, we are excluding the conflicting APIs/models "k8s.io/api/admissionregistration".

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


ahmadhamzh (approved):
/approve

PR #7164, opened 2025-02-14
Prepare 2.27.0 rc.0

xrstf:
https://github.com/kubermatic/kubermatic/pull/14142 changed the API, so we have to bump the KKP dependency here. And since our next release is only 2 days away, an RC is more realistic than a beta anyway.

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


Waseem826 (approved):
/approve

PR #7168, opened 2025-02-17
Add select all namespaces to restore and few enhancements

### LINKED ISSUE #7125: Cluster backups - Test Release 2.27
Test Cluster Backups/Schedule/Restore.

```[tasklist]
### Tasks
- [x] Only clusters that have the cluster backup feature are shown in the clusters list.
- [x] Items that are shown in the table belong to the same selected cluster.
- [x] Create Backup/Schedule/Restore and check if the status is running.
- [x] Test delete one, delete selected.
- [x] Download the backup files from the download button in the actions column.
```

ahmadhamzh:
add select all namespaces for restore dialog, small enhancements, and remove the experimental tag 

image

image

Fixes #7125

**What type of PR is this?**
/kind bug

```release-note
In the cluster backup feature, fix the issue with restoring a backup that includes all namespaces and add the option to restore all namespaces from a backup.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/template.html
+++ b/modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/template.html
@@ -50,7 +50,7 @@ END OF TERMS AND CONDITIONS
       <div class="action-button"
            *ngIf="clusters?.length">
         <button *ngIf="selectedBackups.length"
-                mat-icon-button
+                mat-flat-button
                 type="button"
                 matTooltip="Delete selected"
                 (click)="deleteBackups(selectedBackups)">
@@ -248,8 +248,8 @@ END OF TERMS AND CONDITIONS
           *matHeaderRowDef="columns"></tr>
       <tr mat-row
           *matRowDef="let element; columns: columns"
-          class="km-mat-row km-pointer"
-          [ngClass]="{'km-selected': checkSelected(element.id)}"
+          class="km-mat-row"
+          [ngClass]="{'km-selected': checkSelected(element.id), 'km-pointer': element.spec.includedNamespaces}"
           (click)="toggleBackupDetail(element.name)"></tr>
       <tr mat-header-row
           *matRowDef="let element; columns: toggleableColumn"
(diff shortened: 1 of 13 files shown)


modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/template.html
```diff
@@ -248,8 +248,8 @@
           *matHeaderRowDef="columns"></tr>
       <tr mat-row
           *matRowDef="let element; columns: columns"
-          class="km-mat-row km-pointer"
-          [ngClass]="{'km-selected': checkSelected(element.id)}"
+          class="km-mat-row"
+          [ngClass]="{'km-selected': checkSelected(element.id), 'km-pointer': element.spec.includedNamespaces}"
```
Waseem826, line 252: Why do we need this class on whole row? Can you please share a screenshot for this change? Same question for changes in restores and schedules table.

modules/web/src/app/dynamic/enterprise/cluster-backups/list/restore/add-dialog/component.ts
```diff
@@ -47,23 +53,45 @@ export class AddRestoreDialogComponent implements OnInit {
   private readonly _unsubscribe = new Subject<void>();
   projectID = this._config.projectID;
   backup = this._config.backup;
+  nameSpaces: string | string[] = [];
```
Waseem826, line 56: If the default values is `[]` then type should be `string[]`.

Waseem826 (changes requested):
Some feedback. PTAL

modules/web/src/app/dynamic/enterprise/cluster-backups/list/restore/add-dialog/component.ts
Waseem826: Also, I don't see where we assign a string value to this field.

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backups/template.html
ahmadhamzh: the change only cause we show namespaces if not all namespaces is included, so if no namespaces is shown there is no need for km-pointer

Waseem826 (approved):
/approve

Waseem826 (approved):

PR #7203, opened 2025-03-06
Fix `e2e:mock` script

ahmedwaleedmalik:
Since we don't explicitly set the value for `CYPRESS_MOCK` in the npm script, it was being set as `false` in case of mock tests through https://github.com/kubermatic/dashboard/blob/main/modules/web/package.json#L27

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #7206, opened 2025-03-07
Split large token to multi cookies

### LINKED ISSUE #7182: KKP/Dex login not working with too many AD-groups/roles
### What happened
Kubermatic KKP (or rather the included Dex probably) currently seems to have an issue when attaching an OIDC or OAuth2 IDP (via Dex connectors) if a user having too many roles/groups tries to login.
The log output of Dex reports success, yet the login into the KKP UI does not actually work and goes back to the login page. Customer has confirmed this to be a bug with the amount of groups/roles being present in users during login. If the number of groups is low, then the login will succeed, if there's too many it will fail.

### Expected behavior
Login works properly even for a large number of groups.

### How to reproduce
Try to log in to KKP using a user with large number of groups assigned.

### Environment
- KKP Version: v2.26

### Current workaround
Reduce number of groups for the users or filter the groups using dex configuration (e.g. when using LDAP connector)

reference: #6989

ahmadhamzh:
Fix the issue when the user tries to log in and is a member of many groups/roles, which results in a huge token
Fixes #7182 

**What type of PR is this?**
/kind bug

```release-note
Fix KKP login issue when the ID token is too large to be saved in a cookie, by splitting the token into multiple cookies.
```

**Documentation**:
```documentation
TBD
```

### DIFF
--- a/modules/api/cmd/kubermatic-api/main.go
+++ b/modules/api/cmd/kubermatic-api/main.go
@@ -489,6 +489,7 @@ func createAuthClients(options serverRunOptions, prov providers) (authtypes.Toke
 			auth.NewHeaderBearerTokenExtractor("Authorization"),
 			auth.NewCookieHeaderBearerTokenExtractor("token"),
 			auth.NewQueryParamBearerTokenExtractor("token"),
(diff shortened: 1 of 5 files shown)


Waseem826 (approved):
/approve

PR #7217, opened 2025-03-12
(fix): use OSP from node instead of default OSP from dc

buraksekili:

This PR updates `getOperatingSystemProfile` helper to parse Node annotations instead of relying on the dc spec, which needs to be used as a default, not as an enforced OSP

Fixes https://github.com/kubermatic/kubermatic/issues/14028

**What type of PR is this?**

/kind bug

```release-note
Updated Dashboard API to use correct OSP which is selected while creating a cluster.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/resources/machine/machinedeployment.go
+++ b/modules/api/pkg/resources/machine/machinedeployment.go
@@ -75,13 +75,13 @@ func Deployment(ctx context.Context, c *kubermaticv1.Cluster, nd *apiv1.NodeDepl
 	// 3. Allow empty value to let OSM apply its defaulting logic
 	if osp := nd.Annotations[osmresources.MachineDeploymentOSPAnnotation]; osp == "" {
 		osp = getOperatingSystemProfile(nd, dc)
+		if osp != "" {
+			if md.Annotations == nil {
+				md.Annotations = make(map[string]string)
+			}
 
-		if md.Annotations == nil {
-			md.Annotations = make(map[string]string)
+			md.Annotations[osmresources.MachineDeploymentOSPAnnotation] = osp
 		}
-
-		// Set the osp value (even if empty) to enable OSM handling.
-		md.Annotations[osmresources.MachineDeploymentOSPAnnotation] = osp
 	}
 
 	md.Namespace = metav1.NamespaceSystem
--- a/modules/api/pkg/resources/machine/machinedeployment.go
+++ b/modules/api/pkg/resources/machine/machinedeployment.go
@@ -69,11 +69,18 @@ func Deployment(ctx context.Context, c *kubermaticv1.Cluster, nd *apiv1.NodeDepl
 	// Add Annotations to Machine Deployment
 	md.Annotations = nd.Annotations
 
-	osp := getOperatingSystemProfile(nd)
-	if osp != "" {
+	// OSP is an optional value passed via annotations with fallback logic:
+	// 1. Use existing non-empty annotation if present
+	// 2. Fall back to datacenter-level defaults when annotation is missing/empty
+	// 3. Allow empty value to let OSM apply its defaulting logic
+	if osp := nd.Annotations[osmresources.MachineDeploymentOSPAnnotation]; osp == "" {
+		osp = getOperatingSystemProfile(nd, dc)
+
 		if md.Annotations == nil {
 			md.Annotations = make(map[string]string)
 		}
+
+		// Set the osp value (even if empty) to enable OSM handling.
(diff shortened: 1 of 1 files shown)


SimonTheLeg:
from a Go perspective this looks good to me; however I have never written any OSP code and am definitely not the best reviewer for this (not sure why the bot recommends me here :D)

/cc @ahmedwaleedmalik  
I think you are the expert on API and OSP, I think it would be best if you could have a look at this one

ahmedwaleedmalik:
> recommends

I'll take a proper look tomorrow, from a quick glance the change looks wrong unfortunately.

modules/api/pkg/resources/machine/machinedeployment.go
```diff
@@ -316,26 +316,13 @@ func getProviderConfig(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc *ku
...
-		return dc.Spec.DefaultOperatingSystemProfiles[providerconfig.OperatingSystemFlatcar]
-	case nd.Spec.Template.OperatingSystem.RockyLinux != nil:
-		return dc.Spec.DefaultOperatingSystemProfiles[providerconfig.OperatingSystemRockyLinux]
-	case nd.Spec.Template.OperatingSystem.AmazonLinux != nil:
-		return dc.Spec.DefaultOperatingSystemProfiles[providerconfig.OperatingSystemAmazonLinux2]
-	default:
-		return ""
-	}
+	return ""
 }
```
ahmedwaleedmalik, line 326: This code is required and shouldn't be removed.

modules/api/pkg/resources/machine/machinedeployment.go
```diff
@@ -316,26 +316,13 @@ func getProviderConfig(c *kubermaticv1.Cluster, nd *apiv1.NodeDeployment, dc *ku
...
 
-func getOperatingSystemProfile(nd *apiv1.NodeDeployment, dc *kubermaticv1.Datacenter) string {
-	if dc.Spec.DefaultOperatingSystemProfiles == nil {
-		return ""
+// getOperatingSystemProfile returns the OSP selected while creating the cluster.
+func getOperatingSystemProfile(nd *apiv1.NodeDeployment) string {
+	if osp := nd.Annotations[osmresources.MachineDeploymentOSPAnnotation]; osp != "" {
+		return osp
 	}
```
ahmedwaleedmalik, line 323: OSP is an optional value(passed in annotation) here and it's not required for frontend or API to always pass it. In case the value is not provided, there are two options:
1. When defaults are defined at datacenter level, use them.
2. Otherwise set "k8c.io/operating-system-profile" as empty string and let OSM default/mutate the value.

We can't remove the datacenter default handling from here.

modules/api/pkg/resources/machine/machinedeployment.go
```diff
@@ -69,7 +69,7 @@ func Deployment(ctx context.Context, c *kubermaticv1.Cluster, nd *apiv1.NodeDepl
 	// Add Annotations to Machine Deployment
 	md.Annotations = nd.Annotations
 
-	osp := getOperatingSystemProfile(nd, dc)
+	osp := getOperatingSystemProfile(nd)
```
ahmedwaleedmalik, line 72:
```suggestion
	// Default OSP value if the OSP annotation on the MD is not preset.
	if osp := nd.Annotations[osmresources.MachineDeploymentOSPAnnotation]; osp == "" {
		osp := getOperatingSystemProfile(nd, dc)
```

ahmedwaleedmalik: Something like this should be sufficient to fix this bug, needs testing though.

ahmedwaleedmalik (changes requested):

modules/api/pkg/resources/machine/machinedeployment.go
ahmedwaleedmalik: > Default OSP value if the OSP annotation on the MD is not preset.

Default OSP value if the OSP annotation on the MD is not preset or present but empty. <-- something along these lines

modules/api/pkg/resources/machine/machinedeployment.go
buraksekili: Thanks @ahmedwaleedmalik for the detailed explanation! Really appreciated it.

I've updated the code accordingly, please let me know if it needs more work. Thanks!

ahmedwaleedmalik (approved):
/approve

ahmedwaleedmalik:
@buraksekili can you please backport this to 2.27, 2.26, and 2.25 as well?

PR #7252, opened 2025-04-02
Add kubevirt cpu alloc ratio support

soer3n:

This pr adds the logic for the [new introduced field in kkp seed crd](https://github.com/kubermatic/kubermatic/issues/14245) to control whether cpus for kubevirt machines are configured by `domain.cpu` or `resources.requests.cpu` and `resources.limits.cpu`.

It also modifies the ui to parse in both cases the cpu value properly.

Fixes #

**What type of PR is this?**

/kind feature

```release-note
The kkp api is now aware on how to configure cpus for kubevirt virtual machines based on a new introduced field in kkp seed crd called `spec.datacenters.<example-dc>.spec.kubevirt.enableDedicatedCpus`.
```

**Documentation**:

```documentation
NONE
```


modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
```diff
@@ -662,13 +662,18 @@ export class KubeVirtBasicNodeDataComponent
...
       this.form.get(Controls.NodeAffinityPreset).setValue(this._initialData.nodeAffinityPreset?.Type);
       this.form.get(Controls.NodeAffinityPresetKey).setValue(this._initialData.nodeAffinityPreset?.Key);
       this.nodeAffinityPresetValues = this._initialData.nodeAffinityPreset?.Values || [];
 
+      if (!_.isEmpty(this._initialData.vcpus)) {
+        this.form.get(Controls.CPUs).setValue(this._initialData.vcpus.cores);
+      } else {
+        this.form.get(Controls.CPUs).setValue(parseInt(this._initialData.cpus) || this._defaultCPUs);
+      }
+
```
ahmedwaleedmalik, line 676: This looks rather wrong and hackish. UI should deal with vCPUs and CPUs dedicatedly as individual fields. This should have a 1-1 relation with the machine deployment.

modules/api/pkg/resources/machine/common.go
```diff
@@ -619,6 +618,16 @@ func GetKubevirtProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.Nod
...
+	if dc.Spec.Kubevirt.EnableDedicatedCPUs {
+		vcpus, err := strconv.ParseInt(nodeSpec.Cloud.Kubevirt.CPUs, 0, 64)
+		if err != nil {
+			return nil, err
+		}
+		config.VirtualMachine.Template.VCPUs.Cores = int(vcpus)
+	} else {
+		config.VirtualMachine.Template.CPUs = providerconfig.ConfigVarString{Value: nodeSpec.Cloud.Kubevirt.CPUs}
+	}
+
```
ahmedwaleedmalik, line 630: If a flag on dc should enforce such behaviour then it needs to be used on the UI as well with enough description(tooltips etc) that `EnableDedicatedCPUs` is enabled and `VCPUs` will be configured instead of `CPUs`.

modules/api/pkg/resources/machine/common.go
```diff
@@ -619,6 +618,16 @@ func GetKubevirtProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.Nod
 		},
 	}
 
+	if dc.Spec.Kubevirt.EnableDedicatedCPUs {
```
ahmedwaleedmalik, line 621: Based on https://github.com/kubermatic/dashboard/pull/7252/files#diff-1c943cc76ece28ac24a9cbdb428ae07c785ad35b82d654e0febb0b93db03f2e3R364-R368, this can be highly problematic if I enable `EnableDedicatedCPUs` for a seed which already has clusters/machinedeployments configured with "CPU". For any update(post/put) request for MD, API would start converting the CPU fields to VCPUs and thus configure both CPU and VCPUs fields which should return a validation failure from MC webhooks. 

We'll have the same case if `EnableDedicatedCPUs` is disabled with MDs present.

modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
```diff
@@ -662,13 +662,18 @@ export class KubeVirtBasicNodeDataComponent
...
     if (this._initialData) {
       this.form.get(Controls.Memory).setValue(parseInt(this._initialData.memory) || this._defaultMemory);
-      this.form.get(Controls.CPUs).setValue(parseInt(this._initialData.cpus) || this._defaultCPUs);
       this.form.get(Controls.PrimaryDiskSize).setValue(this._initialData.primaryDiskSize);
       this.form.get(Controls.PrimaryDiskOSImage).setValue(this._initialData.primaryDiskOSImage);
       this.form.get(Controls.NodeAffinityPreset).setValue(this._initialData.nodeAffinityPreset?.Type);
       this.form.get(Controls.NodeAffinityPresetKey).setValue(this._initialData.nodeAffinityPreset?.Key);
       this.nodeAffinityPresetValues = this._initialData.nodeAffinityPreset?.Values || [];
 
+      if (!_.isEmpty(this._initialData.vcpus)) {
```
ahmedwaleedmalik, line 671: No need to use `isEmpty` here, this can simply be a nil/empty check `this._initialData.vcpus?.cores`

ahmedwaleedmalik (changes requested):

modules/api/pkg/resources/machine/common.go
moadqassem: I don't believe this problematic especially that we will have a documentation about this field stating that, if this field is enabled MachineDeployemtns must be rotated so it's fine at this point

modules/api/pkg/resources/machine/common.go
moadqassem: Why is that ? We have a lots of fields in the seed which are not using any tooltip and they have the exact same effect, actually we even have a whole mode that changes the whole behavior of the cloud provider(NamespacedMode) which is also not having tool tip or something similar. So I think this is also ok in this case

modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
moadqassem: @soer3n for the UI I will speak with @ahmadhamzh if he can assist us even though I am sure that we don't need any UI changes. Let's discuss later on but I believe UI changes can be reverted

modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
moadqassem: @soer3n please revert this and most likely we are gonna revert any UI changes

modules/api/pkg/resources/machine/common.go
ahmedwaleedmalik: NamespacedMode is used to determine the infra namespace and has nothing to do with the UI. This however is a field that is present in the machine deployment.

modules/api/pkg/resources/machine/common.go
ahmedwaleedmalik: These sorts of changes should be covered by code, not some special documentation. We have not mentioned this in the [description of the field](https://github.com/kubermatic/kubermatic/blob/main/sdk/apis/kubermatic/v1/datacenter.go#L826-L828), that after enabling this, all your machine deployments need to be rotated.

Adding special cases for one customer that wouldn't really work for anyone else doesn't seem like the right way to go.

modules/api/pkg/resources/machine/common.go
moadqassem: I am not sure how this is any different tbh. You have a field on the seed that effects the handling of cloud provider resources and where they should be located(VMs, Policies, storage classes, PVCs) and you have another field on the seed that changes handling of the VMs(shared or dedicated CPUs). Both handling is happening on the cloud provider level. So how this is any different? And also, NamespcaedMode effects VPCs and Subnets, which are also UI components, and yet no tooltip for that either! So why now we should start using these tooltips 😉

modules/api/pkg/resources/machine/common.go
moadqassem: Again, same explanation here! NamespacedMode is event further more dangerous than this and yet there are no warnings in the UI

ahmedwaleedmalik (approved):
/approve

modules/api/pkg/resources/machine/common.go
ahmedwaleedmalik: CPU is a field in the UI that is being influenced by this flag. What does NamespacedMode has anything to do with the UI? Not sure what your argument here is, and I can't spend more time on this. PR has been approved in whatever state it was.

PR #7253, opened 2025-04-02
Update to Go 1.24.2, golangci-lint 2.x

xrstf:
This bumps the repository to Go 1.24 and the new golangci-lint. I migrated its config and adjusted the code now that staticcheck's integration has been improved and more of it became visible. In contrast to KKP, I disabled ST1003 though, which would have also caused a lot of variable renaming here, but I did not want to "force" this upon the dashboard, but if you want, I can re-enable it and adjust the code for it as well.

I also noticed that `pkg/provider/watcher/` was a 100% duplicate of `pkg/watcher/` and was entirely unused, so I simply deleted it :grin: 

**What type of PR is this?**
/kind chore

```release-note
Update to Go 1.24.2
```

**Documentation**:
```documentation
NONE
```


modules/api/.golangci.yml
```diff
@@ -53,67 +51,91 @@ linters:
...
+          alias: semverlib
+      no-unaliased: true
+    staticcheck:
+      checks:
+        - all
+        # disable "error strings should not be capitalized":
+        # but often errors start with k8s kinds and it's annoying to rephrase every error
+        - '-ST1005'
+        # very opinionated naming conventions
+        - '-ST1003'
```
SimonTheLeg, line 107: tiny nitpick from my side: Can we expand the comment, as to what conventions we don't like?
Asking because this check includes for different things https://staticcheck.dev/docs/checks#ST1003

modules/api/pkg/watcher/kubernetes/settings.go
```diff
@@ -67,9 +67,10 @@ func (watcher *SettingsWatcher) onEvent(delta toolscache.DeltaType, obj interfac
 	}
 
 	if settings != nil && settings.Name == kubermaticv1.GlobalSettingsName {
-		if delta == toolscache.Added || delta == toolscache.Updated {
+		switch delta {
+		case toolscache.Added, toolscache.Updated:
```
SimonTheLeg, line 71: This is slick. I am stealing this for my future code!

modules/api/.golangci.yml
SimonTheLeg: I misunderstood, you cannot choose between the 4 different ones. It is a all or nothing rule.

SimonTheLeg (approved):

Waseem826:
/hold
To avoid retries

PR #7257, opened 2025-04-04
Cherry pick 7188 7248 7252

soer3n:
This is a manual cherry-pick of https://github.com/kubermatic/dashboard/pull/7188, https://github.com/kubermatic/dashboard/pull/7248 and https://github.com/kubermatic/dashboard/pull/7252

The first two cherry-picked pull requests are requirements for the third which is the one which needs to be backported. This is due to the reason that the sdk and the root package of a repository can only be imported in the same version when the sdk package is imported with a path replacement. Therefore we need to ensure that the change regarding sdk standalone module for kkp and machine-controller package is included.

/assign soer3n


Waseem826:
Instead of cherrypicking the changes of `go.mod` from `main` (which requires all these sdk changes), we should point `k8c.io/kubermatic/v2` to latest commit of `release/v2.27` branch after https://github.com/kubermatic/kubermatic/pull/14319 is merged.

soer3n:
> Instead of cherrypicking the changes of `go.mod` from `main` (which requires all these sdk changes), we should point `k8c.io/kubermatic/v2` to latest commit of `release/v2.27` branch after [kubermatic/kubermatic#14319](https://github.com/kubermatic/kubermatic/pull/14319) is merged.

This will not work due to the reason explained in the description. Only pointing to the latest kkp version will break all existing kkp api imports. I tried that locally with a path replacement in the api go.mod file and cherry picking only the last pull request.

So using for instance `k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1` is not possible when using a kkp version which has already the sdk change implemented. These required rewrites are part of the other cherry-picks. The new field which is used for kubevirt requires this sdk already because it was merged after the sdk pr.

Waseem826:
> > Instead of cherrypicking the changes of `go.mod` from `main` (which requires all these sdk changes), we should point `k8c.io/kubermatic/v2` to latest commit of `release/v2.27` branch after [kubermatic/kubermatic#14319](https://github.com/kubermatic/kubermatic/pull/14319) is merged.
> 
> This will not work due to the reason explained in the description. Only pointing to the latest kkp version will break all existing kkp api imports. I tried that locally with a path replacement in the api go.mod file and cherry picking only the last pull request.
> 
> So using for instance `k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1` is not possible when using a kkp version which has already the sdk change implemented. These required rewrites are part of the other cherry-picks. The new field which is used for kubevirt requires this sdk already because it was merged after the sdk pr.

KKP SDK doesn't exist in 2.27. What I meant was keep the other file changes as it is and only update the go.mod according to 2.27 branch.

soer3n:
> > This will not work due to the reason explained in the description. Only pointing to the latest kkp version will break all existing kkp api imports. I tried that locally with a path replacement in the api go.mod file and cherry picking only the last pull request.
> > So using for instance `k8c.io/kubermatic/v2/pkg/apis/kubermatic/v1` is not possible when using a kkp version which has already the sdk change implemented. These required rewrites are part of the other cherry-picks. The new field which is used for kubevirt requires this sdk already because it was merged after the sdk pr.
> 
> KKP SDK doesn't exist in 2.27. What I meant was cherrypick the other file changes as it is and only update the go.mod according to 2.27 branch.

Ok. This should work when i do the same on the cherry-pick for the machine-controller which is also imported and requires a manual cherry pick due to conflicts. I will update this and the machine-controller pr. Thanks for the tip.

soer3n:
@Waseem826 I opened a new pull request without including the sdk changes and to keep the branch name pattern for cherry picks. Could you take a look at https://github.com/kubermatic/dashboard/pull/7264, please?

PR #7273, opened 2025-04-09
Add Input restrictions and escape values to avoid rendering as HTML in dialogs

### LINKED ISSUE #7269: HTML injection prevention
This issue was seen originally in the cluster RBAC section where end user can input any arbitrary characters (including HTML) while adding Users and Groups. Moreover, if the user enters HTML characters then that HTML is rendered in the DOM inside delete user/group confirmation dialog.

Image

This happens because we render the inputs using `innerHTML` in confirmation dialog without sanitizating/escaping HTML. Although Angular has [built-in sanitization](https://angular.dev/best-practices/security#angulars-cross-site-scripting-security-model) when we use `innerHTML` attribute which can prevent most basic scripting or XSS attacks but some advanced scripting techniques might go unchecked.

This problem should also be checked in other places where we use similar rendering approach and can be fixed by adding some validations on inputs where possible along with sanitizing/escaping HTML before rendering in the DOM.

Waseem826:
This PR fixes an issue where a value containing HTML characters was being rendered as HTML in Confirmation Dialog. This has been fixed by escaping the values to convert any HTML characters into HTML entities. As a result this would output the values as text rather than being executed as HTML.

Screenshot 2025-04-11 at 9 47 40 PM

In addition to that, this PR also adds pattern validator on value inputs (which can possibly be rendered as HTML) to prevent certain special characters (``| " < > { } [ ] ` \ ' ; &``) from being input.

Screenshot 2025-04-11 at 9 29 19 PM

Fixes #7269

**What type of PR is this?**

/kind bug

Since we use `innerHTML` to display the message in Confirmation Dialog (and other places), Angular has built-in DOM sanitization which prevents most common cross-site scripting attacks. I chose to convert the HTML characters into HTML entities instead of using Angular Dom Sanitizer's `sanitize` method because:
- Angular uses the same method internally to sanitize the DOM.
- `sanitize` method would alter the value by adding/removing any HTML code blocks.
- Value would still be rendered/executed as HTML.

```release-note
Add special characters restriction on Inputs and escape values to avoid rendering as HTML.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/backup/details/automatic-backup/component.ts
+++ b/modules/web/src/app/backup/details/automatic-backup/component.ts
@@ -15,7 +15,7 @@
 import {Component, OnDestroy, OnInit} from '@angular/core';
 import {MatDialog, MatDialogConfig} from '@angular/material/dialog';
 import {ActivatedRoute, Router} from '@angular/router';
-import {HealthStatus, getBackupHealthStatus} from '@app/shared/utils/health-status';
+import {getBackupHealthStatus, HealthStatus} from '@app/shared/utils/health-status';
 import {BackupService} from '@core/services/backup';
 import {ProjectService} from '@core/services/project';
 import {UserService} from '@core/services/user';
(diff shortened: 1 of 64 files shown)


ahmadhamzh (approved):
/approve

PR #7283, opened 2025-04-11
Prevent installing autoscaler if not defined in app definitions.

### LINKED ISSUE #7216: Hide the "Enable Cluster Autoscaler Application" option when the AppDefinition is empty.
### What happened

In the wizard, on the initial node step, when the user enables the cluster autoscaler and the AppDefinition is empty, an error appears: `ApplicationDefinition.apps.kubermatic.k8c.io "cluster-autoscaler" not found`  This can be confusing for users.
ref
### Expected behavior
If the "cluster-autoscaler" app can't be installed, we should either hide the option to enable the cluster autoscaler or warn the user that the app does not exist.
### Environment
- Version: 2.27

ahmadhamzh:
Prevent users from enabling the Cluster Autoscaler option if the corresponding application is not defined in the application definitions.

image

Fixes #7216 

**What type of PR is this?**
/kind bug

```release-note
 Disable the Cluster Autoscaler option when the cluster autoscaler application is not defined in applications catalog.
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/node-data/component.ts
+++ b/modules/web/src/app/node-data/component.ts
@@ -45,7 +45,7 @@ import {QuotaWidgetComponent} from '@dynamic/enterprise/quotas/quota-widget/comp
 import {QuotaCalculationService} from '@dynamic/enterprise/quotas/services/quota-calculation';
 import {DynamicModule} from '@dynamic/module-registry';
 import {AutocompleteControls} from '@shared/components/autocomplete/component';
-import {CLUSTER_AUTOSCALING_APP_DEF_NAME} from '@shared/entity/application';
+import {ApplicationDefinition, CLUSTER_AUTOSCALING_APP_DEF_NAME} from '@shared/entity/application';
 import {Datacenter} from '@shared/entity/datacenter';
 import {NodeNetworkSpec, OperatingSystemSpec, Taint} from '@shared/entity/node';
 import {OperatingSystemProfile} from '@shared/entity/operating-system-profile';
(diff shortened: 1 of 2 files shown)


modules/web/src/app/node-data/component.ts
```diff
@@ -238,6 +240,13 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       }
     }
 
+    if (!this._isClusterAutoscalingAppExist()) {
+      this.form.get(Controls.EnableClusterAutoscalingApp).setValue(false);
+      this.form.get(Controls.EnableClusterAutoscalingApp).disable();
+      this.autoscalingTooltipText =
+        'To enable autoscaling, the Cluster Autoscaler application must be added to the application definitions.';
```
Waseem826, line 247:
```suggestion
        'To enable autoscaling, the Cluster Autoscaler application must be added to the applications catalog.';
```

modules/web/src/app/node-data/component.ts
```diff
@@ -676,4 +685,10 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       replicas: this._nodeDataService.nodeData.count,
     };
   }
+
+  private _isClusterAutoscalingAppExist(): boolean {
+    return !!this._applicationService.applicationDefinitions.find(
+      app => app.name === 'CLUSTER_AUTOSCALING_APP_DEF_NAME'
```
Waseem826, line 691: `CLUSTER_AUTOSCALING_APP_DEF_NAME` is a constant so it should be treated as a string value.
```suggestion
      app => app.name === CLUSTER_AUTOSCALING_APP_DEF_NAME
```

modules/web/src/app/node-data/template.html
```diff
@@ -249,12 +249,14 @@
         </mat-card-title>
       </mat-card-header>
 
-      <mat-checkbox *ngIf="!isDialogView()"
-                    [name]="Controls.EnableClusterAutoscalingApp"
-                    [formControlName]="Controls.EnableClusterAutoscalingApp">Enable Cluster Autoscaler Application
+      <div fxLayoutAlign=" center">
+        <mat-checkbox *ngIf="!isDialogView()"
```
Waseem826, line 253: This `*ngIf` should be on parent div otherwise tooltip will be visible in dialog view.

Waseem826 (approved):
/approve

PR #7299, opened 2025-04-30
(fix): OpenStack - pass ConfigDrive value to JSON patch during machine updates

buraksekili:
This PR fixes a logic where configDrive is set to null during updates; which causes extra machines to rollout since the JSON patch detects difference between `null` and boolean value.

Fixes https://github.com/kubermatic/kubermatic/issues/14215
**What type of PR is this?**

/kind bug

```release-note
Pass ConfigDrive value to JSON patch during machine updates  for OpenStack
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/machine/convert.go
+++ b/modules/api/pkg/machine/convert.go
@@ -215,6 +215,9 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 		if config.RootDiskSizeGB != nil && *config.RootDiskSizeGB > 0 {
 			cloudSpec.Openstack.RootDiskSizeGB = config.RootDiskSizeGB
 		}
+		if cd := config.ConfigDrive.Value; cd != nil {
+			cloudSpec.Openstack.ConfigDrive = *cd
+		}
 	case providerconfig.CloudProviderHetzner:
 		config := &hetzner.RawConfig{}
 		if err := json.Unmarshal(decodedProviderSpec.CloudProviderSpec.Raw, &config); err != nil {
--- a/modules/api/cmd/kubermatic-api/swagger.json
+++ b/modules/api/cmd/kubermatic-api/swagger.json
@@ -37614,6 +37614,11 @@
           "type": "string",
           "x-go-name": "AvailabilityZone"
         },
+        "configDrive": {
+          "description": "ConfigDrive enables a configuration drive that will be attached to the instance when it boots.",
+          "type": "boolean",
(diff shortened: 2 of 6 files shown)


Waseem826:
Does this require a UI change?

buraksekili:
No, I would say. Currently, while scaling up/down the machine deployments for OpenStack through UI, since `configDrive` is an unknown field for Dashboard, the JSON marshaling sets it to `null`, which causes conflicts if `configDrive` is not set to null previously.


This PR adds `configDrive` to Dashboard API logic and allows API to recognize `configDrive`.

Waseem826:
> No, I would say. Currently, while scaling up/down the machine deployments for OpenStack through UI, since `configDrive` is an unknown field for Dashboard, the JSON marshaling sets it to `null`, which causes conflicts if `configDrive` is not set to null previously.
> 
> This PR adds `configDrive` to Dashboard API logic and allows API to recognize `configDrive`.

Can't we solve this by using `omitempty`?

buraksekili:
@Waseem826 makes sense; however, the root cause of the issue is that `configDrive` field is not yet in the Dashboard API. Therefore, there is no field currently to set `omitempty`

modules/api/pkg/machine/convert.go
```diff
@@ -215,6 +215,9 @@ func GetAPIV2NodeCloudSpec(machineSpec clusterv1alpha1.MachineSpec) (*apiv1.Node
 		if config.RootDiskSizeGB != nil && *config.RootDiskSizeGB > 0 {
 			cloudSpec.Openstack.RootDiskSizeGB = config.RootDiskSizeGB
 		}
+		if cd := config.ConfigDrive.Value; cd != nil {
```
Waseem826, line 218: I meant removing this and just using `omitempty`. Could that work?

buraksekili: Here the `config` comes from the machine-controller (reference: https://github.com/kubermatic/machine-controller/blob/a9dc4e656d96ab9a29f4c7e34fbc7ad3c79237d2/sdk/cloudprovider/openstack/types.go#L55).

The `ConfigDrive` value is a Go struct called [`providerconfig.ConfigVarBool`](https://github.com/kubermatic/machine-controller/blob/a9dc4e656d96ab9a29f4c7e34fbc7ad3c79237d2/sdk/providerconfig/types.go#L274) that is used by other fields in machine-controller as well; so, adding omitempty may cause unexpected results in other fields in my opinion.

Waseem826 (approved):
/approve

PR #7301, opened 2025-05-02
Support KubeVirt Subnet and StorageClasses Location Compatibility 

moadqassem:
This PR adds the support for filtering out KubeVirt VPC subnets based on the location of both the subnet itself and the the storage class that is being used. For example, if a storage class exists in region eu-central in a zone b then only subnets that are existing in region eu-central in the zone b will be listed. In order to activate such a feature, users must enable the property `matchSubnetAndStorageLocation: true` under the kubevlrt cloud provider. 

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Support infra storage classes and provider network subnets location compatibilities
```

**Documentation**:

```documentation
None. 
```


Waseem826 (approved):
/approve

Waseem826 (approved):

PR #7305, opened 2025-05-05
Make subnets required when a VPC is selected

### LINKED ISSUE #7278: Enforce subnet / storage class zone matching in KKP UI

Currently, the subnet selection on the "Initial Nodes" during cluster creation and additional Machine Deployment creation gives users the incorrect impression that subnet selection is optional. However, in some environments, subnet selection is mandatory.

Additionally, the UI does not enforce matching between selected subnet zones and the topology zones of the chosen storage class for VMs. This mismatch can lead to configuration errors.

#### Acceptance Criteria

- The KKP UI clearly indicates when subnet selection is mandatory
- Users are forced to select a subnet during cluster and Machine Deployment creation
- UI validation informs the user when there's a mismatch between subnet zones and storage class zones.

Internal reference: 8041

ahmadhamzh:
Make the Subnets field required in both the initial node settings (Wizard mode) and the Machine Deployment dialog when a VPC is selected.

Fixes #7278

**What type of PR is this?**
/kind bug

```release-note
Make the Subnets field required when a VPC is selected, in both Wizard and Machine Deployment modes.
```

**Documentation**:
```documentation
NONE
```


modules/web/src/app/node-data/basic/provider/kubevirt/template.html
```diff
@@ -114,8 +114,8 @@
                [options]="subnets"
                [formControlName]="Controls.Subnet"
                [label]="subnetLabel"
-               hint="Optional: Use specific subnet for the machines. VPC needs to be selected first to use this option."
-               [required]="false"
+               hint="Use specific subnet for the machines. VPC needs to be selected first to use this option."
```
Waseem826, line 117: This description looks wrong. It says VPC is required for this option so should we disable this when VPC is not selected?

Waseem826:
Apart from description, rest of the PR looks fine to me. It would be great if we can reword the description in this PR otherwise it would require another cherrypick.

modules/web/src/app/node-data/basic/provider/kubevirt/template.html
ahmadhamzh: Right now, if no VPC is selected, there are no subnets, so the field is always empty and not required. The label says "No subnets available or VPC not selected," which I thought would make it clear to the user that they need to select a VPC to see subnets. So I figured maybe there's no need to disable the field, since it has no subnets anyway.

ahmadhamzh: i wanted to remove the `Optional` cause if VPC is selected then it's not optional

Waseem826 (approved):
/approve

PR #7308, opened 2025-05-06
[release/v2.26] Bump KKP Dashboard version and dependencies for May 2025

archups:
This PR is to bump KKP Dashboard version and dependencies for KKP patch release v2.26.8 - May 2025.

Fixes #

**What type of PR is this?**
/kind documentation

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```


modules/api/go.mod
```diff
@@ -1,8 +1,8 @@
 module k8c.io/dashboard/v2
 
-go 1.22.8
+go 1.23.0
```
Waseem826, line 3: Why do we need to bump the go version in release branch?

xrstf: This happened because of a Cilium update.

xrstf: Arguably by accident, but the cookie has crumbled and now it's too late. For the future we should/will be stricter about allowing Go minor upgrades in release branches, I hope.

PR #7318, opened 2025-05-11
Add support for global viewer role in middleware

rajaSahil:
- Adds support for the `globalViewer` role in Kubermatic.
- Users with `globalViewer: true` in their User CR are automatically granted read-only access to all projects and clusters. This is achieved through dynamic group and role injection in the API middleware, without the need for explicit UserProjectBinding resources.

Fixes #

**What type of PR is this?**
/kind feature

- Global viewers are now resolved dynamically in the `UserInfo` middleware layer.
- `isGlobalViewer` and `isAdmin` are mutually exclusive, validated by webhook.

```release-note
Users marked with globalViewer: true now receive read-only access to all projects and clusters via dynamic groups and roles injection. No need to create UserProjectBindings for them.
```

**Documentation**:

```documentation
TBD
```

### DIFF
--- a/modules/api/pkg/handler/v1/project/project.go
+++ b/modules/api/pkg/handler/v1/project/project.go
@@ -259,7 +259,7 @@ func ListEndpoint(userInfoGetter provider.UserInfoGetter, projectProvider provid
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
 
-		if (req.DisplayAll && userInfo.IsAdmin) || (userInfo.IsGlobalViewer) {
+		if req.DisplayAll && (userInfo.IsAdmin || userInfo.IsGlobalViewer) {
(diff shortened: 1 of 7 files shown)


modules/api/pkg/handler/v1/project/project.go
```diff
@@ -259,7 +259,7 @@ func ListEndpoint(userInfoGetter provider.UserInfoGetter, projectProvider provid
 			return nil, common.KubernetesErrorToHTTPError(err)
 		}
 
-		if req.DisplayAll && userInfo.IsAdmin {
+		if (req.DisplayAll && userInfo.IsAdmin) || (userInfo.IsGlobalViewer) {
```
ahmedwaleedmalik, line 262:
```suggestion
		if (req.DisplayAll && (userInfo.IsAdmin || userInfo.IsGlobalViewer) {
```

Shouldn't it be something like this?

ahmedwaleedmalik (approved):
/approve

@rajaSahil please create a testing ticket for this.

PR #7324, opened 2025-05-14
Disable user SSH-Key feature gate UI

### LINKED ISSUE #7279: Add flag to disable SSH-Key functionality in KKP UI
We need the ability to globally disable all SSH-Key related functionality in the KKP UI. When this feature is enabled, all references to SSH-Keys should be completely hidden from the user interface.

#### Solution
Introduce a new flag in `kubermaticsettings` called `disableSSHKeys`. When set to `true`, the UI should hide:

- The "SSH Keys" section under the "Access" menu.
- The "SSH Keys" dropdown, "+ Add SSH Key" button
- "Use SSH Key Agent" checkbox on Cluster Creation Wizard (screen #1).
- The "SSH Keys" information on the Cluster Overview page.

Internal reference: 8047

ahmadhamzh:
add hide user ssh-key functionality from dashboard if `disableUserSSHKey` is true
Fixes #7279 

**What type of PR is this?**
/kind feature

```release-note
Add new feature gate to disable User SSH key 
```

**Documentation**:

```documentation
TBD
```


modules/web/src/app/sshkey/template.html
```diff
@@ -14,7 +14,8 @@
 limitations under the License.
 -->
 
-<mat-card appearance="outlined"
+<mat-card *ngIf="isUserSshKeyEnabled"
```
Waseem826, line 17: I think it would be better if we redirect the user to project overview or 404 page. Wdyt?

ahmadhamzh: we don't do that right now with other pages that can be disable like `externalcluster` or `backubs`

ahmadhamzh: or do you suggest to have a guard like `isEnabledPath` that can check if the path is not enable redirect to a new not enabled page?

modules/web/src/app/core/services/auth/guard.ts
```diff
@@ -103,3 +104,23 @@ export class AuthzGuard implements CanActivate {
     return from(this._router.navigate([View.Projects]));
   }
 }
+
+@Injectable()
+export class isEnabledPathGuard implements CanActivate {
```
Waseem826, line 109: This is not related to authentication so please move this to a separate file, perhaps create a directory named `ssh-key` and move this in `guard.ts` file there. You can also move `ssh-key.ts` inside that directory to avoid confusion.
As for the naming, I would just call it `SSHKeyGuard`.

Waseem826 (approved):
/approve
🚀

Issue #7343, opened 2025-05-20
Handling backup source


csengerszabo:
**What would you like to be added? / User Story**:
As a KKP user
I want to know where a particular backup was created from
In order to restore them correctly.

**Details**:

- Implementing labels for backups onto the UI https://github.com/kubermatic/kubermatic/issues/14198
- We should know if a backup was uploaded or created from a cluster, and if it was created from a cluster, we should know which cluster was the source of the backup

PR #7345, opened 2025-05-20
fix(7343): add labels for backup objects

rajaSahil:
This PR adds `spec.labels` to ClusterBackup objects created via the KKP controllers. These labels help identify and distinguish controller-initiated backups from those manually uploaded via the UI. 

Ref #7343 

**What type of PR is this?**

/kind feature

```release-note
Cluster backups created by KKP controllers now include spec.labels to distinguish controller-initiated backups from those manually uploaded via the UI. 

```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/ee/clusterbackup/backup/handler.go
+++ b/modules/api/pkg/ee/clusterbackup/backup/handler.go
@@ -60,7 +60,11 @@ type clusterBackupBody struct {
 
 const (
 	UserClusterBackupNamespace = "velero"
-	backupOrigin               = "kkp-controllers"
+
+	projectIdKey    = "project-id"
+	clusterIdKey    = "cluster-id"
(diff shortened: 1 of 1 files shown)


modules/api/pkg/ee/clusterbackup/backup/handler.go
```diff
@@ -113,6 +124,14 @@ func CreateEndpoint(ctx context.Context, request interface{}, userInfoGetter pro
 	}, nil
 }
 
+func getLabels(backupOrigin, projectID, clusterID string) map[string]string {
```
Waseem826, line 127: Perhaps we should start them `kubermatic/` to clarify that these are added by KKP. Also, can we please create some constants for these keys instead of hardcoded string values?

Waseem826 (approved):
/approve

PR #7349, opened 2025-05-22
fix(build): fix build for post dashboard ce main

rajaSahil:
This PR fixes a build issue on the post-dashboard-ce-main pipeline. The problem was caused by an import of group-project-binding/handler, which is part of our `ee` code. Since the `ce` build doesn't include the `ee` build tag, all files in that package were excluded—causing the build to fail.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/provider/kubernetes/member.go
+++ b/modules/api/pkg/provider/kubernetes/member.go
@@ -287,7 +287,7 @@ func (p *ProjectMemberProvider) MapUserToRoles(ctx context.Context, user *kuberm
 
 	roles := sets.New[string]()
 	if user.Spec.IsAdmin {
-		roles.Insert("owners")
+		roles.Insert(provider.OwnersRole)
 		return roles, nil
 	}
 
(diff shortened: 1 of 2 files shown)


modules/api/pkg/provider/kubernetes/member.go
```diff
@@ -293,7 +292,7 @@ func (p *ProjectMemberProvider) MapUserToRoles(ctx context.Context, user *kuberm
 	}
 
 	if user.Spec.IsGlobalViewer {
-		roles.Insert(handler.ViewersRole)
+		roles.Insert("viewers")
```
ahmedwaleedmalik, line 295: Instead of moving away from constants, move https://github.com/kubermatic/dashboard/blob/main/modules/api/pkg/ee/group-project-binding/handler/handler.go#L47-L51 somewhere else so that they are accessible for CE/EE both.

ahmedwaleedmalik (changes requested):

ahmedwaleedmalik (approved):
Great job 👍 

/approve

PR #7369, opened 2025-05-26
Display KubeVirt Subnet CIDRs in UI

moadqassem:
This PR supports showing the kubevirt subnets CIDR in the UI. 

Fixes #

**What type of PR is this?**
/kind feature

```release-note
Display KubeVirt Subnet CIDRs in UI
```

**Documentation**:

```documentation
None
```

### DIFF
--- a/modules/api/pkg/provider/cloud/kubevirt/networks.go
+++ b/modules/api/pkg/provider/cloud/kubevirt/networks.go
@@ -18,6 +18,7 @@ package kubevirt
 
 import (
 	"context"
+	"fmt"
 
 	"k8c.io/dashboard/v2/pkg/provider/cloud/kubevirt/providernetworks/kubeovn"
 
@@ -56,7 +57,8 @@ func GetProviderNetworkSubnets(ctx context.Context, client ctrlruntimeclient.Cli
 
 	subnets := make([]string, 0, len(subs))
 	for _, subnet := range subs {
-		subnets = append(subnets, subnet.Name)
+		subnetName := fmt.Sprintf("%v (%v)", subnet.Name, subnet.CIDRBlock)
+		subnets = append(subnets, subnetName)
 	}
 
 	return subnets, nil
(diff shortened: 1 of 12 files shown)


modules/api/pkg/provider/cloud/kubevirt/networks.go
```diff
@@ -56,7 +57,8 @@ func GetProviderNetworkSubnets(ctx context.Context, client ctrlruntimeclient.Cli
 
 	subnets := make([]string, 0, len(subs))
 	for _, subnet := range subs {
-		subnets = append(subnets, subnet.Name)
+		subnetName := fmt.Sprintf("%v (%v)", subnet.Name, subnet.CIDRBlock)
```
Waseem826, line 60: I think we should handle this change on frontend because by changing value of `name`, frontend will send this exact value to backend which might cause problems.

Waseem826: Currently, only `name` field is sent in response by API so we'll need to add this `cidrblock` field to response.

moadqassem: Ah yeah good catch! Let me take care of it.

modules/web/src/app/node-data/basic/provider/kubevirt/template.html
```diff
@@ -119,7 +119,7 @@
                (changed)="onSubnetChange($event)"
```
Waseem826, line 119: We also need this `valueFormatter` if we want to display custom value for selected subnet. This would require a change in `getSubnetOptionName` method so it will be passed a `string` value of selected subnet name and then we need to find the subnet from `subnets` array based on this name. Here's a similar example: https://github.com/kubermatic/dashboard/blob/main/modules/web/src/app/node-data/basic/provider/aws/component.ts#L221

```suggestion
               (changed)="onSubnetChange($event)"
               [valueFormatter]="getSubnetOptionName.bind(this)"
```

PR #7389, opened 2025-06-02
Fix token cookie cleaning by passing path argument to delete method

Waseem826:
The `path` argument was missing while calling cookie delete method due to which token cookie was not being deleted. This would result in application using an invalid token hence rendering the dashboard useless until token cookie was deleted manually.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/core/services/auth/service.ts
+++ b/modules/web/src/app/core/services/auth/service.ts
@@ -190,10 +190,8 @@ export class Auth {
   }
 
   private _deleteToken(): void {
-    if (this._cookieService.check(this._cookie.token)) {
-      this._cookieService.delete(this._cookie.token);
-      return;
-    }
+    this._cookieService.delete(this._cookie.token, '/');
+    this._tokenService.token = null;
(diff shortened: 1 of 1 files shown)


modules/web/src/app/core/services/auth/service.ts
```diff
@@ -190,18 +190,15 @@ export class Auth {
   }
 
   private _deleteToken(): void {
-    if (this._cookieService.check(this._cookie.token)) {
-      this._cookieService.delete(this._cookie.token);
-      return;
-    }
+    this._cookieService.delete(this._cookie.token, '/');
+    this._tokenService.token = null;
```
KhizerRehan, line 194: I see Setter accepts string as an argument and it works because of not having ""strictNullChecks": false" by default should we update signature of setter to accept null too for safety reasons?

TokenService

```ts
set token(token: string | null) {
  ...
}
```

PR #7401, opened 2025-06-03
Kyverno bug fixes and enhancements

ahmadhamzh:

this PR fix: 

- Project scope policies should not be visible on the admin panel. By default admins can only add “Global” level policies.
- Policies on project level page limit usability because all policies are disabled. showing the status of policies
- Unable to add policy on a cluster because the “Add Policy” dialog doesn’t display any policies.
- Cluster scoped policy (using cluster label selector) is not visible on the cluster details page.
- Cluster and project label selector keys are not removed when editing a policy.

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


Waseem826 (approved):
/approve

PR #7413, opened 2025-06-10
add logic to handle new field for kv cpu configuration too

soer3n:

This pr adds logic to handle the new introduced field in kkp seed for handling cpu assignment to kubevirt virtual machines. The defaulting for these virtual machines is also changing with this pull requests to use `.domain.cpu` in kubevirt spec as the default instead of resrouce requests and limits which was the default in previous kkp versions.

ref: https://github.com/kubermatic/kubermatic/pull/14621

Fixes #

**What type of PR is this?**

/kind feature

```release-note
[Breaking Changes] The `usePodResourcesCPU` feature will replace `enableDedicatedCPUs` flag. In the time of deprecation both are taking effect but the new value will have more priority. When `enableDedicatedCPUs` is set to `false` which is also the default value, you need to set `usePodResourcesCPU` to `true` to keep the same behaviour as before for new created machines. If `enableDedicatedCPUs` was set to `true` nothing needs to be changed.
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/1889
```

### DIFF
--- a/modules/api/pkg/resources/machine/common.go
+++ b/modules/api/pkg/resources/machine/common.go
@@ -623,7 +623,7 @@ func GetKubevirtProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.Nod
 	// - EnableDedicatedCPUs defaults (false) to .resources.requests/limits
 	// - UsePodResourcesCPU defaults (false) to .domain.cpu
 	// as long as we have the deprecated and new flag we need to account for 4 cases:
-	// 
+	//
 	// | Old Flag (EnableDedicatedCPUs) | New Flag (UsePodResourcesCPU) | Source for KubeVirt VM CPU |
 	// |---------|---------|---------|
 	// | true | false | .domain.cpu |


modules/api/pkg/resources/machine/common.go
```diff
@@ -619,14 +619,23 @@ func GetKubevirtProviderConfig(cluster *kubermaticv1.Cluster, nodeSpec apiv1.Nod
...
-	if dc.Spec.Kubevirt.EnableDedicatedCPUs {
+	// defaulting for both flags regarding virtual machine cpu assignment:
+	// - EnableDedicatedCPUs defaults (false) to .resources.requests/limits
+	// - UsePodResourcesCPU defaults (false) to .domain.cpu
+	// as long as we have the deprecated and new flag we need to take care about 4 cases
+	// 1) old flag is set to true and new flag is set to false => .domain.cpu will be used in kubevirt vm
+	// 2) old flag is set to true and new flag is set to true => .domain.cpu will be used in kubevirt vm
+	// 3) old flag is set to false and new flag is set to false => .domain.cpu will be used in kubevirt vm
+	// 4) old flag is set to false and new flag is set to true => .resources.requests/limits will be used
+	// in kubevirt vm
```
KhizerRehan, line 630:
```suggestion
	// defaulting for both flags regarding virtual machine cpu assignment:
	// - EnableDedicatedCPUs defaults (false) to .resources.requests/limits
	// - UsePodResourcesCPU defaults (false) to .domain.cpu
	// as long as we have the deprecated and new flag we need to account for 4 cases:
	// 
	// | Old Flag (EnableDedicatedCPUs) | New Flag (UsePodResourcesCPU) | Source for KubeVirt VM CPU |
	// |---------|---------|---------|
	// | true | false | .domain.cpu |
	// | true | true | .domain.cpu |
	// | false | false | .domain.cpu |
	// | false | true | .resources.requests/limits |
```

If this improved comment can help in readability otherwise please ignore this comment.
thx

PR #7415, opened 2025-06-10
Preserve CustomDiskSize checkbox value and fields on Edit Machine Deployment

### LINKED ISSUE #7281: "Custom Disk" setting is not preserved on the edit MachineDeployment dialogue when not enforced
### What happened

When setting up a custom disk size on a machine deployment, the setting is not shown when opening the edit dialogue:

https://github.com/user-attachments/assets/ea7cf069-8a61-4a64-ae40-435cbab6a153

### Expected behavior

The "Custom Disk" checkbox should be checked once the asynchronous call to fetch the MD spec is resolved, as it happens with the rest of the settings. Once manually checking it, we can see that the size that was previously set is displayed on the "Disk Size in GB" input box.

### How to reproduce

We have a second environment that does not display this behaviour due to:

```yaml
spec:
[...]
  providerConfiguration:
    openStack:
      enforceCustomDisk: true
```

... being set on the `kubermaticsetting.yaml`. So, as long as this is not the case, the bug should reproduce.

### Environment
- UI Version: 2.27.2
- API Version: 2.27
- Cloud Provider: OpenStack

KhizerRehan:
This PR fixes to preserve "Custom Disk" checkbox in Edit Machine Deployment Dialog and display DiskSize field.

https://github.com/user-attachments/assets/6a0bf9e7-baf4-4099-a6ea-a74f169560a6

Fixes #7281 

**What type of PR is this?**
/kind bug

Checks presence of diskSize in case of "Custom Disk" option is selected

```release-note
Shows custom disk fields when a custom disk is configured in the Machine Deployment edit dialog.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/node-data/basic/provider/openstack/component.ts
+++ b/modules/web/src/app/node-data/basic/provider/openstack/component.ts
@@ -283,9 +283,11 @@ export class OpenstackBasicNodeDataComponent extends BaseFormValidator implement
         `PT${this._nodeDataService.nodeData.spec.cloud.openstack.instanceReadyCheckTimeout}`.toUpperCase()
       ).asSeconds();
 
+      const diskSize = this._nodeDataService.nodeData.spec.cloud.openstack.diskSize;
       this.form.get(Controls.UseFloatingIP).setValue(this._nodeDataService.nodeData.spec.cloud.openstack.useFloatingIP);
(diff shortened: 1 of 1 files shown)


Waseem826 (approved):
/approve

ahmedwaleedmalik (approved):
/approve

PR #7435, opened 2025-06-18
Default Backup Sync Period to 0 if unset or removed

### LINKED ISSUE #7222: ClusterBackupStorageLocation 'Backup Sync Period' should be mandatory / regex should not validate for `""`

While creating CBSLs from the web UI, the form defaults to 0s. Removing this value, but not filling it does not result on an error as it happens when validating the regex `MMmSSs`:

|   |  |
|---|---|

Submitting the form, raises the following 500 error from the `/api/v2/projects/<project>/clusterbackupstoragelocation`:

Image

When removing the default value from the 'Backup Sync Period' and moving to a different field:
-  The default value: `0` should be restored. 
-  The error message indicating that the regex does not validate for `""` should be displayed.
 
----

**Addidional question:** why the endpoint(s) is not a mandatory field too? Is there a scenario in which an S3 endpoint is not provided for a CBSL?

KhizerRehan:

This PR Default Backup Sync Period to 0 if unset or removed and make field as required

https://github.com/user-attachments/assets/cae5faf1-ab4f-4f34-a5c1-fe6db12f1143

Fixes #7222 

**What type of PR is this?**

/kind bug

```release-note
Backup Sync Period defaults to "0" if unset or removed in ClusterBackupStorageLocation.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/wizard/step/cluster/template.html
+++ b/modules/web/src/app/wizard/step/cluster/template.html
@@ -493,22 +493,20 @@ limitations under the License.
                           *ngIf="isKubernetesDashboardEnabled">
               Kubernetes Dashboard
             </mat-checkbox>
-            <div fxLayout="row"
-                 fxLayoutAlign=" center">
-              <mat-checkbox [formControlName]="Controls.OPAIntegration"
-                            id="km-wizard-opa-integration-checkbox"
-                            [matTooltipDisabled]="!form.get(Controls.OPAIntegration).disabled">
-              </mat-checkbox>
+
+            <mat-checkbox [formControlName]="Controls.OPAIntegration"
+                          id="km-wizard-opa-integration-checkbox"
+                          [matTooltipDisabled]="!form.get(Controls.OPAIntegration).disabled">
               <div fxLayout="row"
                    fxLayoutAlign=" center">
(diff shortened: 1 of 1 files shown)


modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -192,7 +193,7 @@ export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestro
           prefix: this.form.get(Controls.Prefix).value,
           caCert: this.form.get(Controls.CaCert).value,
         },
-        backupSyncPeriod: this.form.get(Controls.BackupSyncPeriod).value,
+        backupSyncPeriod: this.form.get(Controls.BackupSyncPeriod).value || null,
```
Waseem826, line 196: Since `0` is a valid value, `||` would set the value to `null`.

KhizerRehan: Good Catch @Waseem826

PR #7444, opened 2025-06-19
Set default backup sync period to null if value is empty

### LINKED ISSUE #7222: ClusterBackupStorageLocation 'Backup Sync Period' should be mandatory / regex should not validate for `""`

KhizerRehan:

This PR sets the default backupSyncPeriod to null if the form value is an empty string.

Fixes #7222 

**What type of PR is this?**

/kind bug

```release-note
Unset backup sync period if value is empty.
```

**Documentation**:

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #7453, opened 2025-06-25
KubeLB: Fix a bug with kubelb enforcement

ahmedwaleedmalik:
When KubeLB was enforced for a datacenter, we were not "enabling" KubeLB which results in it not being deployed for the cluster and users can't change it afterwards as well since it's a disabled control in the UI.

Fixes #

**What type of PR is this?**

/kind bug

```release-note
KubeLB: Fix a bug where enforcement on a datacenter was not enabling KubeLB for the user clusters in the dashboard.
```

**Documentation**:

```documentation
NONE
```


Waseem826 (approved):
/approve

PR #7427, opened 2025-06-13
Remove flakiness from ssh-keys integration tests

KhizerRehan:
**What this PR does / why we need it**:

 This PR fixes flakiness in ssh-keys integration tests

image

image

**Which issue(s) this PR fixes**:

Fixes #6940 

**What type of PR is this?**

/kind bug
/find failing-test

**Special notes for your reviewer**:

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

PR #7469, opened 2025-06-30
[release/v2.27] Remove flakiness from ssh-keys integration tests (#7427)

KhizerRehan:
This is an manual cherry-pick of #7427 

/assign KhizerRehan

Special notes for your reviewer:

Even after cherry-picking changes from #7427, I had to make further adjustments to get the config working under Cypress 13 with Node 20

Summary:

- Cypress v13: Fails on cypress.config.ts due to lack of native TypeScript support. Node.js 20 throws Unexpected token ':' on type annotations. Also requires CommonJS-safe plugin imports.

- Cypress v14: Uses esbuild internally to strip TypeScript, so the same config works without changes.

```release-note
NONE
```

### DIFF
--- a/modules/web/cypress.config.ts
+++ b/modules/web/cypress.config.ts
@@ -13,11 +13,12 @@
 // limitations under the License.
 
 import {defineConfig} from 'cypress';
-import failFast from 'cypress-fail-fast/plugin';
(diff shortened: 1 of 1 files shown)


Waseem826 (approved):
/approve

PR #7478, opened 2025-07-04
WIP Hide application managed by kkp from create user cluster wizard

ahmadhamzh:
hide the system applications (managed by KKP) from the application list in the create user cluster wizard

**What type of PR is this?**
/kind feature

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/shared/components/application-list/component.ts
+++ b/modules/web/src/app/shared/components/application-list/component.ts
@@ -320,7 +320,7 @@ export class ApplicationListComponent implements OnInit, OnChanges, OnDestroy {
 
   private get _visibleApplications(): Application[] {
     let filteredApplications = this.applications || [];
-    if (!this.showSystemApplications) {
+    if (!this.showSystemApplications || this.view !== ApplicationsListView.Default) {
       filteredApplications = filteredApplications.filter(application => !this.isSystemApplication(application));
     }
     return filteredApplications;


Waseem826:
Can we still edit auto-scaler application values in cluster creation wizard?

ahmadhamzh:
@Waseem826 actually that's why i did this change so users can't edit the system application (like auto-scaler application) in the wizard

Waseem826:
> > > > Editing cluster-autoscaler values was allowed in 2.27 so this change was made to continue same behaviour. Has something change now?
> > 
> > 
> > > 
> > 
> > 
> > > yeh i think we want to disallow that
> > 
> > 
> > > 
> > 
> > 
> > > This PR prevents showing/editing applications managed by KKP in the Create Cluster wizard. @moadqassem
> > 
> > 
> > I'm confused why we want to prevent this because users can also install system applications in existing clusters and they're allowed to edit default values.
> > Perhaps we should only make the YAML editor readonly? But then it should also be changed for adding applications in existing clusters. We'll also need to update the docs which mention editing default values of system applications, e.g. https://docs.kubermatic.com/kubermatic/v2.28/architecture/concept/kkp-concepts/applications/default-applications-catalog/cluster-autoscaler/
> 
> To idea here is, cluster autoscaler is a system application which means it shouldn't be shown in the application step similar how cilium is done. Regarding changing the values file, the user can actually do that, it's not a prerequisite

Users can edit the Cilium application values in cluster wizard (On 2nd step under Advanced Network Settings). If we want to hide the cluster-autoscaler application from wizard then IMO we should add another way (in the wizard) to update the values.

PR #7500, opened 2025-07-15
Add functionality to configure the Cluster Autoscaler application

### LINKED ISSUE #7495: Move the Cluster Autoscaler configuration into Initial Nodes step in Cluster Wizard
As a KKP User
I want to configure the Cluster Autoscaler at the Initial Nodes step, where I also switch it on
In order not to be confused by having to set it up among other Applications in the Applications step.

Internal reference: 8248

ahmadhamzh:
Move the configuration of the Cluster Autoscaler application to the initial node step in the user cluster wizard, and hide the system applications in the application step.

Screenshot from 2025-07-15 02-08-22

Fixes #7495

**What type of PR is this?**
/kind feature

```release-note
Add a new button in the initial node step of the user cluster wizard to configure the Cluster Autoscaler application.
```

**Documentation**:
```documentation
https://github.com/kubermatic/docs/pull/2013
```

### DIFF
--- a/modules/web/src/app/node-data/component.ts
+++ b/modules/web/src/app/node-data/component.ts
@@ -371,9 +371,7 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       .get(Controls.EnableClusterAutoscalingApp)
       .valueChanges.pipe(takeUntil(this._unsubscribe))
       .subscribe(enable => {
-        this._applicationService.applications = this._applicationService.applications.filter(
-          app => app.spec.applicationRef.name !== CLUSTER_AUTOSCALING_APP_DEF_NAME
-        );
+        this._filterOutAutoscalerApp();
         if (enable) {
           this._applicationService.applications = [
             ...this._applicationService.applications,
@@ -487,12 +485,10 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     dialog
       .afterClosed()
       .pipe(take(1))
-      .subscribe(app => {
+      .subscribe((app: Application) => {
         if (app) {
-          const apps = this._applicationService.applications.filter(
-            app => app.spec.applicationRef.name !== CLUSTER_AUTOSCALING_APP_DEF_NAME
-          );
-          this._applicationService.applications = [...apps, app];
+          this._filterOutAutoscalerApp();
+          this._applicationService.applications = [...this._applicationService.applications, app];
           this.autoscalerApplication = app;
         }
       });
@@ -731,5 +727,11 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
       ...this._quotaCalculationService.quotaPayload,
       replicas: this._nodeDataService.nodeData.count,
     };
+  }
+
+  private _filterOutAutoscalerApp(): void {
+    this._applicationService.applications = this._applicationService.applications.filter(
+      app => app.spec.applicationRef.name !== CLUSTER_AUTOSCALING_APP_DEF_NAME
+    );
   }
 }
--- a/modules/web/src/app/node-data/template.html
+++ b/modules/web/src/app/node-data/template.html
@@ -264,7 +264,7 @@ limitations under the License.
                 (click)="onEditAutoScalerApp()">
           <i class="km-icon-mask km-icon-edit"
              matButtonIcon></i>
-          <span>Edit Cluster Autoscaler App</span>
+          <span>Edit Application Values</span>
         </button>
       </div>
 


modules/web/src/app/node-data/component.ts
```diff
@@ -353,6 +366,21 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
...
+      .subscribe(enable => {
+        this._applicationService.applications = this._applicationService.applications.filter(
+          app => app.spec.applicationRef.name !== CLUSTER_AUTOSCALING_APP_DEF_NAME
+        );
+        if (enable) {
+          this._applicationService.applications = [
+            ...this._applicationService.applications,
+            this.autoscalerApplication,
+          ];
+        }
```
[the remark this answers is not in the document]
ahmadhamzh: I think we leave them because users can install this autoscaler themselves and not rely on our app(With this PR, they can't do that through the wizard anymore, but they still can from the cluster page), so maybe that's why we don’t remove them. This is how it was before, so I'm not sure if we want to change that.
@Waseem826 do you have any thoughts?

modules/web/src/app/wizard/step/applications/component.ts
```diff
@@ -98,32 +87,6 @@ export class ApplicationsStepComponent extends StepBase implements OnInit, OnDes
...
-        })
-      )
-      .subscribe(appDef => {
-        if (appDef) {
-          const application = this.createApplicationInstallation(appDef);
-          this.onApplicationAdded(application);
-        }
-      });
-
-    this._onApplicationsChanged();
```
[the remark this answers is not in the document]
ahmadhamzh: yes, cause we want to hide any system-applications (applications managed by KKP)

modules/web/src/app/node-data/component.ts
```diff
@@ -450,6 +478,26 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
     this.DNSServers = ips;
   }
 
+  onEditAutoScalerApp(): void {
```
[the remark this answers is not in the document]
ahmadhamzh: Yes, that’s correct because it’s a system application, so the disabled field needs to stay the same.

modules/web/src/app/node-data/component.ts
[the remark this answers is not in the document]
Waseem826: Lets keep the min/max replicas fields as it is.

modules/web/src/app/node-data/component.ts
```diff
@@ -691,7 +721,14 @@ export class NodeDataComponent extends BaseFormValidator implements OnInit, OnDe
...
   }
 
-  private _isClusterAutoscalingAppExist(apps: ApplicationDefinition[]): boolean {
-    return !!apps.find(app => app.name === CLUSTER_AUTOSCALING_APP_DEF_NAME);
+  private _filterOutAutoscalerApp(updateApplication: boolean, app: Application): void {
+    this._applicationService.applications = this._applicationService.applications.filter(
+      app => app.spec.applicationRef.name !== CLUSTER_AUTOSCALING_APP_DEF_NAME
+    );
+
+    if (updateApplication) {
```
Waseem826, line 729: This should not be part of this method since this method's purpose is to filter out the autoscaler app, not add it back.

modules/web/src/app/node-data/template.html
```diff
@@ -249,15 +249,26 @@
...
+          <i class="km-icon-info km-pointer"
+             [matTooltip]="autoscalingTooltipText"></i>
+        </div>
+        <button *ngIf="form.get(Controls.EnableClusterAutoscalingApp).value"
+                mat-flat-button
+                color="quaternary"
+                (click)="onEditAutoScalerApp()">
+          <i class="km-icon-mask km-icon-edit"
+             matButtonIcon></i>
+          <span>Edit Cluster Autoscaler App</span>
```
Waseem826, line 267: IMO this looks more appropriate since we're only allowing values to be edited.
```suggestion
          <span>Edit Application Values</span>
```

modules/web/src/app/shared/components/application-list/edit-application-dialog/component.ts
```diff
@@ -158,7 +154,7 @@ export class EditApplicationDialogComponent implements OnInit, OnDestroy {
         this._builder.control(
           {
             value: this.application.namespace,
-            disabled: this.isEnforcedApplication(),
+            disabled: this.isEnforcedApplication() ? true : isSystemApplication(this.application?.labels),
```
Waseem826, line 157: This will also prevent users from editing the values on cluster details page.

modules/web/src/app/wizard/component.ts
```diff
@@ -350,7 +355,12 @@ export class WizardComponent implements OnInit, OnDestroy {
         template.nodeDeployment.name = namePrefix;
         this._clusterSpecService.initializeClusterFromClusterTemplate(template);
         this._nodeDataService.initializeNodeDataFromMachineDeployment(template.nodeDeployment);
-        this.applicationService.applications = template.applications;
+        template.applications.forEach(app => {
+          if (app.spec.applicationRef.name === CLUSTER_AUTOSCALING_APP_DEF_NAME) {
+            (app.labels ??= {})[ApplicationLabel.ManagedBy] = ApplicationLabelValue.KKP;
+          }
```
Waseem826, line 361: Why do we need this? Isn't `labels` already part of ApplicationInstallation object?

Waseem826 (changes requested):
Sorry for duplicate review but I noticed few things that might be important to address.

modules/web/src/app/shared/components/application-list/edit-application-dialog/component.ts
ahmadhamzh: do we want them to edit these values ? in the cluster details page? cause it's system applications ?

modules/web/src/app/wizard/component.ts
ahmadhamzh: in cluster templates, applications are saved inside an annotation named: `kubermatic.io/initial-application-installations-request:` and it doesn't have the labels in the object

Waseem826: So it was broken previously?
Also, speaking of cluster templates, did you test if the updated cluster autoscaler values are persisted when editing/customizing cluster template?

modules/web/src/app/shared/components/application-list/edit-application-dialog/component.ts
Waseem826: Yes, we do. They can also edit Cilium application values.

ahmadhamzh: where ?
image

Waseem826: Sorry @ahmadhamzh you're correct. I confused this with something else.

PR #7513, opened 2025-07-29
Fix OpenStack network selection to handle networks without name

### LINKED ISSUE #7099: During cluster creation, pre-existing OpenStack networks with no name lead to UX problems
### What happened

If an OpenStack network has no name:
- the dropdown entry shows no text at all
- no subnet can be chosen, if this network is chosen

### Expected behavior

If an OpenStack network has no name:
- it's ID (maybe in brackets, like OpenStack Horizon does) should be shown
- the subnet query should still work (probably also needs to use the ID instead of the name)

### How to reproduce

1. Create network and subnet in OpenStack. Do not provide give the network a name.
2. Create a UC and chose a pre-existing network
3. Notice the created network is listed with no text
4. Pick the created network
5. Notice that the subnet list is empty

### Environment
- UI Version: v2.26.2
- API Version: v2.26.2
- Domain: ?
- Others:

### Current workaround

Make sure pre-existing networks have a name

### Affected user persona

End Users creating a UC with custom OpenStack settings

### Screenshots

Network with no name in OpenStack Horizon (middle one):

Image

And the same (highlighted) in KKP:

Image

No subnets loaded if that network was chosen:

Image

KhizerRehan:
This PR resolves issues with unnamed OpenStack networks by displaying their ID in the dropdown and properly loading associated subnets.

**Before**
image

**After**
**Network**
Screenshot 2025-07-31 at 10 10 05 AM

**Subnets**
Screenshot 2025-07-31 at 10 10 44 AM

Fixes #7099 

**What type of PR is this?**

/kind bug

API Provides a fallback mechanism in case it fails to list by name it search network with network ID which is always generated by openstack network provider.

```release-note
Fixes network selection to display network ID when name is missing in OpenStack.
```

**Documentation**:

```documentation
tbd
```

### DIFF
--- a/modules/api/pkg/provider/cloud/openstack/helper.go
+++ b/modules/api/pkg/provider/cloud/openstack/helper.go
@@ -160,18 +160,39 @@ func getTenants(authClient *gophercloud.ProviderClient, region string) ([]osproj
 }
 
 func getSubnetForNetwork(netClient *gophercloud.ServiceClient, networkIDOrName string) ([]ossubnets.Subnet, error) {
-	var allSubnets []ossubnets.Subnet
-
-	networks, err := getAllNetworks(netClient, osnetworks.ListOpts{Name: networkIDOrName})
-	if err != nil {
-		return nil, fmt.Errorf("failed to list networks: %w", err)
+	findNetwork := func(opts osnetworks.ListOpts, searchBy string) (string, error) {
+		networks, err := getAllNetworks(netClient, opts)
+		if err != nil {
+			return "", fmt.Errorf("failed to list networks by %s: %w", searchBy, err)
+		}
+		if len(networks) == 1 {
+			return networks[0].ID, nil
+		}
+		if len(networks) > 1 {
+			return "", fmt.Errorf("got %d networks for %s '%s', expected one at most", len(networks), searchBy, networkIDOrName)
+		}
+		return "", nil
 	}
 
-	networkID := networkIDOrName
-	if len(networks) == 1 {
-		networkID = networks[0].ID
-	} else if len(networks) > 1 {
-		return nil, fmt.Errorf("got %d networks for idOrName '%s', expected one at most", len(networks), networkIDOrName)
+	const (
+		searchByName = "name"
+		searchByID   = "id"
+	)
+
+	// Try by name first
+	networkID, err := findNetwork(osnetworks.ListOpts{Name: networkIDOrName}, searchByName)
+	if err != nil {
+		return nil, err
+	}
+	if networkID == "" {
+		// Fallback to ID
+		networkID, err = findNetwork(osnetworks.ListOpts{ID: networkIDOrName}, searchByID)
+		if err != nil {
+			return nil, err
+		}
+		if networkID == "" {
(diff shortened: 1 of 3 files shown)


modules/web/src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
```diff
@@ -275,6 +292,12 @@ export class OpenstackProviderExtendedCredentialsComponent
   ngOnDestroy(): void {
     this._unsubscribe.next();
     this._unsubscribe.complete();
+    this._networkChangeSubject.complete();
+  }
+
+  networkDisplayName(id: string): string {
+    const network: OpenstackNetwork = this.networks.find(network => network.id === id);
+    return network ? network.name || network.id : '';
```
Waseem826, line 300: We should display the ID in `(ID: <network_id>)` format.

```suggestion
    return network ? network.name || `(ID: ${network.id})` : '';
```

modules/web/src/app/wizard/step/provider-settings/provider/extended/openstack/default/component.ts
```diff
@@ -111,6 +112,7 @@ export class OpenstackProviderExtendedCredentialsComponent
   private _applicationCredentialID = '';
   private _applicationCredentialSecret = '';
   private _preset = '';
+  private _currentNetworkId = '';
```
Waseem826, line 115: I can't find where're we using it for combobox.

modules/web/src/app/wizard/step/provider-settings/provider/extended/openstack/default/template.html
```diff
@@ -41,10 +41,12 @@
...
                (changed)="onNetworkChange($event)"
                [label]="networksLabel"
+               [valueFormatter]="networkDisplayName.bind(this)"
                inputLabel="Select network..."
-               filterBy="name">
+               filterBy="name"
+               selectBy="id">
     <div *option="let network">
-      {{network.name}}
+      {{network.name || network.id}}
```
Waseem826, line 49: We should display the ID in (ID: <network_id>) format here as well.
```suggestion
      {{network.name || network.id}}
```

Waseem826 (changes requested):
Some feedback. PTAL

modules/api/pkg/provider/cloud/openstack/helper.go
```diff
@@ -160,26 +160,48 @@ func getTenants(authClient *gophercloud.ProviderClient, region string) ([]osproj
...
+
+	const (
+		searchByName = "name"
+		searchByID   = "id"
+	)
 
-	networks, err := getAllNetworks(netClient, osnetworks.ListOpts{Name: networkIDOrName})
+	// Try by name first
+	networkID, err := findNetwork(osnetworks.ListOpts{Name: networkIDOrName}, searchByName)
 	if err != nil {
```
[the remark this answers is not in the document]
Waseem826: Ok. I thought it would throw an error if it can't find network by name but if it returns an empty list then its fine.

Waseem826 (approved):
/approve

PR #7543, opened 2025-09-03
[release/v2.26] Bump Go build image version

archups:
This PR is to bump Go build image version.

Fixes #

**What type of PR is this?**
/kind chore

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/.prow/api.yaml
+++ b/.prow/api.yaml
@@ -32,7 +32,7 @@ presubmits:
       preset-goproxy: "true"
     spec:
       containers:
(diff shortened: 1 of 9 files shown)


archups:
> Is there a breaking change which requires this upgrade? Also, Angular version in 2.26 doesn't work with Node 22.

Node version is not bumped into this Go build image. It is just the version update solution opted to avoid conflict in the build image version with existing one. Node version is still 20

archups:
@Waseem826 Build image tagging issue is fixed. PTAL. Thanks!

PR #7639, opened 2025-10-07
Fix cluster template bugs

### LINKED ISSUE #7632: Cluster template related bugs
### What happened

**GCP :**

when customize or editing a GCP cluster template, disk type and machine type values are not loaded

Image

**OS :**
"Skip Router Reconciliation " checkbox state is not persisted in cluster template which i think needs to be fixed.

Image

ahmadhamzh:
* fix disk types and machine types values are not loaded in cluster template
* fix "Skip Router Reconciliation " checkbox state is not persisted in cluster template

Fixes #7632 

**What type of PR is this?**
/kind bug

```release-note
Fix disk types and machine types values are not loaded in cluster template for Google Cloud Provider
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/node-data/basic/provider/gcp/component.ts
+++ b/modules/web/src/app/node-data/basic/provider/gcp/component.ts
@@ -29,7 +29,7 @@ import {NodeDataService} from '@core/services/node-data/service';
 import {DynamicModule} from '@app/dynamic/module-registry';
 import _ from 'lodash';
 import {merge, Observable} from 'rxjs';
-import {filter, map, switchMap, take, takeUntil, tap} from 'rxjs/operators';
+import {filter, map, switchMap, takeUntil, tap} from 'rxjs/operators';
 import {FilteredComboboxComponent} from '@shared/components/combobox/component';
 import {GCPNodeSpec, NodeCloudSpec, NodeSpec} from '@shared/entity/node';
 import {GCPDiskType, GCPMachineSize, GCPZone} from '@shared/entity/provider/gcp';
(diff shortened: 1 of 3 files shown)


modules/web/src/app/node-data/basic/provider/gcp/component.ts
```diff
@@ -266,7 +262,6 @@ export class GCPBasicNodeDataComponent extends BaseFormValidator implements OnIn
     this.diskTypes = [];
     this.selectedDiskType = '';
     this.diskTypeLabel = DiskTypeState.Empty;
-    this._diskTypesCombobox.reset();
```
Waseem826, line 269: This method indirectly triggers the `onDiskTypeChange` method so after removing this, it might not clear the `this._nodeDataService.nodeData.spec.cloud.gcp.diskType` value when you change the zone. Can you please verify that it still works?

Waseem826 (commented):
Some feedback. PTAL

modules/web/src/app/node-data/basic/provider/gcp/component.ts
ahmadhamzh: yeh i test it and it's working as expected, i will update some changes for the edit dialog
i don't see a reason to clear the value, we just need to update it with the default value, and the `_setDefaultDiskType` and `_setDefaultMachineType` are trigered when zone changes and the `onDiskTypeChange` and `onMachineTypeChange` method update the  `this._nodeDataService.nodeData.spec.cloud.gcp.diskType`

Waseem826 (approved):
/approve

PR #7652, opened 2025-10-16
UI/UX Improvments

### LINKED ISSUE #7618: UI/UX improvements
Please comment or update the description with all the UI/UX related improvements/issues which are found during release testing for 2.29 release.

ahmadhamzh:

* **project quota widget :**

Clusters List Page
image

Machine Deployments List:
image

* **Text should fits inside panel:**

image

* **display "Config Drive" status in Node details:**

image

* **hide the "Enable Cluster Autoscalar Application" and "Edit Application" options from Add/Edit machine deployment:**

image

* **Delete button in applications list in user cluster should be hidden for System applications:**

image

* **only enable Min/Max replicas fields when "Enable Cluster Autoscaler Application" is checked:**

[Screencast from 2025-10-16 15-05-37.webm](https://github.com/user-attachments/assets/9ac7a7f2-a04a-4df5-9d91-3d74778c63f0)

* **The Anexia Icon is not fully displayed in the summary step of the cluster wizard:**
image

* **a second scrollbar is displayed in certain providers**

image

* **can't create a cluster with 0 replicas**

[Screencast from 2025-10-16 15-19-15.webm](https://github.com/user-attachments/assets/4de85dbd-c723-4c69-b5e9-3ea1d1cac751)

* **Change preset linkages title:**

image

* **Which issue(s) this PR fixes**:
Fixes #7618

**What type of PR is this?**
/kind bug
/kind design

```release-note
NONe
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/cluster/details/cluster/machine-deployment-details/style.scss
+++ b/modules/web/src/app/cluster/details/cluster/machine-deployment-details/style.scss
@@ -60,3 +60,7 @@
 .quota-widget {
   width: 500px;
 }
+
+.km-content-wrap div:last-child {
+  height: fit-content;
+}
--- a/modules/web/src/app/cluster/details/cluster/node-list/template.html
+++ b/modules/web/src/app/cluster/details/cluster/node-list/template.html
@@ -270,6 +270,10 @@ limitations under the License.
               <div key>Instance Ready Check Timeout</div>
               <div value>{{element.spec.cloud.openstack.instanceReadyCheckTimeout}}</div>
             </km-property>
+            <km-property-boolean *ngIf="element.spec.cloud.openstack"
+                                 label="Config Drive"
+                                 [value]="element.spec.cloud.openstack?.configDrive">
+            </km-property-boolean>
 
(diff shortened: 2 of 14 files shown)


modules/web/src/app/shared/components/labels/component.ts
```diff
@@ -67,6 +69,20 @@ export class LabelsComponent implements OnInit, OnChanges {
     return this.oneLineLimit && this.hideExtraLabels;
   }
 
+  getLabelKey(label: string): string {
+    if (label.length > this._maxLabelLength) {
```
KhizerRehan: I don't have any better solution ATM but i am thinking in case label/value with 21 char ..... so we will be shwoing 20 char... and then but tooltip shows full value which make sense.

TBH, If we should handle with CSS or making more dynamic as "@Input"  but for now it works fine

  private readonly _maxLabelLength = 20;
  private readonly _displayedLength = 17;


Approach-2

OR  can we Add Input `@Input() disableTruncation` and it should work for specific componenent where needed otherwise default functionality should work?


thoughts?

PR #7653, opened 2025-10-16
Improve encryption key validation for cluster encryption at rest

KhizerRehan:
This PR fixes the encryption key validation to enforce proper base64 format and 32-byte length requirements for cluster encryption at rest configuration.when users enter invalid encryption keys in both  the cluster creation wizard and cluster edit forms.

[Ref Comment: ](https://github.com/kubermatic/kubermatic/issues/15009#issuecomment-3405203150)

Screenshot 2025-10-16 at 6 52 28 PM

Fixes In case User provide In-Valid Encryption Key 

Screenshot 2025-10-16 at 7 22 24 PM

**What type of PR is this?**

```release-note
 Add strict validation for cluster encryption keys requiring proper 32-byte base64 format
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/shared/validators/encryption-key.validator.ts
+++ b/modules/web/src/app/shared/validators/encryption-key.validator.ts
@@ -20,10 +20,7 @@ export class EncryptionKeyValidator implements Validator {
     if (!value) {
       return null;
     }
-
-    // 32-byte encryption key must encode to exactly 44 base64 characters
-    // Format: 43 base64 characters + exactly 1 padding character ('=')
-    // This strict pattern prevents malformed keys that could decode to wrong byte length
+    // 32-byte key must be exactly 44 base64 characters (43 chars + 1 padding)
     const base64Regex = /^[A-Za-z0-9+/]{43}=$/;
     if (!base64Regex.test(value)) {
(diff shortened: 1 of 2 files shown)


modules/web/src/app/shared/validators/encryption-key.validator.ts
```diff
@@ -0,0 +1,39 @@
...
+import {AbstractControl, ValidationErrors, Validator} from '@angular/forms';
+
+export class EncryptionKeyValidator implements Validator {
+  validate(control: AbstractControl): ValidationErrors | null {
+    const value = control.value;
+    if (!value) {
+      return null;
+    }
+    // 32-byte key must be exactly 44 base64 characters (43 chars + 1 padding)
+    const base64Regex = /^[A-Za-z0-9+/]{43}=$/;
```
KhizerRehan: Hi, @adoi 

Regex Adds Strict Validation to check
  1. 32-byte keys must encode to exactly 44 base64 characters
  2. The format is specifically 43 base64 characters + 1 padding character

so, it string "Rand" generated or User provided should always contain "Padding" character to pass validation check

Let me know if any concerns

PR #7673, opened 2025-10-30
Fix display details when preset selected in cluster summary

KhizerRehan:
This PR fixes an issue where cluster configuration details were not being displayed when a preset was selected in the cluster summary view

Before:
e.g Dev (Provider: Openstack)
Screenshot 2025-10-30 at 12 50 53 PM

After:
Screenshot 2025-10-30 at 12 50 00 PM

**What type of PR is this?**
/kind bug

```release-note
Fix cluster summary not displaying provider details when using presets
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/shared/components/cluster-summary/template.html
+++ b/modules/web/src/app/shared/components/cluster-summary/template.html
@@ -329,7 +329,7 @@ limitations under the License.
           <div value>{{cluster.credential}}</div>
         </km-property>
 
-        <ng-container *ngIf="!cluster.credential">
+        <ng-container *ngIf="displaySettings()">
           <!-- AWS -->
           <ng-container *ngIf="cluster.spec.cloud?.aws">
             <km-property *ngIf="cluster.spec.cloud?.aws?.securityGroupID">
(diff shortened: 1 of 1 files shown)


modules/web/src/app/shared/components/cluster-summary/template.html
```diff
@@ -329,7 +329,7 @@ <h4>IPv6</h4>
           <div value>{{cluster.credential}}</div>
         </km-property>
 
-        <ng-container *ngIf="!cluster.credential">
+        <ng-container *ngIf="displaySettings()">
```
ahmadhamzh: Wouldn’t it be easier to just check if there’s no preset (!cluster.credential), since all the properties inside the ng-container are disabled when a preset is selected?

modules/web/src/app/shared/components/cluster-summary/template.html
```diff
@@ -385,7 +385,7 @@ <h4>IPv6</h4>
               <div value>{{cluster.spec.cloud?.azure.vnet}}</div>
             </km-property>
 
-            <km-property-boolean *ngIf="cluster.spec.cloud?.azure?.assignAvailabilitySet"
+            <km-property-boolean *ngIf="cluster.spec.cloud?.azure?.assignAvailabilitySet !== undefined"
```
ahmadhamzh: i think usually if the value is false we don't show it on the summary step so i think no need for this change

PR #7729, opened 2025-12-11
Change the quota calculations to use the binary system instead of the decimal system

### LINKED ISSUE #7715: project quota calculations should be done with Storage units (MiB, GiB) instead of decimal units MB, GB etc
### What happened
Currently quota calculations for memory are assuming the units MB, GB as decimal units. Due to this the % quota remaining calculations are getting calculated incorrectly.

e.g. if the memory for the VM is 12288MB - it essentially is 12288 MiB = 12GiB. 
Similarly, if the memory quota for project is mentioned as 24GB - it essentially means it is 24GiB.

So if we get a VM with 12288 MiB Memory, we have consumed exactly 50% of the quota (12 GiB / 24 GiB). Currently, this calculation is done assumoing 12288 MB = 11.44 GiB. So quota consumed comes at ~ 47.66%.

### Expected behavior
Quota calculations are done correctly with all units being Storage units expressed as power of 2. e.g. 2^20 = 1MiB.

### How to reproduce
Add quota to project and create a VM as described above.

### Environment
- UI Version: Any
- API Version: Any

### Current workaround
None

### Affected user persona
KKP Admin

### Additional information
internal-reference: 8515

ahmadhamzh:
This PR updates the resource quota calculation logic to use binary units (IEC) instead of decimal units (SI).

Fixes #7715

**What type of PR is this?**
/kind bug

there is some issues in getting the global usage for Anexia, Kubevirt and AWS providers,
a ticket has been created [here](https://github.com/kubermatic/kubermatic/issues/15272)

```release-note
Updated resource quota calculations to use binary units (base-1024).
```

**Documentation**:
```documentation
NONE
```

### DIFF
--- a/modules/api/pkg/api/v2/convert.go
+++ b/modules/api/pkg/api/v2/convert.go
@@ -27,7 +27,7 @@ import (
 
 func ConvertToAPIQuota(resourceDetails kubermaticv1.ResourceDetails) Quota {
 	quota := Quota{}
-	gip := 1024 * 1024 * 1024
+	bytesPerGiB := 1024 * 1024 * 1024
 
 	if resourceDetails.CPU != nil {
 		cpu := resourceDetails.CPU.Value()
@@ -37,7 +37,7 @@ func ConvertToAPIQuota(resourceDetails kubermaticv1.ResourceDetails) Quota {
 	// Get memory and storage denoted in GB
 	if resourceDetails.Memory != nil && !resourceDetails.Memory.IsZero() {
 		changeToBinary := TreatDecimalAsBinary(resourceDetails.Memory)
-		memory := float64(changeToBinary.Value()) / float64(gip)
+		memory := float64(changeToBinary.Value()) / float64(bytesPerGiB)
 		// round to 2 decimal places
 		memory = math.Round(memory*100) / 100
 		quota.Memory = &memory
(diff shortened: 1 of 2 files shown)


modules/api/pkg/ee/resource-quota/handler.go
```diff
@@ -533,10 +537,10 @@ func getGCPResourceDetails(provider ProviderNodeTemplate, nc *kubermaticprovider
 func getHetznerResourceDetails(provider ProviderNodeTemplate, nc *kubermaticprovider.NodeCapacity) error {
 	nc.WithCPUCount(provider.HetznerSize.Cores)
 
-	if err := nc.WithMemory(int(provider.HetznerSize.Memory), "G"); err != nil {
+	if err := nc.WithMemory(int(provider.HetznerSize.Memory), "Gi"); err != nil {
```
KhizerRehan: Clarification why do need "Gibibyte" for memoy instead of "Mi" like we are doing for other providers?

modules/api/pkg/api/v2/convert.go
KhizerRehan: Can you fix "Message" 
```suggestion
			return kubermaticv1.ResourceDetails{}, fmt.Errorf("error parsing quota Storage %w", err)
```

KhizerRehan (commented):
Feedback/Clarification (PTAL)

PR #7733, opened 2025-12-16
WIP: Handle Error notification throttling with exponential backoff

### LINKED ISSUE #7643: UI/UX Constant Error Popups Caused by Polling on Failed API Requests

When viewing a cluster, users frequently encounter repeated error pop-ups every few seconds due to continuous frontend polling. This results in a poor user experience, as the error dialog keeps reappearing whenever an API endpoint fails or a service is temporarily unavailable, as described below.

Reported Example:

Image

Image

### Example Endpoint:
- nodes/metrics service continuously returns 503 Service Unavailable

###  Reproduction Steps (Example)

1. Create a cluster. (e.g OpenStack)
2. Navigate to Cluster Details → Machine Deployment Details (Nodes).
3. Observe that the `nodes/metrics` endpoint fails periodically on each refresh interval.

Image

### Impact
-  Results in **poor UX** due to repeated error popups triggered by polling.

### **Suggest how to fix it, if possible**

We should explore ways to improve error handling and suppression in such cases  for example:

- Improving the UX to handle repeated API failures more gracefully
- Considering backoff or retry strategies for repeated polling failures.
- Consider switching from polling to websocket instead of polling.

KhizerRehan:
This PR fixes repeated error notification spam by adding smart throttling, reducing noise while still showing important errors.

Fixes #7643 

Env Flag
k get kubermaticsettings -o yaml | grep enableThrottling

Local Testing: `web/src/assets/config/config.json`

```ts
   "initial_delay_minutes": 2,
    "max_delay_minutes": 10,
    "backoff_multiplier": 2,
    "entry_expiration_hours": 5
```

It was decided to have configurable values defeined in configMap under ui.config

k get kubermaticconfigurations.kubermatic.k8c.io -n kubermatic -oyaml

Prod Testing: 

```ts
   "initial_delay_minutes": <set_for_prod>, 60
    "max_delay_minutes":  <set_for_prod> e.g 120 
    "backoff_multiplier":  <set_for_prod> e.g 2
    "entry_expiration_hours":  <set_for_prod> e.g 8
```

**What type of PR is this?**
/kind bug

```release-note
 Add error notification throttling with exponential backoff and auto-mute to reduce repetitive error alerts
```

**Documentation**:

```documentation
NONE
```


modules/web/src/app/core/interceptors/error-notifications.ts
```diff
@@ -89,31 +101,41 @@ export class ErrorNotificationsInterceptor implements HttpInterceptor {
...
         error: (httpError: HttpErrorResponse) => {
-          if (this._shouldSilenceRequest(req)) {
-            return;
-          }
-
-          if (!httpError) {
-            return;
-          }
-
-          let error = this._toError(httpError);
```
KhizerRehan: FYI: renamed to `_convertToError`

KhizerRehan (commented):
Comments for Reviewer

ahmadhamzh:
Some concerns regarding UX:

* Currently, some endpoints return error notifications when credentials are invalid (e.g., /api/v2/projects/{project_id}/providers/digitalocean/sizes, /api/v2/projects/{project_id}/providers/gcp/networks, /api/v2/projects/{project_id}/providers/gcp/{dc}/zones).

   With this new implementation, if the user enters incorrect credentials, they will see the error once. However, if they change the credentials again and they are still invalid, no error is shown. This could allow the user to continue to the next step (without seeing any zones, networks, or sizes) and without any indication that the credentials are still invalid, potentially causing confusion.
   
 * In some cases, when adding or editing an object, certain inputs may trigger an error from the provider. The first time this happens, the error is displayed correctly. However, if the user tries again with another invalid input, the error is no longer shown, which could be confusing.

    For example, when adding a machine (MD) with an unsupported size, a notification appears after closing the dialog. But if the user tries again with a different unsupported size, no error is shown, even though the action still fails.

I’m not sure how often this occurs or in which other places it might happen, but I thought it was worth mentioning.

PR #7794, opened 2026-01-27
Project list: backend search param supports cluster name and ID search

### LINKED ISSUE #7514: Search on projects overview page should also search for clusters

In 1st level support, customers often report problems with UCs giving only the cluster name or, if we're lucky, cluster ID. But they rarely mention the project name, let alone project ID. Thus it would be nice if the search function on `/projects` would also search for cluster names and IDs and show me the projects with a match.

I guess this could go even further, depening on what data is available without adding too many queries. Things like searching by user's name (and/or email) because sometimes they don't mention which cluster at all - either just owners or better all users. Or by machine deployment name, because sometimes we only get a worker node's name.

csengerszabo:
  Switch project search to use a backend search parameter that matches project fields and cluster name/ID, avoiding per-project cluster lookups and enabling cluster-name search at scale.

  Fixes #7514

  **What type of PR is this?**
  /kind feature
  /kind api-change

  - API: adds `search` query param to `/projects` (supports matching by project name/id/owners/labels and cluster name/id).
  - UI: non-empty search now calls `searchProjects`; empty search still loads full list.
  - Tests updated in API and UI.

  ```release-note
  Projects search now supports backend search by project fields and cluster name/ID, improving performance versus per-project cluster lookups.
  ```

  **Documentation**:
  
  ```documentation
NONE
  ```

### DIFF
--- a/modules/api/pkg/handler/v1/project/project.go
+++ b/modules/api/pkg/handler/v1/project/project.go
@@ -270,11 +270,11 @@ func ListEndpoint(userInfoGetter provider.UserInfoGetter, projectProvider provid
 		}
 
 		if req.DisplayAll && (userInfo.IsAdmin || userInfo.IsGlobalViewer) {
-			allProjects, err := getAllProjectsForAdmin(ctx, userInfo, projectProvider, memberProvider, userProvider, clusterProviderGetter, seedsGetter)
+			allProjects, clusterMatchedProjectIDs, err := getAllProjectsForAdmin(ctx, userInfo, projectProvider, memberProvider, userProvider, clusterProviderGetter, seedsGetter, req.Search)
 			if err != nil {
 				return nil, err
 			}
-			return filterProjectsBySearch(ctx, allProjects, req.Search, clusterProviderGetter, seedsGetter)
+			return filterProjectsBySearch(allProjects, req.Search, clusterMatchedProjectIDs)
 		}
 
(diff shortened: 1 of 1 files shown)


Waseem826:
> > We're already loading clusters in the API for calculating cluster count per project and adding cluster name/ID will only increase the response payload size. Regarding stale data, I think it should be fine since we're refreshing projects list every 10 seconds.
> 
> I think sending a GET request on each keystroke isn’t right. At the same time, if we expect `multiple projects have multiple 1000s of clusters`, that would also make the response object very large.
> 
> I’d suggest adding a debounce time between keystrokes and the request (maybe 500 ms), and also enforcing a minimum number of characters for searching. For example, if the input is fewer than 3 characters, the search wouldn’t run. This would reduce the number of matches, which could otherwise be very large since we expect 1000s of clusters. @Waseem826 @csengerszabo

I agree with adding debounce but may be we shouldn't add search text limit since it will result in bad UX (we don't add this restriction anywhere else). If its common to have 1000 clusters per project then keeping the search using API makes sense.

modules/web/package-lock.json
Waseem826: These changes are unrelated to this PR so we should revert them and put them in separate PR if necessary.

ahmadhamzh (commented):
some concerns on the API changes

PR #7800, opened 2026-01-29
Fix url building logic causing issues with other ingress controller

4ch3los:

While migrating our Infrastructure to a new Ingress Controller(in our case traefik) we realized that there is an issue with the download Kubeconfig function. 
The current implementation construct the URL to look like https://kubermatic.example.com//api/v1/kubeconfig?project_id<...> with a double / infront of the path. The nginx controller seems to have merge_slashes active by default which results in a 301 redirect to https://kubermatic.example.com/api/v1/kubeconfig?project_id<...> with one slash and a working process. Other Ingress controllers dont do this, resulting in the browser not setting the csfr_token cookie and after login, failing to download the kubeconfig. 
Thats why we should fix the url generation. This pr does exactly this, by a) unifying the variables across all environments(checked the usages and the option with a leading / should be better) and b) making the crafted url contains one and only one slash.

**What type of PR is this?**

/kind bug

Regarding the near archival of the nginx ingress this bug is rather important.

```release-note
- Fixed Kubeconfig download with non nginx ingresses
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/core/services/cluster.ts
+++ b/modules/web/src/app/core/services/cluster.ts
@@ -255,7 +255,8 @@ export class ClusterService {
 
   getShareKubeconfigURL(projectID: string, clusterID: string, userID: string, oidcKubeLogin = false): string {
     const location = this._location.endsWith('/') ? this._location : `${this._location}/`;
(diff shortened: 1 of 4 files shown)


ahmadhamzh (approved):
/approve

Issue #7706, opened 2025-11-20
UI/UX enhancement


ahmadhamzh:
### What happened
Please add any minor UI/UX bugs with a proper description.

ahmadhamzh:
When trying to filter presets by provider and selecting a provider that has no presets, the spinner is displayed and never disappears.

<img width="1522" height="496" alt="Image" src="https://github.com/user-attachments/assets/1fd5cd2d-5a26-445f-8110-acb5c536eba7" />

ahmadhamzh:
If I'm in the wizard to create, edit, or customize a cluster template and then I cancel, it currently navigates to the clusters list page. I think it should instead redirect to the cluster template list page.

https://github.com/user-attachments/assets/1705b9d3-00ff-49a9-9744-5791e0c3266d

Waseem826:
Card background and colors in the dialog (header, footer) needs to be updated after upgrading to Angular version 20.

<img width="1684" height="577" alt="Image" src="https://github.com/user-attachments/assets/2c14e663-db99-4338-a980-c7d91ee54b93" />

<img width="821" height="372" alt="Image" src="https://github.com/user-attachments/assets/642d942b-fa90-4c72-832b-36f610eafb49" />

PR #7867, opened 2026-02-19
Fix: Loading spinner remains indefinitely when filtering presets returns zero results

KhizerRehan:
This PR fixes this UI/UX [issue](https://github.com/kubermatic/dashboard/issues/7706#issuecomment-3555944748) 

Fixes: #7706
Ref: https://github.com/kubermatic/dashboard/issues/7706#issuecomment-3555944748

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/presets/component.ts
+++ b/modules/web/src/app/settings/admin/presets/component.ts
@@ -124,7 +124,7 @@ export class PresetListComponent implements OnInit, OnDestroy, OnChanges {
             presetStats$.push(this._presetService.getPresetStatsBy(preset.name));
           });
           this.isBusyCounter++;
-          return forkJoin(presetStats$);
+          return presetStats$.length > 0 ? forkJoin(presetStats$) : of([]);
         }),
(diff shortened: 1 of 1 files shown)


modules/web/src/app/settings/admin/presets/component.ts
```diff
@@ -124,7 +124,7 @@ export class PresetListComponent implements OnInit, OnDestroy, OnChanges {
             presetStats$.push(this._presetService.getPresetStatsBy(preset.name));
           });
           this.isBusyCounter++;
-          return forkJoin(presetStats$);
+          return presetStats$.length > 0 ? forkJoin(presetStats$) : of([]);
```
KhizerRehan: Reason:

forkJoin([]) -> never emits value therefore (next) handler is never executed it goes to complete handler therefore busyCounter was never `decremented)

where as of([]) -> emits 1 value therefore in case empty obs$ array we can use `of([])` to emit atleast `[]` so that next handler can be executed.

PR #7913, opened 2026-03-05
Fix enforce event rate in edit cluster dialog

ahmadhamzh:
Disable the Event Rate Limit plugin option from the plugins dropdown list when it's enforced by admin, to prevent the user from unchecking it.

**Before:**
image

**After:**
image

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```


ahmadhamzh:
@Waseem826 we need code freeze approve

PR #7921, opened 2026-03-10
Handle country code lookup errors gracefully in dynamic datacenters

### LINKED ISSUE #7904: Dynamic Datacenters fails on Non-ISO Country String
### What happened

QA Env: v2.30.0-rc.1

The API (`GET /api/v1/dc`) `country: "conformance"` an 11-character string, not a valid ISO 3166-1 alpha-2 (2-char) or alpha-3 (3-char) code. The `country-code-lookup` library's `byIso()` function **throws an exception** (rather than returning null/undefined) when given input that is neither 2 nor 3 characters long. `getCountryName()` only guards against falsy values, not thrown exceptions, so the error propagates uncaught.

### How to reproduce
1. Log in as an Administrator.
2. Navigate to **Admin Settings** → **Dynamic Datacenters**.
3. Observe the page crashes immediately with `ERROR Error: cannot determine ISO code type`.

### Relevant logs / screenshots

Image

API response excerpt (one of 31 affected entries):
```json
{
  "metadata": { "name": "..." },
  "spec": {
    "country": "conformance",
    ...
  }
}
```

### Current workaround
None. The page becomes completely inaccessible to administrators when any datacenter has a non-ISO country string. If this is caused by the API response, it needs to be fixed to return a valid country value in the datacenter field that complies with ISO 3166-1 alpha-2 (2-character) or alpha-3 (3-character) country codes.

## Affected Places
- Create Cluster Wizard  -> Provider (Kubevirt)
- Admin Setting -> Dynamic Datacenters page

### Environment
- UI Version: v2.30.0-rc.1
- Browser & Version: N/A — needs investigation
- Others: Occurs when the seed returns datacenters with `country: "conformance"` (invalid) ISO string

KhizerRehan:
This PR fixes a crash in the dynamic datacenters admin settings when countryCodeLookup.byIso() throws an error for  invalid or unrecognized ISO country codes, by wrapping the call in a try-catch and falling back to the raw code.

**BEFORE:**
image

**AFTER**

image

Screenshot 2026-03-10 at 2 14 10 AM

Fixes #7904 

**What type of PR is this?**
/kind bug

```release-note
Fix crash in dynamic datacenters admin page when encountering invalid ISO country codes.
```

**Documentation**:

```documentation
NONE
```

### DIFF
--- a/modules/web/src/app/settings/admin/dynamic-datacenters/component.ts
+++ b/modules/web/src/app/settings/admin/dynamic-datacenters/component.ts
@@ -110,8 +110,12 @@ export class DynamicDatacentersComponent implements OnInit, OnDestroy, OnChanges
       return '';
     }
 
-    const country = countryCodeLookup.byIso(code);
-    return country ? country.country : code;
+    try {
+      const country = countryCodeLookup.byIso(code);
+      return country ? country.country : code;
+    } catch {
+      return code;
+    }
   }
 
   private _setCountries(datacenters: Datacenter[]) {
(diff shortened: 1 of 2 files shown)


Waseem826 (approved):
/approve

Issue #1105, opened 2019-03-05
Dashboard secure login


p0lyn0mial:
**Note:** Issues relating to this epic are managed [via ZenHub](https://app.zenhub.com/workspace/o/kubermatic/dashboard/issues/1105).

At the moment the dashboard uses OAuth Implicit Grant flow for login. This is bad  for at least two reasons. Firstly Implicit flow is not considered secure, mainly because the ID token is returned in URL. Secondly its lifetime is usually short, it’s one day in our case and requires users to constantly log in into the app.

To solve the mentioned issues we could use Authorization Code, below is simplified diagram depicting interactions between the actors.

User (Browser) <—— ( 1, 4 ) ——> Server < ———— ( 2, 3) ———> OIDC Provider

In step 1 a user sends a request to a server without exposing any credentials like Client ID (Implicit Grant), next the server and the provider exchange a series of requests. At the end the server gets both the ID and Refresh tokens. In Step 4 the tokens are send back to the user. 

Tokens are considered confidential and should be stored carefully, therefore:
- The ID token should be valid only for a short period of time, for example 10 minutes.
- The Refresh token could be valid for 30 days or so.
- The ID token doesn’t have to be persisted in a cookie or web storage. It could be simply saved in memory. The token could be saved in a global variable accessible to our code. On startup, the app could use the Refresh token to get a new ID token.
- The Refresh token could be stored in a secure cookie read only on the server
set httpOnly, secure and SameSite=strict flags for the cookie
- The Refresh token stored in a cookie should be send only to some endpoints not all (set cookie path) If the path is not set then the cookie is send for the domain, for example each request to dev.kubermatic.io includes the cookie. This means that the cookie could be read by the API server, the controllers and the dashboard. This is not necessary because the cookie could be read only by two endpoints. One for refreshing the ID token and one for handling "download kubeconfig" endpoint.
- The Refresh token is used to get a new ID token

cschieder:
@kubermatic/ui-team this issue is over a year old, is it still relevant?

floreks:
Quite relevant. This requires some backend changes though.

ahmedwaleedmalik:
This is still relevant and important for our customers.

archups:
Tracked as part of https://support.kubermatic.com/a/tickets/6068?current_tab=details support ticket.

KhizerRehan:
Fixed: https://github.com/kubermatic/dashboard/pull/7960
/close

PR #7960, opened 2026-03-30
Migrate KKP dashboard auth from implicit flow to OAuth authorization code flow

### LINKED ISSUE #7940: OAuth Authorization Code Flow Migration
# OAuth Authorization Code Flow Migration

## Overview

The current authentication mechanism relies on the OAuth **Implicit Flow**, where tokens are issued directly to the browser and handled by the frontend. This approach introduces several security concerns, including token exposure in URL fragments and access via JavaScript.

To address these limitations, this proposal recommends migrating to the **Authorization Code Flow**, with token exchange handled by the backend.

---

## Motivation

The Implicit Flow is no longer considered best practice due to:

- Token leakage via URL fragments
- Exposure to XSS attacks (tokens accessible in JS)
- Lack of refresh token support
- Short-lived sessions requiring frequent re-authentication

---

## Proposed Architecture

- Use **Authorization Code Flow with OIDC**
- Perform code exchange on the backend
- Store tokens in **HttpOnly, Secure cookies**
- Introduce **refresh tokens** for long-lived sessions
- Implement **token rotation** for improved security

---

## Benefits

| Area              | Improvement |
|------------------|------------|
| Security         | Tokens are never exposed to the browser |
| Session UX       | Long-lived sessions (up to 30 days) |
| Token Handling   | Backend-managed lifecycle |
| Standards        | Aligns with modern OAuth best practices |

---

### Login Flow
This diagram illustrates the authentication process from the moment the user initiates login until the backend sets the authentication cookies.

Key points:
- Authorization code is exchanged on the backend
- Tokens are never exposed to the browser
- Session is established via HttpOnly cookies

Image

### Refresh Flow

This diagram illustrates how the application maintains an active session using refresh tokens.

Key points:
- The ID token is short-lived (1 hour)
- The client performs a silent refresh before expiration (55 minutes)
- The backend uses the refresh token to issue a new ID token
- The refresh token is rotated on each use for improved security

Image

### Logout Flow

This diagram shows how a user session is securely terminated.

Key points:
- The client sends a logout request to the backend
- The backend clears authentication cookies
- Any stored session or refresh token is invalidated

Image

---

## API Design

| Endpoint               | Method | Description |
|-----------------------|--------|-------------|
| /api/v2/auth/login    | GET    | Initiates authentication flow |
| /api/v2/auth/callback | GET    | Handles OIDC callback |
| /api/v2/auth/refresh  | POST   | Refreshes tokens |
| /api/v2/auth/logout   | POST   | Terminates session |

---

## Token Strategy

- **ID Token TTL:** 1 hour  (configurable via the OIDC provider)
- **Refresh Token TTL:** 30 days   (configurable via the OIDC provider)
- **Rotation:** Enabled (recommended)  
- **Refresh Mechanism:** Silent background refresh  

This ensures a balance between security and user experience, minimizing re-authentication while limiting token exposure.

---

## Considerations

- CSRF protection for cookie-based authentication
- Secure cookie configuration (`HttpOnly`, `Secure`, `SameSite`)
- Handling refresh failures (fallback to login or prompt)

---

## Conclusion

Migrating to the Authorization Code Flow significantly improves the security posture of the application while providing a smoother user experience through long-lived sessions and controlled token management.

---

## Feedback

Please share any concerns, suggestions, or edge cases that should be addressed before implementation.

ahmadhamzh:

Migrate KKP dashboard authentication from OAuth implicit flow to authorization code flow with PKCE. Tokens are now exchanged server-side and stored in HttpOnly cookies.

Fixes #7940, https://github.com/kubermatic/dashboard/issues/1105

**What type of PR is this?**
/kind design

```release-note
Migrate KKP dashboard authentication from OAuth implicit flow to authorization code flow with PKCE.
```

**Documentation**:
```documentation
TBD
```

**Test issue**:
```test-issue
https://github.com/kubermatic/dashboard/issues/7982
```


modules/api/hack/run-api.sh
```diff
@@ -69,7 +69,7 @@ set -x
   -prometheus-url=http://localhost:9090 \
   -address=127.0.0.1:8080 \
   -oidc-url=https://dev.kubermatic.io/dex \
-  -oidc-authenticator-client-id=kubermatic \
+  -oidc-authenticator-client-id=kubermaticIssuer \
```
ahmadhamzh: Changed `-oidc-authenticator-client-id` from `kubermatic` to `kubermaticIssuer`                                                                                                       
  because the new auth code flow issues tokens using the issuer client (kubermaticIssuer).                                                                                              
  The token verifier must use the same client ID to accept these tokens (audience claim must match).                                                                                    
  The old `kubermatic` public client was only needed for the implicit flow which is being removed.

KhizerRehan: Clarification:
- does client-id dependent on type of code flow? or even in implicit flow it was using `kubermaticissuer` this is changed w.r.t to running api locally right?

modules/api/pkg/handler/v2/kubermatic-dashboard/routing.go
```diff
@@ -0,0 +1,71 @@
+/*
```
KhizerRehan, line 1: IMO, this package should be renamed for me I thought -> some other dashboard

- It is something related to loginflow but IMHO, this package should be renamed. wdyt?

modules/api/pkg/handler/v2/kubermatic-dashboard/handler.go
```diff
@@ -0,0 +1,487 @@
...
+
+const (
+	nonceCookieName   = "nonce"
+	nonceCookieMaxAge = 180
+	callbackPath      = "/api/v2/auth/callback"
+)
+
+const pkceVerifierLen = 64
+
+// generateCodeVerifier creates a cryptographically random PKCE code verifier (RFC 7636).
```
KhizerRehan, line 44: I see native impl but while checking codebase we are using  `"golang.org/x/oauth2"` package why didn't can't we use existing methods

- https://pkg.go.dev/golang.org/x/oauth2@v0.34.0#GenerateVerifier

As far I can see in Lib

> GenerateVerifier generates a PKCE code verifier with 32 octets of randomness. This follows recommendations in [RFC 7636](https://rfc-editor.org/rfc/rfc7636.html).

modules/api/pkg/handler/v2/kubermatic-dashboard/handler.go
```diff
@@ -0,0 +1,487 @@
...
+func generateCodeVerifier() (string, error) {
+	buf := make([]byte, pkceVerifierLen)
+	if _, err := rand.Read(buf); err != nil {
+		return "", fmt.Errorf("failed to generate code verifier: %w", err)
+	}
+	return base64.RawURLEncoding.EncodeToString(buf), nil
+}
+
+// deriveCodeChallenge computes the S256 code challenge from a code verifier.
+func deriveCodeChallenge(verifier string) string {
```
KhizerRehan, line 54: can we use
- https://pkg.go.dev/golang.org/x/oauth2@v0.34.0#S256ChallengeFromVerifier

modules/api/pkg/handler/v2/kubermatic-dashboard/handler.go
```diff
@@ -0,0 +1,487 @@
...
+	hash := sha256.Sum256([]byte(verifier))
+	return base64.RawURLEncoding.EncodeToString(hash[:])
+}
+
+// appendPKCEParams adds code_challenge and code_challenge_method to an auth URL.
+func appendPKCEParams(authURL, codeChallenge string) string {
```
KhizerRehan, line 60: can you check
https://cs.opensource.google/go/x/oauth2/+/refs/tags/v0.34.0:pkce.go;l=57 if we can use lib method instead of manual string concat?

modules/web/src/environments/environment.ts
```diff
@@ -31,7 +31,5 @@ export const environment = {
   newRestRoot: '/api/v2',
   wsRoot: `${wsProtocol}//${host}/api/v1/ws`,
   avoidWebsockets: false,
-  oidcProviderUrl: 'https://dev.kubermatic.io/dex/auth',
-  oidcConnectorId: null,
```
KhizerRehan, line 35: So, it will be build in backend or via  argument?

`-oidc-url`

modules/web/src/app/core/services/auth/service.ts
```diff
@@ -12,214 +12,105 @@
...
-
-  login(): void {
-    this._cookieService.set(this._cookie.autoredirect, 'true', 1, '/', null, false, 'Strict');
-  }
-
-  logout(): Observable<boolean> {
-    return this._userService
-      .logout()
+    this._refreshSub = timer(msUntilRefresh)
       .pipe(
```
KhizerRehan, line 95: Want to understand we are not handing as interval does it trigger auto since we save expiration time in memory via checkStatus 

does it trigger auto?  like how it will be called like when dex reaches to expiration timeout?

KhizerRehan (commented):
Feedback/Clairfication

modules/api/pkg/handler/v2/kubermatic-dashboard/routing.go
```diff
@@ -0,0 +1,71 @@
...
+	router.Methods(http.MethodGet).
+		Path("/auth/login").
+		Handler(a.loginHandler())
+
+	router.Methods(http.MethodGet).
+		Path("/auth/callback").
+		Handler(a.callbackHandler())
+
+	router.Methods(http.MethodPost).
+		Path("/auth/refresh").
```
KhizerRehan, line 47: How I tested but something I want to ask that

- Run API/UI Locally
- const REFRESH_BUFFER_SECONDS = 5; // instead of 60s
- Changed Dex Expiration time e.g 1m 
-> Hard Reload -> Cleared Existing Login (if Already) -> Login Again (to Get updated setting) of dex / expiration which was verified.


FYI: I did checked Cookies -> Expires Time was changed -> when 	`refresh` api call was triggered (like moving expires time `ahead` e.g 1m more) -> and if 3 calls were done so  3 times cookie was updated (just letting you know that cookie) was updated.

Image



However  As soon I interacted and moved e.g Projects Overview by Selection any project. I see logout screen and IDK it like it ask to Login agian shouldn't "Refresh" flow -> shouldn't Logout to me OR refersh of Page should keep me logged In?  

Or Am I missing something in testing?


Image


Image

modules/api/pkg/handler/v2/authflow/routing.go
```diff
@@ -0,0 +1,71 @@
+/*
```
ahmadhamzh: i will change it to `authflow` WDYT ?

modules/api/pkg/handler/v2/authflow/handler.go
```diff
@@ -0,0 +1,487 @@
...
+}
+
+// deriveCodeChallenge computes the S256 code challenge from a code verifier.
+func deriveCodeChallenge(verifier string) string {
+	hash := sha256.Sum256([]byte(verifier))
+	return base64.RawURLEncoding.EncodeToString(hash[:])
+}
+
+// appendPKCEParams adds code_challenge and code_challenge_method to an auth URL.
+func appendPKCEParams(authURL, codeChallenge string) string {
```
ahmadhamzh: that would require some changes to [AuthCodeURL](https://github.com/kubermatic/dashboard/blob/9c5ba4a0cf27531022dfa824fe126a13528d5b09/modules/api/pkg/provider/auth/types/types.go#L50-L51) and since it’s used in multiple places I’d prefer not to modify it just to avoid manual string concatenation.

modules/web/src/environments/environment.ts
ahmadhamzh: in backend

modules/web/src/app/core/services/auth/service.ts
ahmadhamzh: Yes, it’s supposed to be triggered automatically 1 minute before the token is about to expire.

KhizerRehan:
@ahmadhamzh can we close this ticket https://github.com/kubermatic/dashboard/issues/1105 also as part of this ticket? 

Can we add this in "Fixes" in PR description?

PR #7972, opened 2026-04-03
BackupStorageLocation UX improvements

SamuAlfageme:

This introduces a few UX improvements on the handling of `BackupStorageLocations` (BSLs) for the UC Backup feature. These apply to both, the Cluster Creation Wizard and "Edit cluster" dialogue.

 - New  "Create Backup Storage Location' button on clicking "Cluster Backup" checkbox

**Image Update**

KR: FYI
- This option is availaible for Create Cluster Wizard not for (Edit Cluster) to open a new dialog over a dialog

Screenshot 2026-07-29 at 3 41 09 PM

- Add field validations on the BSL form(s):
  - Basic regexp validation for DNS on 'Region' and 'Endpoint URL' (+ `http(s)` for the latter).
  
Screenshot 2026-07-29 at 3 42 04 PM

- Mandatory 'Endpoint' & 'Region' unless using 'custom config' (see behavior in gif below).

Screenshot 2026-07-29 at 3 43 50 PM

- Allow to copy BSL error messages on click from the BackupStorageLocation list view

<div align="center">
   
</div>

- Fix: reorder BSL selector & Skip router reconciliation on the 'Edit cluster' modal: the checkbox to enable UC Backup and the BSL dropdown were split by the new 'Skip router reconciliation' on OpenStack DCs

image

**Which issue(s) this PR fixes**: N/A 

**What type of PR is this?**
/kind design

```release-note
Cluster Creation/Edition: Improve BackupStorageLocation UX: display only available BSLs and allow to create new ones.
```

**Documentation**:

```documentation
TBD
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -568,10 +568,15 @@ export class EditClusterComponent implements OnInit, OnDestroy {
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(cbslList => {
         this.backupStorageLocationsList = cbslList.filter(bsl => this._isBackupStorageLocationAvailable(bsl));
-        this.backupStorageLocationLabel = this.backupStorageLocationsList.length ? BSLListState.Ready : BSLListState.Empty;
+        this.backupStorageLocationLabel = this.backupStorageLocationsList.length
+          ? BSLListState.Ready
+          : BSLListState.Empty;
 
         const backupStorageLocationControl = this.form.get(Controls.BackupStorageLocation);
-        if (backupStorageLocationControl && !this.backupStorageLocationsList.some(bsl => bsl.name === backupStorageLocationControl.value)) {
+        if (
+          backupStorageLocationControl &&
+          !this.backupStorageLocationsList.some(bsl => bsl.name === backupStorageLocationControl.value)
+        ) {
           backupStorageLocationControl.reset();
         }
       });
--- a/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -14,9 +14,11 @@
 
 import {ChangeDetectorRef, Component, Input, OnDestroy, OnInit} from '@angular/core';
 import {FormBuilder, FormControl, FormGroup, Validators} from '@angular/forms';
-import {MatDialogRef} from '@angular/material/dialog';
+import {MatDialog, MatDialogRef} from '@angular/material/dialog';
+import {MatSelectChange} from '@angular/material/select';
 import {ClusterBackupService} from '@app/core/services/cluster-backup';
 import {UserClusterConfigService} from '@app/core/services/user-cluster-config';
+import {AddBackupStorageLocationDialogComponent} from '@app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component';
 import {DynamicModule} from '@app/dynamic/module-registry';
 import {BackupStorageLocation} from '@app/shared/entity/backup';
 import {NODEPORTS_IPRANGES_SUPPORTED_PROVIDERS, NodeProvider} from '@app/shared/model/NodeProviderConstants';
@@ -66,7 +68,7 @@ import {HTTP_PROXY_URL_PATTERN_VALIDATOR, IPV4_IPV6_CIDR_PATTERN, NO_PROXY_PATTE
 import {KmValidators} from '@shared/validators/validators';
 import _ from 'lodash';
 import {Observable, Subject} from 'rxjs';
-import {map, switchMap, take, takeUntil, tap} from 'rxjs/operators';
+import {filter, map, switchMap, take, takeUntil, tap} from 'rxjs/operators';
 
 enum Controls {
   Name = 'name',
@@ -138,6 +140,7 @@ export class EditClusterComponent implements OnInit, OnDestroy {
   isAuditWebhookBackendHidden = false;
   backupStorageLocationsList: BackupStorageLocation[];
   backupStorageLocationLabel: BSLListState = BSLListState.Ready;
+  readonly createBackupStorageLocationOptionValue = '__create_backup_storage_location__';
   isAllowedIPRangeSupported: boolean;
   readonly isEnterpriseEdition = DynamicModule.isEnterpriseEdition;
   readonly CLUSTER_DEFAULT_NODE_SELECTOR_NAMESPACE = CLUSTER_DEFAULT_NODE_SELECTOR_NAMESPACE;
@@ -174,6 +177,7 @@ export class EditClusterComponent implements OnInit, OnDestroy {
     private readonly _builder: FormBuilder,
     private readonly _clusterService: ClusterService,
     private readonly _datacenterService: DatacenterService,
+    private readonly _matDialog: MatDialog,
     private readonly _matDialogRef: MatDialogRef<EditClusterComponent>,
     private readonly _notificationService: NotificationService,
     private readonly _settingsService: SettingsService,
@@ -563,9 +567,37 @@ export class EditClusterComponent implements OnInit, OnDestroy {
       .listBackupStorageLocation(projectID)
       .pipe(takeUntil(this._unsubscribe))
       .subscribe(cbslList => {
-        this.backupStorageLocationsList = cbslList;
-        this.backupStorageLocationLabel = cbslList.length ? BSLListState.Ready : BSLListState.Empty;
+        this.backupStorageLocationsList = cbslList.filter(bsl => this._isBackupStorageLocationAvailable(bsl));
+        this.backupStorageLocationLabel = this.backupStorageLocationsList.length ? BSLListState.Ready : BSLListState.Empty;
+
+        const backupStorageLocationControl = this.form.get(Controls.BackupStorageLocation);
+        if (backupStorageLocationControl && !this.backupStorageLocationsList.some(bsl => bsl.name === backupStorageLocationControl.value)) {
+          backupStorageLocationControl.reset();
+        }
       });
+  }
+
+  onBackupStorageLocationSelectionChange(event: MatSelectChange<string>): void {
+    if (event.value !== this.createBackupStorageLocationOptionValue) {
+      return;
+    }
+
+    const backupStorageLocationControl = this.form.get(Controls.BackupStorageLocation);
+    const previousValue = this.cluster.spec?.backupConfig?.backupStorageLocation?.name ?? '';
+    backupStorageLocationControl.setValue(previousValue, {emitEvent: false});
+
+    this._matDialog
+      .open(AddBackupStorageLocationDialogComponent, {
+        data: {projectID: this.projectID},
+      })
+      .afterClosed()
+      .pipe(take(1), filter(Boolean))
+      .subscribe(() => this._getCBSL(this.projectID));
+  }
+
+  private _isBackupStorageLocationAvailable(bsl: BackupStorageLocation): boolean {
+    const status = bsl.status?.phase || bsl.spec?.status || '';
+    return status.toLowerCase() === 'available';
   }
 
   getObservable(): Observable<Cluster> {
--- a/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -18,7 +18,6 @@ import {MatDialog, MatDialogRef} from '@angular/material/dialog';
 import {MatSelectChange} from '@angular/material/select';
 import {ClusterBackupService} from '@app/core/services/cluster-backup';
 import {UserClusterConfigService} from '@app/core/services/user-cluster-config';
-import {AddBackupStorageLocationDialogComponent} from '@app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component';
 import {DynamicModule} from '@app/dynamic/module-registry';
 import {BackupStorageLocation} from '@app/shared/entity/backup';
 import {NODEPORTS_IPRANGES_SUPPORTED_PROVIDERS, NodeProvider} from '@app/shared/model/NodeProviderConstants';
@@ -582,7 +581,7 @@ export class EditClusterComponent implements OnInit, OnDestroy {
       });
   }
 
-  onBackupStorageLocationSelectionChange(event: MatSelectChange<string>): void {
+  async onBackupStorageLocationSelectionChange(event: MatSelectChange<string>): Promise<void> {
     if (event.value !== this.createBackupStorageLocationOptionValue) {
       return;
     }
@@ -591,6 +590,10 @@ export class EditClusterComponent implements OnInit, OnDestroy {
     const previousValue = this.cluster.spec?.backupConfig?.backupStorageLocation?.name ?? '';
     backupStorageLocationControl.setValue(previousValue, {emitEvent: false});
 
+    const AddBackupStorageLocationDialogComponent = await DynamicModule.AddBackupStorageLocationDialogComponent;
+    if (!AddBackupStorageLocationDialogComponent) {
+      return;
+    }
     this._matDialog
       .open(AddBackupStorageLocationDialogComponent, {
         data: {projectID: this.projectID},
--- a/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -14,8 +14,7 @@
 
 import {ChangeDetectorRef, Component, Input, OnDestroy, OnInit} from '@angular/core';
 import {FormBuilder, FormControl, FormGroup, Validators} from '@angular/forms';
-import {MatDialog, MatDialogRef} from '@angular/material/dialog';
-import {MatSelectChange} from '@angular/material/select';
+import {MatDialogRef} from '@angular/material/dialog';
 import {ClusterBackupService} from '@app/core/services/cluster-backup';
 import {UserClusterConfigService} from '@app/core/services/user-cluster-config';
 import {DynamicModule} from '@app/dynamic/module-registry';
@@ -47,6 +46,7 @@ import {Datacenter, SeedSettings} from '@shared/entity/datacenter';
 import {AdminSettings, StaticLabel} from '@shared/entity/settings';
 import {KeyValueEntry} from '@shared/types/common';
 import {AdmissionPlugin, AdmissionPluginUtils} from '@shared/utils/admission-plugin';
+import {isBackupStorageLocationAvailable} from '@shared/utils/backup';
 import {
   CLUSTER_DEFAULT_NODE_SELECTOR_HINT,
   CLUSTER_DEFAULT_NODE_SELECTOR_NAMESPACE,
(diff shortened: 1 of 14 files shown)


modules/web/src/app/wizard/step/cluster/template.html
```diff
@@ -494,11 +494,13 @@ <h4>IPv6</h4>
             <mat-form-field>
               <mat-label>{{backupStorageLocationLabel}}</mat-label>
               <mat-select [formControlName]="Controls.BackupStorageLocation"
+                          (selectionChange)="onBackupStorageLocationSelectionChange($event)"
                           disableOptionCentering>
                 @for (bsl of backupStorageLocationsList; track bsl) {
                 <mat-option [value]="bsl.name">{{bsl.displayName}}
                 </mat-option>
                 }
+                <mat-option [value]="createBackupStorageLocationOptionValue">+ Create Backup Storage Location</mat-option>
```
ahmadhamzh, line 503: I don’t think adding an option to create a new BSL inside the dropdown is the right approach. I’d prefer something similar to how we handle SSH keys.

I think something like this makes more sense:
Image

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -529,11 +532,48 @@ export class EditClusterComponent implements OnInit, OnDestroy {
...
+        if (
+          backupStorageLocationControl &&
+          !this.backupStorageLocationsList.some(bsl => bsl.name === backupStorageLocationControl.value)
+        ) {
+          backupStorageLocationControl.reset();
+        }
       });
   }
 
+  async onBackupStorageLocationSelectionChange(event: MatSelectChange<string>): Promise<void> {
```
ahmadhamzh, line 550: Async/await is working fine, but since the rest of the component uses RxJS pipelines, I think it’s better to stay consistent with that approach.

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -529,11 +532,48 @@ export class EditClusterComponent implements OnInit, OnDestroy {
...
+    const AddBackupStorageLocationDialogComponent = await DynamicModule.AddBackupStorageLocationDialogComponent;
+    if (!AddBackupStorageLocationDialogComponent) {
+      return;
+    }
+    this._matDialog
+      .open(AddBackupStorageLocationDialogComponent, {
+        data: {projectID: this.projectID},
+      })
+      .afterClosed()
+      .pipe(take(1), filter(Boolean))
```
ahmadhamzh, line 568: Maybe something like this would be more consistent with the rest of the code.

```suggestion
from(DynamicModule.AddBackupStorageLocationDialogComponent)
      .pipe(
        filter(Boolean),
        switchMap(component =>
          this._matDialog.open(component, {data: {projectID: this.projectID}}).afterClosed()
        ),
        take(1),
        filter(Boolean),
      )
```

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -529,11 +532,48 @@ export class EditClusterComponent implements OnInit, OnDestroy {
...
+      .subscribe(() => this._getCBSL(this.projectID));
+  }
+
+  private _isBackupStorageLocationAvailable(bsl: BackupStorageLocation): boolean {
+    const status = bsl.status?.phase || bsl.spec?.status || '';
```
ahmadhamzh, line 573: i don't think we need to check for status cause it will never be `available`
```suggestion
   return bsl.status?.phase.toLowerCase() === 'available'
```

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -217,4 +225,64 @@ export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestro
     }
     return bsl;
   }
+
+  private _dnsNameValidator(): ValidatorFn {
```
ahmadhamzh, line 229: We usually add validators in `src/app/shared/validators`

ahmadhamzh (changes requested):

modules/web/src/app/wizard/step/cluster/template.html
SamuAlfageme: Good point! I forgot this pattern exists already somewhere in the Dashboard. I'll adapt the PR to use it too  👍

SamuAlfageme: @ahmadhamzh btw, for the "edit UC" view, opening the 'Create BSL' modal over the 'Edit Cluster' modal is something that is not done somewhere else in the Dashboard IIRC, does that work for you?

ahmadhamzh: Good point — I think we shouldn’t have that at all. So for cluster editing, we shouldn’t include an option to add a storage location at least for now.

KhizerRehan: IMO, dialog over a dialog is not a problem because i have seen this sort of behaviour in other applicaiton when supporting inline option e.g add-bsl dialog but I think what Ahmed said we shouldn't provide this option for edit cluster?

But I am also thinking what if there are existing cluster and need to backup in different locaiton how those scenario would be handled?

KhizerRehan: I like suggestion what ahmed has suggested of adding "+ Add BSL Button"  which is better instead of adding inside Dropdown option

- More clear/explicit
- And we can opt this change adding behaviour for wizard (if not going for edit dialog)

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
KhizerRehan: We do have a file `shared/validators/others.ts` where have defined all regex in case needed in other place IMO, regex that is reusable can be put into this file

let me wdyt?

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -217,4 +225,64 @@ export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestro
...
+    return (control: AbstractControl): ValidationErrors | null => {
+      const value = control.value?.trim();
+      if (!value) {
+        return null;
+      }
+
+      const labels = value.split('.');
+      return labels.every(label => this._dnsLabelRegex.test(label)) ? null : {invalidDnsName: true};
```
KhizerRehan, line 237: just to double confirm?

values like
- us-east-1  ✅ 
- us east 1  ❌ 
- us-east-1-  ❌ 
- _us-east-1  ❌ 

are not valid DNS

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -217,4 +225,64 @@ export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestro
...
+  }
+
+  private _endpointURLValidator(): ValidatorFn {
+    return (control: AbstractControl): ValidationErrors | null => {
+      const value = control.value?.trim();
+      if (!value) {
+        return null;
+      }
+
+      if (!/^https?:\/\//i.test(value)) {
```
KhizerRehan, line 248: Just to double confirm if following values are not valid test value?

- http://minio.local.  ❌  (FQDN trailing dot rejected before)
- https://[::1]:9000 (IPv6) ❌ 
- https://s3_internal.lan:9000 (underscore host) ❌

modules/web/src/app/wizard/step/cluster/component.ts
```diff
@@ -1215,11 +1219,49 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
...
+      .afterClosed()
+      .pipe(take(1))
+      .pipe(filter(Boolean))
+      .subscribe(() => this._getCBSL(this._selectedProjectID));
+  }
+
+  private _isBackupStorageLocationAvailable(bsl: BackupStorageLocation): boolean {
+    const status = bsl.status?.phase || bsl.spec?.status || '';
+    return status.toLowerCase() === 'available';
+  }
```
KhizerRehan, line 1263: FYI: 
This comment should be taken into consideraiton incase we OPT to choose to changes in (Edit Cluster Dialog

_Otherwise ignore this refactoring_

I see this method duplicate can we move to `src/app/shared/entity/backup.ts` path wdyt?

I was checking codebase something we are doing https://github.com/kubermatic/dashboard/blob/bdafe71cc7f8e04fa644a5201c363a093a86b6a8/modules/web/src/app/shared/entity/health.ts#L76 like that using `namespaces` concept.


Ideally, i though it shouldn't be part of `entity/` folder bcz IMO, we have kept all classes or interfaces but also it make sense to keep backup relateed things there even if there are shared utils as like defined here
`src/app/shared/utils`

something like we are doing here https://github.com/kubermatic/dashboard/blob/bdafe71cc7f8e04fa644a5201c363a093a86b6a8/modules/web/src/app/shared/utils/cluster.ts#L42

as this method is used in both place
- wizard
- edit cluster dialog

KhizerRehan (commented):
Feedback(PTAL)

modules/web/src/app/wizard/step/cluster/template.html
ahmadhamzh: > But I am also thinking what if there are existing cluster and need to backup in different locaiton how those scenario would be handled?

users can add them from the cluster backups > Storage Locations and then use them in the edit dialog

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
ahmadhamzh: well after another look i don't think we need this method just to repeat the same regex check.
we can update the regex value to include the `.`

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -64,6 +64,7 @@ enum Controls {
 })
 export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestroy {
   private readonly _unsubscribe = new Subject<void>();
+  private readonly _dnsLabelRegex = /^[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?$/i;
```
ahmadhamzh, line 67: this can be change to include the `.` and then we don't need to use `_dnsNameValidator()` method.
we can add 
`export const DNS_VALIDATOR = Validators.pattern(/^[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?(?:\.[a-z0-9](?:[a-z0-9-]{0,61}[a-z0-9])?)*$/i)
`

 to the `shared/validators/others.ts` file

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -108,12 +109,18 @@ export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestro
         this._config.bslObject?.spec.backupSyncPeriod ?? '0',
         CBSL_SYNC_PERIOD
       ),
-      [Controls.Region]: this._builder.control(this._config.bslObject?.spec.config?.region ?? ''),
-      [Controls.Endpoints]: this._builder.control(this._config.bslObject?.spec.config?.s3Url ?? ''),
+      [Controls.Region]: this._builder.control(this._config.bslObject?.spec.config?.region ?? '', [
+        this._dnsNameValidator(),
```
ahmadhamzh, line 113:
```suggestion
        DNS_VALIDATOR,
```

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -217,4 +225,64 @@ export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestro
...
+        return null;
+      }
+
+      if (!/^https?:\/\//i.test(value)) {
+        return {invalidEndpointUrl: true};
+      }
+
+      try {
+        const url = new URL(value);
+        if (!['http:', 'https:'].includes(url.protocol)) {
```
ahmadhamzh, line 254: i think if we have this check then we don't need the the previous condition `if (!/^https?:\/\//i.test(value))`  cause i think both do the same check, right ?

KhizerRehan:
@SamuAlfageme  thanks for the contribution!

Could you please review the feedback and let us know if anything is unclear? We’ve added a few comments for clarity and alignment.

If possible, please address the requested updates. Appreciate your effort.

KhizerRehan:
/hold
/retitle WIP: [UC Backup] BackupStorageLocation UX improvements

SamuAlfageme:
Hi @KhizerRehan, thank you for taking the time (together with @ahmadhamzh) to review this PR! I've been a bit busier than usual, but I should find some time to address your comments later this week.

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/component.ts
```diff
@@ -173,4 +194,9 @@ export class BackupStorageLocationsListComponent implements OnInit, OnDestroy {
         this.dataSource.data = this.backupStorageLocations;
       });
   }
+
+  private _isBSLAvailable(bsl: BackupStorageLocation): boolean {
```
ahmadhamzh: we can use the new shared function `isBackupStorageLocationAvailable`

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/template.html
```diff
@@ -103,12 +105,32 @@
...
+      </mat-error>
+      }
     </mat-form-field>
     <mat-form-field>
-      <mat-label>Endpoints</mat-label>
+      <mat-label>Endpoint URL</mat-label>
       <input [formControlName]="Controls.Endpoints"
              matInput>
       <mat-hint>Specify the AWS S3 URL here.</mat-hint>
+      @if (form.get(Controls.Endpoints).hasError('required')) {
```
ahmadhamzh: i have a concern, Why are `Region` and `Endpoint URL` required?
Per the [Velero AWS plugin docs](https://github.com/velero-io/velero-plugin-for-aws/blob/main/backupstoragelocation.md#:~:text=%3A%20cloud%2Dcredentials-,config%3A,-%23%20The%20AWS), all BSL config params are optional. Also in the dialog, checking "Add custom config" makes both optional again anyway, so the same object can be created without them.

ahmadhamzh (commented):
PTAL:

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/template.html
ahmadhamzh: because i tried creating a `bsl` without any endpoints, and that worked fine. the status was `available`. however, without a region it didn't work, and i got the error `invalid region: region was not a valid dns name`.

i'm not sure if that means the region is required or if we're sending something wrong in the api.

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/component.ts
KhizerRehan: thx fixed

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/template.html
KhizerRehan: Hi, @ahmadhamzh  as per internal discussion in daily (Since it was decided that we can revert "Required validator" for region/endpoint URL.

Please see this [commit](https://github.com/kubermatic/dashboard/pull/7972/changes/ca2992b547755f42f79473c5c9d1a0429b3673cc)

As per Velero Docs: It looks optional those fields from UI side I have update  "mat-hint" as indication that field is optional (let me know wdyt) if "Optional" keyword can be removed.

Reasons:
- Clear / Quick indication those fields optional
- I saw when "Add Custom Config" option was clicked

region/endpoint were : ''  ->  Empty config values were previously sent as empty strings, which Velero  might rejects during validation (invalid region: region was not a valid dns name). So I have Omit those empty fields as part of payload (but incase filled) yml will show correctly (which IMO, is fine)

See Screenshot
 
**Dev**
Screenshot 2026-08-05 at 5 48 24 PM

Screenshot 2026-08-05 at 5 48 29 PM


**PR:**
Updated Image

image

Screenshot 2026-08-05 at 5 48 03 PM


Rest of feedback is catered:


/cc: @buraksekili  @rajaSahil let me if any input or concerns

Thanks

KhizerRehan: - [Internal Slack Ref](https://kubermatic.slack.com/archives/C08PNJ0R8F3/p1784631811177699)

/cc: @ahmadhamzh

modules/web/src/app/dynamic/enterprise/cluster-backups/list/backup-storage-location/add-dialog/component.ts
```diff
@@ -180,7 +191,39 @@ export class AddBackupStorageLocationDialogComponent implements OnInit, OnDestro
...
+      return false;
+    }
+
+    if (config?.s3Url && endpointUrlValidator()(new FormControl(config.s3Url))) {
+      this.customConfigError = ENDPOINT_URL_ERROR_MESSAGE;
+      return false;
+    }
+
+    return true;
   }
```
KhizerRehan, line 227: @ahmadhamzh this method is added to handle as case

Found an issue while testing: users could enter invalid values in the YAML editor while the form remained valid. I’ve handled validation both ways.

* Invalid `s3Url` or `region` in the YAML editor now disables the form and shows the same validators as the form fields.
* Reused the same validation regex as the form fields.
* YAML editor now displays the same validation errors as the form fields.


Screenhots:

Image
Image

ahmadhamzh (approved):
/approve

KhizerRehan:
/hold
This PR will be merged after Release 2.31: [Internal Slack Discussion](https://kubermatic.slack.com/archives/C08PNJ0R8F3/p1786370792390759?thread_ts=1784631811.177699&cid=C08PNJ0R8F3)

SamuAlfageme:
@ahmadhamzh @KhizerRehan Sorry folks, some stuff that dragged my attention away from this PR happened, and then summer happened and I completely forgot about coming back to apply your review comments... thank you for taking it over @KhizerRehan <3

Great work 🙌

KhizerRehan:
@ahmadhamzh 

I have resolved the merge conflicts and rebased the PR with `main`.

We have removed the `required` validators for Region and Endpoint URL, but kept the regex validation for better user experience.

Since whether these fields are required depends on the S3 storage being used (AWS S3 or other S3-compatible storage), handling this dynamically would require additional custom validation logic.

To avoid adding unnecessary complexity, I initially thought keeping these fields required would be simpler. However, the current changes look good to me, and I think we can proceed with merging and closing this PR.

If needed, we can revisit the storage-specific validation separately in the future.



Thank you

/cc: @buraksekili 

Screenshot 2026-09-01 at 5 00 36 PM

ahmadhamzh (approved):
/approve

PR #7999, opened 2026-04-14
Fix KubeLB option precedence enforced over enabled

### LINKED ISSUE #7902: KubeLB enhancement
### What happened
there is two issues:
* **KubeLB sub-options remain enabled after disabling main KubeLB option**
 When creating a cluster, if the user enables KubeLB and selects any of its sub-options (e.g., Use Load Balancer Class or Enable Gateway API), then later disables the main KubeLB option, the selected sub-option values are still being sent.

* **Double tool-tip issue**

Image

KhizerRehan:
This PR fixes KubeLB option visibility so that enforced takes precedence over the enabled flag

#### Behaviour Matrix

| enforced | enabled | enableForAllDatacenters | Result | Rationale |
|----------|---------|-------------------------|--------|-----------|
| true | any | any | show + locked | enforced wins |
| false / undef | true | any | show | enabled wins |
| false / undef | false | any | hide | explicit per-DC disable |
| false / undef | undef | true | show | seed fallback |
| false / undef | undef | false / undef | hide | no signal to show |

Fixes https://github.com/kubermatic/dashboard/issues/7902

**What type of PR is this?**
/kind bug

```release-note
Fixes KubeLB option precedence so enforced datacenters always show the option regardless of the enabled flag.
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
+++ b/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
@@ -511,15 +511,15 @@ export class EditClusterComponent implements OnInit, OnDestroy {
   }
 
   private _isKubeLBEnabled(datacenter: Datacenter, seedSettings: SeedSettings): boolean {
-    // If enabled is explicitly false at datacenter level, hide regardless of enforced
+    if (datacenter.spec.kubelb?.enforced) {
+      return true;
+    }
     if (datacenter.spec.kubelb?.enabled === false) {
       return false;
     }
-    // If enforced or enabled at datacenter level, show
-    if (datacenter.spec.kubelb?.enforced || datacenter.spec.kubelb?.enabled) {
+    if (datacenter.spec.kubelb?.enabled) {
       return true;
     }
-    // Fall back to seed-level setting
     return !!seedSettings?.kubelb?.enableForAllDatacenters;
   }
 
--- a/modules/web/src/app/shared/components/cluster-summary/component.ts
+++ b/modules/web/src/app/shared/components/cluster-summary/component.ts
@@ -90,14 +90,14 @@ export class ClusterSummaryComponent implements OnInit {
   }
 
   get kubeLBAssignPublicIPWarning(): boolean {
-    // Check all sources: cluster spec, datacenter enforced/enabled, and seed-level setting
-    // Respect explicit datacenter disable (enabled === false)
     const dcKubeLB = this.datacenter?.spec?.kubelb;
     let isKubeLBActive = !!this.cluster?.spec?.kubelb?.enabled;
     if (!isKubeLBActive) {
-      if (dcKubeLB?.enabled === false) {
+      if (dcKubeLB?.enforced) {
+        isKubeLBActive = true;
+      } else if (dcKubeLB?.enabled === false) {
         isKubeLBActive = false;
-      } else if (dcKubeLB?.enforced || dcKubeLB?.enabled) {
+      } else if (dcKubeLB?.enabled) {
         isKubeLBActive = true;
       } else {
         isKubeLBActive = !!this.seedSettings?.kubelb?.enableForAllDatacenters;
--- a/modules/web/src/app/wizard/step/cluster/component.ts
+++ b/modules/web/src/app/wizard/step/cluster/component.ts
@@ -1367,15 +1367,15 @@ export class ClusterStepComponent extends StepBase implements OnInit, ControlVal
   }
 
   private _isKubeLBEnabled(datacenter: Datacenter, seedSettings: SeedSettings): boolean {
-    // If enabled is explicitly false at datacenter level, hide regardless of enforced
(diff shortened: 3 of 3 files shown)


KhizerRehan:
@shinushaju kindly, check behaviour matrix. It should behave according to precedence order defined. However the issue of not returning `false` value from backend still needs to be fixed when explicitly set

References:
- https://navendu.me/tils/9-5-25-omitempty-option-omits-boolean-fields-with-false-values/
- https://stackoverflow.com/questions/72206365/go-omitempty-bool-field-not-showing-false-type

shinushaju (approved):

KhizerRehan:
/hold

Concern/Clarification Added

**Priority order:** `enforced > enabled > enableForAllDatacenters`

As discussed with @ahmadhamzh, it was mentioned that `enableForAllDatacenters` should act as a fallback when both `enforced` and `enabled` are `false`. However, I want to clarify a concern based on what @shinushaju mentioned in this ticket:
[https://github.com/kubermatic/kubermatic/issues/15713](https://github.com/kubermatic/kubermatic/issues/15713)

> *When KubeLB is disabled for a specific datacenter in the Seed, the API should explicitly return `enabled: false`, and the KKP Dashboard should hide the KubeLB option for that datacenter accordingly.*

If that is the expected behavior, and `enableForAllDatacenters` is also considered after `enabled`, then it implies that we can only hide the KubeLB option when **both** conditions are met:

* `enableForAllDatacenters: false`
* `enabled: false` (per datacenter)

This raises another question: if we want to enable KubeLB for all `other` datacenters, would we then need to explicitly set `enabled: true` for each individual datacenter?

In my opinion, `enableForAllDatacenters` acts as a global default flag, which should enable KubeLB for all datacenters unless explicitly overridden. In that case, a per-datacenter `enabled: false` should always take precedence over `enableForAllDatacenters`.

Otherwise, the only way to hide KubeLB would be to set both flags to `false`, 

Based on this, I’m trying to understand whether the logic introduced in the PR aligns with this intended behavior.

Therefore this PR is on hold for further discussion

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -511,15 +511,15 @@ export class EditClusterComponent implements OnInit, OnDestroy {
   }
 
   private _isKubeLBEnabled(datacenter: Datacenter, seedSettings: SeedSettings): boolean {
-    // If enabled is explicitly false at datacenter level, hide regardless of enforced
+    if (datacenter.spec.kubelb?.enforced) {
+      return true;
+    }
     if (datacenter.spec.kubelb?.enabled === false) {
```
ahmadhamzh, line 517: according to my understand this sholdn't return false unless the  `seedSettings?.kubelb?.enableForAllDatacenters` is also false, so we need to check that before returning false

ahmadhamzh (commented):
so from what i understand if `seedSettings?.kubelb?.enableForAllDatacenters` is true then kubeLB will be available in all datacenters regardless of the `dc.spec.kubelb?.enabled` value, 
@ahmedwaleedmalik is that correct ?

KhizerRehan:
@shinushaju what is the status from backend side?

ahmedwaleedmalik:
> so from what i understand if `seedSettings?.kubelb?.enableForAllDatacenters` is true then kubeLB will be available in all datacenters regardless of the `dc.spec.kubelb?.enabled` value, @ahmedwaleedmalik is that correct ?

The precedence logic in this PR is correct 👍 

@ahmadhamzh re: your review comment - per datacenter`enabled: false` should **not** be gated on `enableForAllDatacenters`. An explicit per-DC disable must win over the seed-wide default; otherwise admins can't carve out a DC and disable kubelb for it from a globally-enabled rollout without flipping the global flag.

The remaining blocker is backend (kubermatic/kubermatic#15713), not this PR. This is a typical go problem, due to `omitempty` we can't distinguish between unset and explicit false. This is required for the frontend to determine this like @KhizerRehan mentioned. So we need to fix this on the backend i.e. KKP side first.

KhizerRehan:
/approved

🧪  Tested locally based Backend PR got [merged](https://github.com/kubermatic/kubermatic/pull/15954) but required to update kkp dependency in dashboard.

image

Should we update KKP dependency in this PR upto this [kkp commit: ](https://github.com/kubermatic/kubermatic/commit/b2c74a27a74d376b5613982a89e2a2ccaf3fb731) or main?

cc: @ahmadhamzh @archups @buraksekili

buraksekili:
let's update KKP dependencies (in go.mod file) as part of this PR 👍

KhizerRehan:
/hold
Need to rebase and merge following PR: https://github.com/kubermatic/dashboard/pull/8116

ahmadhamzh (approved):
/approve

PR #8010, opened 2026-04-17
Handle VSphere Project-Level OS Restrictions

### LINKED ISSUE #8009: VSphere Ignores Project-Level OS Restrictions
### What happened

When an operating system is disabled at the **project level** (Edit Project → Allowed Operating Systems), the VSphere provider does not respect the restriction. The OS toggle in the UI hides the disabled OS, but VSphere's default template selection can still reference it results an inconsistent state between what the user sees and what VSphere picks.

When disabled OS at (Project Level)

Image

Currently: Shows wrong (VmTemplate) -> Workaround select OS again to select correct VmTemplate 

Image

### Expected behavior

VSphere should respect project-level OS restrictions. The effective allowed-OS list should be the intersection of admin-level and project-level settings (more restrictive wins), matching the OS selector shown in the UI.

Image

### How to reproduce

1. As an admin, ensure Ubuntu is enabled in Admin Settings → Allowed Operating Systems.
2. Edit Project → Allowed Operating Systems → disable Ubuntu.
3. Navigate to wizard create a new cluster using the VSphere provider.
4. Navigate to the Initial Nodes step.  -> check "VM Template" -> points to "ubuntu" which is wrong as "ubuntu" OS is disabled.

**Observed:** The OS toggle hides Ubuntu, but VSphere may still default to or reference the Ubuntu template.

### Current workaround

- Disable the restricted operating system at the admin (global) level in addition to the project level.
- On Initial node step **reselect** OS to trigger change and select correct VMTemplate

### Related Issue 
- Following Issue already resolved the selection of NEXT availaible OS. Therefore selection of Template would work correctly in case "Ubuntu" is disabled.
- Reference: https://github.com/kubermatic/dashboard/pull/7927

KhizerRehan:

This PR fixes an issue where the VSphere provider ignored project-level operating system restrictions and could default its
template selection to an OS that was disabled at the project scope, even though the OS selector correctly hid it.

#### After Fix
image

Defaults to Correct VMTemplate based on Available/Selected OS

Screenshot 2026-04-17 at 4 05 08 PM
  
Fixes #8009 

**What type of PR is this?**
/kind bug

```release-note
Fixed VSphere provider ignoring project-level allowed operating system restrictions during cluster creation.
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
TBD
```

### DIFF
--- a/modules/web/src/app/core/services/node-data/service.ts
+++ b/modules/web/src/app/core/services/node-data/service.ts
@@ -110,10 +110,6 @@ export class NodeDataService {
     return this._allowedOperatingSystems$.asObservable();
   }
 
-  get allowedOperatingSystems(): AllowedOperatingSystems {
-    return this._allowedOperatingSystems$.value;
-  }
-
   set allowedOperatingSystems(value: AllowedOperatingSystems) {
     this._allowedOperatingSystems$.next(value);
   }
--- a/modules/web/src/app/core/services/node-data/service.ts
+++ b/modules/web/src/app/core/services/node-data/service.ts
@@ -38,10 +38,11 @@ import {VSphereService} from '@core/services/provider/vsphere';
 import {PresetsService} from '@core/services/wizard/presets';
 import {MachineDeployment, OPERATING_SYSTEM_PROFILE_ANNOTATION} from '@shared/entity/machine-deployment';
 import {NodeNetworkSpec, OperatingSystemSpec, Taint} from '@shared/entity/node';
+import {AllowedOperatingSystems, DEFAULT_ADMIN_SETTINGS} from '@shared/entity/settings';
 import {OperatingSystem} from '@shared/model/NodeProviderConstants';
 import {NodeData} from '@shared/model/NodeSpecChange';
 import _ from 'lodash';
-import {ReplaySubject} from 'rxjs';
+import {BehaviorSubject, Observable, ReplaySubject} from 'rxjs';
 import {NodeDataAlibabaProvider} from './provider/alibaba';
 import {NodeDataAnexiaProvider} from './provider/anexia';
 import {NodeDataAWSProvider} from './provider/aws';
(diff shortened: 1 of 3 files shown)


modules/web/src/app/node-data/basic/provider/vsphere/component.ts
```diff
@@ -126,11 +124,14 @@ export class VSphereBasicNodeDataComponent extends BaseFormValidator implements
       this._nodeDataService.nodeData.spec.operatingSystem
     );
 
-    this._settingsService.adminSettings.pipe(take(1)).subscribe(settings => {
-      if (settings.allowedOperatingSystems) {
-        this.allowedOperatingSystems = settings.allowedOperatingSystems;
-      }
-    });
+    this._nodeDataService.allowedOperatingSystems$
```
KhizerRehan, line 127: Note for Reviewer:

It was observed that the parent component already handles admin and project-level filtering for `allowedOperatingSystems`. Since this logic exists at the `NodeDataComponent` level and is consumed by providers via `NodeDataService`, duplicating the same filtering in the vSphere component is unnecessary.

To avoid duplication, the vSphere component now reads the filtered value directly from the service. The merged value is propagated through the allowedOperatingSystems `setter`, ensuring a consistent and centralised flow also keeping single source of truth from service.

let me know if any concerns

modules/web/src/app/core/services/node-data/service.ts
```diff
@@ -102,6 +106,18 @@ export class NodeDataService {
     return this._operatingSystemChanges;
   }
 
+  get allowedOperatingSystems$(): Observable<AllowedOperatingSystems> {
+    return this._allowedOperatingSystems$.asObservable();
+  }
+
+  get allowedOperatingSystems(): AllowedOperatingSystems {
```
ahmadhamzh: I think you don’t use this getter anywhere, right ?

ahmadhamzh (commented):
PTAL

ahmadhamzh (approved):
/approve

PR #8028, opened 2026-04-22
Fix azure availability zone selector loader

mgoltzsche:
**What this PR does / why we need it**:

Apparently the project selector Observable (`projectService.selectedProject`) that is used as prerequisite for the availability zone loader emits events repeatedly, each subsequent event cancelling the previously started request to load the availability zones and restarting the request.
The solution is to only handle the first event of the selectedProject Observable and ignore subsequent ones, letting the availability zone request finish loading.

**Which issue(s) this PR fixes**:

Fixes #7784

**What type of PR is this?**
/kind bug

**Special notes for your reviewer**:

This PR was created in collaboration with @ahmadhamzh.

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
Fixes the Azure availability zone selector data population within the MachineDeployment edit dialog
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
https://github.com/kubermatic/kubermatic/issues/15798
```

Issue #7784, opened 2026-01-20
Azure Availability zones selector times out at the MachineDeployment creation


kron4eg:
### What happened

When trying to add a new Machine Deployment in the existing Azure cluster selector "Availability zones" fails to load with the timeout.

### Expected behavior

"Availability zones" are loaded.

### How to reproduce

* Create Azure cluster
* DO NOT enable "Availability Sets" in the cluster creation wizard
* Wait until control plane is created and cluster is settled
* Add new Machine Deployment, observe the timeout (see screenshot)

<img width="2000" height="2182" alt="Image" src="https://github.com/user-attachments/assets/7077cc61-1c3f-4678-bc68-471794729ec8" />

### Environment
- UI Version: latest main (at the time g15dccb2d5)
- API Version: latest main (at the time g15dccb2d5)
- Domain:
- Others:

### Current workaround

None

### Affected user persona

User cluster administrator

mgoltzsche:
fwiw node size and zone selector take 15-20s to load:
<img width="1131" height="939" alt="Image" src="https://github.com/user-attachments/assets/d3b03399-af72-461e-b898-c2d5e4e83d0b" />

UPDATE: Turns out the long latency or a timeout wasn't the actual problem but that the frontend cancels and retries the request immediately repeatedly within the cluster edit dialog.

PR #8030, opened 2026-04-22
Update development documentation & log warning on seed connectivity failure

mgoltzsche:

* Update the documentation to clarify which directory the commands refer to and to refer to the `run-api.sh` script and mention that the old `kubermatic-api` binary should be deleted before testing a newer repo revision (otherwise it may result in weird behaviour as explained in the next point).
* Log a warning when the seed cluster connectivity fails as opposed to swallowing the error and letting the endpoint return an empty user cluster list. This is to avoid confusion I ran into when I could see master resources but no seed-level/user cluster resources while at the same time no error was logged/shown (apparently due to an older build of the `kubermatic-api` binary).

Relates to #7784 and #8028

**What type of PR is this?**
/kind documentation

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -12,7 +12,7 @@ This approach assumes that you have all required dependencies available on your 
 
 ## Preparation
 
-Install the application dependencies by running:
+Install the application dependencies by running within the `modules/web` directory:
 
 ```bash
 npm ci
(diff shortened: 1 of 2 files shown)


buraksekili (approved):
/lgtm
/approve

PR #8043, opened 2026-04-27
Add -oidc-authflow-redirect-uri flag to fix redirect URI behind ingress

buraksekili:

The authflow login and callback handlers use `buildRedirectURI` to construct the redirect URI sent to Dex. This function checks `r.TLS` to decide the scheme - if TLS is nil, it uses `http://`. When the API runs behind a TLS-terminating ingress, `r.TLS` is always nil, so the redirect URI becomes `http://<host>/api/v2/auth/callback`. Dex rejects this because it doesn't match the registered `https://` redirect URI.

This PR adds a `-oidc-authflow-redirect-uri` flag that lets operators explicitly set the redirect URI. When the flag is empty (the default), `buildRedirectURI` uses `https://` unconditionally instead of checking `r.TLS`. This is the correct behavior for production where the API sits behind an ingress.

For local development, the flag can be set to `http://127.0.0.1:8080/api/v2/auth/callback` (already added to `run-api.sh`).

Fixes the redirect URI mismatch error that occurs when logging into the dashboard behind an ingress.

**What type of PR is this?**
/kind bug

The existing `-oidc-issuer-redirect-uri` flag serves a different OAuth client (`kubermaticIssuer` for kubeconfig flows) and cannot be reused here. This new flag is specific to the dashboard auth authorization code flow.

Ref: https://github.com/kubermatic/dashboard/pull/7960

```release-note
NONE
```
> Its NONE because the feature itself is not released yet.

**Documentation**:
```documentation
TBD
```

> We should document this and allow KKP to configure this field.

**Test issue**:
```test-issue
NONE
```

### DIFF
--- a/modules/api/hack/run-api.sh
+++ b/modules/api/hack/run-api.sh
@@ -57,6 +57,7 @@ API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-client-id=$OIDC_ISSUER_CLIENT_ID"
 API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-client-secret=$OIDC_ISSUER_CLIENT_SECRET"
 API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-redirect-uri=$OIDC_ISSUER_REDIRECT_URI"
 API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-cookie-hash-key=$OIDC_ISSUER_COOKIE_HASH_KEY"
+API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-authflow-redirect-uri=http://127.0.0.1:8080/api/v2/auth/callback"
 
 echodate "Starting API..."
 set -x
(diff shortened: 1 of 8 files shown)


modules/api/hack/run-api.sh
```diff
@@ -57,6 +57,7 @@ API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-client-id=$OIDC_ISSUER_CLIENT_ID"
 API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-client-secret=$OIDC_ISSUER_CLIENT_SECRET"
 API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-redirect-uri=$OIDC_ISSUER_REDIRECT_URI"
 API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-issuer-cookie-hash-key=$OIDC_ISSUER_COOKIE_HASH_KEY"
+API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-authflow-redirect-uri=http://127.0.0.1:8080/api/v2/auth/callback"
```
KhizerRehan, line 60: Now, I am seeing error while running locally and i assume we need to register `127.0.0.1` do we need to register 127.0.0.1 in `redirectURIs`?

Image

buraksekili: Could you please update as follows locally, and see if it helps?
```suggestion
API_EXTRA_ARGS="$API_EXTRA_ARGS -oidc-authflow-redirect-uri=http://localhost:8080/api/v2/auth/callback"
```

buraksekili: @KhizerRehan  figured out this should be `localhost:8000`, not `localhost:8080`, but for now, the PR will be closed. I just noted this down for future reference.

PR #7675, opened 2025-11-03
[Migration] Run Migration to Update Control Flow Syntax and Self Closing Tag

KhizerRehan:
This PR migrates Angular templates from legacy control flow syntax (*ngIf, *ngFor, *ngSwitch) to the new control flow syntax ("@ if", "@ for", "@ switch") and updates self-closing tags to use the new syntax (<tag />). This is a modernization effort to align with Angular's current best practices and improve template readability.

Changes:

- Replaced "*ngIf" directives with "@ if" blocks
- Replaced "*ngFor" directives with "@ for" loops with track expressions
- Replaced "*ngSwitch" with "@ switch" blocks
- Updated self-closing HTML tags to use />

**Which issue(s) this PR fixes**:

Fixes #7674 7674

/kind chore

**What type of PR is this?**

**Special notes for your reviewer**:

**Documentation**:

```documentation
NONE
```

**Does this PR introduce a user-facing change? Then add your Release Note here**:

```release-note
Migrate Angular templates to new control flow syntax and self-closing  tags for code consistency and modern Angular standards compliance
```

PR #8049, opened 2026-04-29
Fix Duplicate ng-template Names

KhizerRehan:

Angular enforces unique ng-template names to avoid ambiguity in schematic processing, and this issue was discovered while testing the Angular migration script, as duplicate template names could lead to corrupt or incorrect transformation output.

> FYI: This restriction was introduced with Angular 17's control-flow schematic and is enforced in all
subsequent versions (Angular 18, 19, 20+).

#### **Angular GitHub references:**
- PR [#53204](https://github.com/angular/angular/pull/53204) — "Update CF migration to skip templates with duplicate ng-template names" (Nov 28, 2023)

Ref: https://github.com/kubermatic/dashboard/pull/7675

**What type of PR is this?**
/kind cleanup

**FYI:**
Found while running the migration script: it didn’t auto-convert some files and showed a warning, but this is a pure rename to satisfy Angular’s requirement for unique _ng-template_ names within a component, with no **behavior** change.

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/shared/components/tab-card/tab/component.ts
+++ b/modules/web/src/app/shared/components/tab-card/tab/component.ts
@@ -22,5 +22,5 @@ import {Component, Input, TemplateRef, ViewChild, ContentChild} from '@angular/c
 export class TabComponent {
   @Input() label: string;
   @ViewChild(TemplateRef) template: TemplateRef<any>;
-  @ContentChild('tabLabel') labelTemplate: TemplateRef<any>;
+  @ContentChild(TemplateRef) labelTemplate: TemplateRef<any>;
 }
--- a/modules/web/src/app/cluster/details/cluster/template.html
+++ b/modules/web/src/app/cluster/details/cluster/template.html
@@ -611,7 +611,7 @@ limitations under the License.
     </km-tab>
     @if (cluster.spec.opaIntegration?.enabled) {
     <km-tab label="OPA Constraints">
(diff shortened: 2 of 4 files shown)


modules/web/src/app/shared/components/tab-card/tab/component.ts
```diff
@@ -22,5 +22,5 @@ import {Component, Input, TemplateRef, ViewChild, ContentChild} from '@angular/c
 export class TabComponent {
   @Input() label: string;
   @ViewChild(TemplateRef) template: TemplateRef<any>;
-  @ContentChild('tabLabel') labelTemplate: TemplateRef<any>;
+  @ContentChild(TemplateRef) labelTemplate: TemplateRef<any>;
```
KhizerRehan, line 25: Note for Reviewer:

Why I did changed from "string" to TemplateRef

* `@ ContentChild('tabLabel')` uses a fixed template name as a shared contract between parent and child, so every `<km-tab>` must define `#tabLabel`, which can lead to duplicate names in the same template when multiple siblings exist.

* `@ ContentChild(TemplateRef)` has no naming contract and simply picks up the projected template per instance, so each `<km-tab>` resolves its own content independently with no collisions.

ahmadhamzh (approved):
/approve

PR #8059, opened 2026-05-07
Fix km-tab label regression introduced by #8049

KhizerRehan:
 This PR fixes the regression introduced by #8049 that broke the Clusters Overview page by causing the <km-tab> label-template content  query to capture unrelated TemplateRef instances produced by structural directives inside the tab body (e.g. *matHeaderCellDef from  mat-table).        

image

#### Why It Broke
@ContentChild(TemplateRef) matches the first TemplateRef instance in projected content. Angular structural directives such as    `*matHeaderCellDef, *matCellDef, *matRowDef, *ngIf,` etc. desugar to <ng-template> and produce TemplateRef instances. 

Fixes #

**What type of PR is this?**
/kind bug                                                 
/kind regression

```release-note
NONE
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/cluster/details/cluster/template.html
+++ b/modules/web/src/app/cluster/details/cluster/template.html
@@ -611,7 +611,7 @@ limitations under the License.
     </km-tab>
     @if (cluster.spec.opaIntegration?.enabled) {
     <km-tab label="OPA Constraints">
-      <ng-template>
+      <ng-template kmTabLabel>
         OPA Constraints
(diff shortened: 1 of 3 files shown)


KhizerRehan:
I could have reverted #8049 PR but IMO, directive approach is still better as it would help incase Angular [migrations](https://angular.dev/reference/migrations) run. There won't be issue due to duplicate names.

ahmadhamzh (approved):
/approve

PR #8107, opened 2026-06-03
Add per-cluster HTTP(S) proxy configuration

### LINKED ISSUE #7609: Allow per-cluster HTTP proxy configuration via KKP API/UI

Our customers would like the ability to set individual HTTP proxy settings for each of their user clusters. Currently, this configuration is only supported at the management layer and cannot be set per cluster.

**Request**: 
Expose HTTP proxy configuration options in the KKP API and UI so that individual users can configure HTTP proxy settings on a per-cluster basis.

**Internal reference:** 8379

KhizerRehan:
This PR adds the ability to configure a per-cluster HTTP(S) proxy (httpProxy/noProxy) from the dashboard API and UI, so clusters in restricted/air-gapped networks can route node egress through a proxy independent of the datacenter/seed defaults.

**Cluster Wizard**

Screenshot 2026-06-15 at 8 11 21 PM

Screenshot 2026-06-15 at 8 11 16 PM

**Cluster Edit**
Screenshot 2026-06-15 at 10 10 45 PM

**Cluster Summary**

image

**Cluster Details Page**

Screenshot 2026-06-03 at 5 56 49 PM

**Validations**
Screenshot 2026-06-03 at 11 30 10 AM

**Cluster Template Persistance**
Screenshot 2026-06-03 at 12 59 15 PM

Additional Change
Screenshot 2026-06-15 at 8 11 09 PM

  

Fixes #7609 

**What type of PR is this?**
/kind feature

```release-note
Add support for configuring a per-cluster HTTP(S) proxy (httpProxy/noProxy) from the cluster wizard, edit dialog, and cluster templates.
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/2236
```

**Test issue**:

```test-issue
TBD
```


buraksekili (commented):
/lgtm

LGTM from my side. I checked the attached screenshot - it helped me to review, thank you @KhizerRehan!

I'll leave the review of the UI changes to @ahmadhamzh.

modules/web/src/app/cluster/details/cluster/edit-cluster/template.html
```diff
@@ -156,6 +156,7 @@
       </km-chip-list>
       }
 
+
```
ahmadhamzh: is this a leftover empty line ?

modules/web/src/app/cluster/details/cluster/edit-cluster/template.html
```diff
@@ -416,6 +417,38 @@
       </mat-form-field>
       }
 
+      <!-- Per-cluster HTTP(S) proxy override (operating-system-manager). -->
+      <div>
+        <mat-card-header class="km-no-padding km-proxy-settings">
```
ahmadhamzh: `Proxy Settings` is sub field of `Network Configuration` right ? so i think we shouldn't use `mat-card-header` here

also i have a concern about the placement of `Proxy Settings`, should it be located under `proxy mode`, are they related to each other ?

modules/web/src/app/cluster/details/cluster/edit-cluster/template.html
```diff
@@ -416,6 +417,38 @@
...
+          <mat-card-title>Proxy Settings <i class="km-icon-info km-pointer"
```
ahmadhamzh: i think the info icon needs to be a little bit closer to the title.

Image

modules/api/pkg/test/e2e/utils/apiclient/models/proxy_settings.go
```diff
@@ -13,8 +13,8 @@ import (
 	"github.com/go-openapi/swag"
 )
 
-// ProxySettings ProxySettings allow configuring a HTTP proxy for the controlplanes
-// and nodes.
+// ProxySettings ProxySettings holds the per-cluster HTTP(S) proxy configuration. Empty values mean the
+// cluster inherits the datacenter/seed proxy settings.
```
ahmadhamzh: the reason this change happened is that the new type `ProxySettings` struct in `pkg/api/v1/types.go` has the same name as `kubermaticv1.ProxySettings` causing a swagger name conflict,
the generator resolved it by overwriting the existing `ProxySettings` definition with the new one.

the impact is documentation only so the API works correctly, it can be fixed by renaming the new type to something like `ClusterProxySettings`

ahmadhamzh (commented):
PTAL

modules/web/src/app/cluster/details/cluster/edit-cluster/template.html
KhizerRehan: - I used to align font-size similar to other headings to avoid extra css manually
- regarding layout position: I am thought placing at bottom is fine but i have don't have any input and not sure if they are related but i kept under Network Configuration bottom of all configurations

let me know wdyt or any better place?

modules/api/pkg/test/e2e/utils/apiclient/models/proxy_settings.go
KhizerRehan: Oh! I see you mean we ahve 2 levels ProxySettings
- Datacenter Level
- Cluster Level

To avoid confusion at Swagger level at API side we can cluster level -> to `ClusterProxySettings` but i think i checked KKP for seed and cluster level we are using same 

```
type ProxySettings struct 
```

should we remove the new one and use existing one thoughts?

@ahmadhamzh

modules/web/src/app/cluster/details/cluster/edit-cluster/template.html
ahmadhamzh: my bad i should mention this comment in `src/app/wizard/step/cluster/template.html` , but to be honest im not sure whether we need a header here or not, i think the edit dialog is not organised very well,
any way so at least this comment should be fixed in the wizard view

modules/api/pkg/test/e2e/utils/apiclient/models/proxy_settings.go
ahmadhamzh: > should we remove the new one and use existing one thoughts?

yeh i think that is better since it's already have the `httpProxy` and `noProxy` .

KhizerRehan: fixed: Changes Struct name to ClusterProxySettings

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -651,9 +669,26 @@ export class EditClusterComponent implements OnInit, OnDestroy {
...
+    patch.spec.componentsOverride = this._getProxyComponentsOverride();
+
     return this._clusterService.patch(this.projectID, this.cluster.id, patch, true).pipe(take(1));
   }
 
+  // Per-cluster HTTP(S) proxy override. Empty values clear the override so the cluster
+  // re-inherits the datacenter/seed proxy settings.
+  private _getProxyComponentsOverride(): ComponentSettings {
+    const noProxyRaw = this.form.get(Controls.NoProxy)?.value;
+    const noProxyValues: string[] = Array.isArray(noProxyRaw) ? noProxyRaw : (noProxyRaw?.tags ?? []);
```
ahmadhamzh: since the [chip list wrapped tags issue](https://github.com/kubermatic/dashboard/pull/8113) PR is merged maybe we can fetch the latest changes and fix it here, right ?

ahmadhamzh (commented):
one small comment

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
KhizerRehan: Thx!
Fixed

ahmadhamzh (approved):
/approve

PR #8108, opened 2026-06-04
Pass correct KubeVirt instancetype kind to providerSpec

soer3n:

The KubeVirt instancetype API response did not include a kind field, so the frontend derived it from the category with a hardcoded mapping: Custom →  VirtualMachineClusterInstancetype. This was wrong for user-deployed namespaced VirtualMachineInstancetype resources, which are also listed under the custom category. The machine providerSpec was created with the wrong kind, causing kkp ee resource quota validation in the kkp admission webhook to fail with  "instancetype not found".

this pr:
  - adds kind to the API response for each instancetype (correct per wrapper type)
  - removes the hardcoded category→kind switch in the frontend; uses kind from the API directly
  - fixes MachineDeployment edit/pre-selection to search the loaded list by name+kind to find the correct display category, with a name-only fallback for specs saved before this fix

Fixes #

**What type of PR is this?**

/kind bug
/kind regression

Depends on the [companion kubermatic pr](https://github.com/kubermatic/kubermatic/pull/15958) that fixes the ee webhook to handle user-deployed namespaced instancetypes. The kubermatic pr must be deployed first without the kind field in the api response, `selectedInstanceType?.kind` is undefined, which omits the field from the providerSpec and KubeVirt defaults back to VirtualMachineClusterInstancetype.

```release-note
none
```

**Documentation**:

```documentation
none
```

**Test issue**:

```test-issue
tbd
```

### DIFF
--- a/modules/api/cmd/kubermatic-api/swagger.json
+++ b/modules/api/cmd/kubermatic-api/swagger.json
@@ -42466,6 +42466,11 @@
       "description": "VirtualMachineInstanctype represents a KubeVirt VirtualMachineInstanctype",
       "type": "object",
       "properties": {
+        "kind": {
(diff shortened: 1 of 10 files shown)


KhizerRehan:
@soer3n  Incase /backport-needed kindly, take care to prev versions

PR #8117, opened 2026-06-11
Admin settings not loaded after authentication

ahmadhamzh:
The `adminSettings()` getter always returns the default admin settings because the authentication check happens before the `/status` endpoint responds. As a result, `this._auth.authenticated()` is `false` on the initial call, causing `_adminSettingsWatch$` to be initialized with the default settings.

Since both the WebSocket and HTTP approaches for fetching the settings are already protected by authentication, I don't think this additional check is necessary.

**What type of PR is this?**
/kind bug

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

**Test issue**:
```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/core/services/settings.ts
+++ b/modules/web/src/app/core/services/settings.ts
@@ -51,9 +51,9 @@ export class SettingsService {
   get adminSettings(): Observable<AdminSettings> {
     if (!this._adminSettingsWatch$) {
       this._adminSettingsWatch$ = iif(
-        () => this._auth.authenticated(),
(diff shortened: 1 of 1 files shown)


ahmadhamzh:
/retitle WIP Admin settings not loaded after authentication

ahmadhamzh:
/retitle Admin settings not loaded after authentication

KhizerRehan:
/approve
/lgtm

Do we need to backport also for prev releases I guess Yes 

/cc @ahmadhamzh

ahmadhamzh:
> Do we need to backport also for prev releases I guess Yes

@KhizerRehan the issue happen cause of the new authorization flow and that is not in any release yet so only main is enough

PR #8135, opened 2026-06-22
Mount ca-bundle into web terminal pod to fix TLS error

### LINKED ISSUE #7993: Web Terminal kubectl fails with TLS verification error (Unknown Authority)
### What happened
The KKP user cluster web terminal is unable to execute `kubectl` commands. The terminal fails during API discovery because it cannot verify the TLS certificate of the OIDC provider (Keycloak)

```bash
E0330 10:53:02.146610     109 memcache.go:265] "Unhandled Error" err="couldn't get current server API group list: Get \"https://ipfkqwqlqv.xx-seed.kkp.xxx.net:6443/api?timeout=32s\": Get \"https://keycloak.xxx.xxx.net/realms/kkp1/.well-known/openid-configuration\": tls: failed to verify certificate: x509: certificate signed by unknown authority"
```

### Expected behavior
`kubectl` commands should work from the user cluster without any issue.  

### How to reproduce

1. Log in to the KKP Dashboard.
2. Open the Web Terminal for a user cluster that uses an OIDC provider (Keycloak) with a custom/internal CA.
3. Attempt to run any `kubectl` command (e.g., `kubectl get pods`).
4. Observe the `x509: certificate signed by unknown authority` error in the terminal output.

### Environment
- UI Version: 2.29.4
- API Version:  2.29.4
- Domain:
- Others:

### Additional information
Root cause analysis: 
The `ca-bundle` ConfigMap containing the custom CA for Keycloak exists on the user cluster, but it is not mounted into the web terminal's pod. Consequently, `kubectl` lacks the necessary trust chain to validate the identity provider's certificate during the authentication flow.

ahmadhamzh:
Mount the `ca-bundle` ConfigMap, and set `SSL_CERT_FILE` so `kubectl` trusts it.

Fixes #7993 

**What type of PR is this?**
/kind bug

```release-note
Fixed the web terminal failing with a TLS "unknown authority" error
```

**Documentation**:
```documentation
NONE
```

**Test issue**:
```test-issue
NONE
```

### DIFF
--- a/modules/api/pkg/handler/websocket/terminal.go
+++ b/modules/api/pkg/handler/websocket/terminal.go
@@ -74,6 +74,10 @@ const (
 
 	webTerminalImage                   = resources.WEBTerminalImage
 	webTerminalContainerKubeconfigPath = "/etc/kubernetes/kubeconfig/kubeconfig"
+
+	webTerminalCABundleVolume = "ca-bundle"
+	webTerminalCABundleDir    = "/etc/kubermatic/certs"
+	webTerminalCABundlePath   = webTerminalCABundleDir + "/" + resources.CABundleConfigMapKey
 )
 
 type TerminalConnStatus string
@@ -638,6 +642,10 @@ func genWebTerminalPod(userAppName, userEmailID string, options *kubermaticv1.We
 			Name:  "PS1",
 			Value: "\\s-\\v \\w \\$ ",
 		},
+		{
+			Name:  "SSL_CERT_FILE",
+			Value: webTerminalCABundlePath,
+		},
 	}
 
 	if options != nil && options.AdditionalEnvironmentVariables != nil {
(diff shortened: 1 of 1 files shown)


modules/api/pkg/handler/websocket/terminal.go
```diff
@@ -817,6 +842,11 @@ func getVolumeMounts() []corev1.VolumeMount {
 			Name:      webTerminalConfigVolume,
 			MountPath: "/etc/config",
 		},
+		{
+			Name:      webTerminalCABundleVolume,
+			MountPath: webTerminalCABundleDir,
+			ReadOnly:  true,
+		},
 	}
 }
```
buraksekili, line 851: one thing, i'd like to get your thoughts on this `Optional` configuration. i guess, we can drop it on the ca-bundle volume. When `SSL_CERT_FILE` is set, Go loads roots from that file instead of the default cert files, so if the ConfigMap is ever missing the file won't exist but the env still points at it, it breaks TLS. The ConfigMap is reconciled unconditionally, so just make the volume required and fail fast.

some references that i used for that SSL_CERT_FILE behaviour of the go
- https://pkg.go.dev/crypto/x509
- https://github.com/golang/go/issues/79496

let me know your thoughts!

buraksekili (commented):
overall lgtm, thank you @ahmadhamzh. i just have a quick suggestion/question

modules/api/pkg/handler/websocket/terminal.go
ahmadhamzh: Makes sense, made it required.
a missing `ca-bundle` means something's wrong — better to stay pending and recover when it's created than start with no CA bundle.

buraksekili (approved):
/lgtm
/approve

👌

KhizerRehan:
@ahmadhamzh @buraksekili 
I guess we need to cherry-pick this PR too?

PR #8161, opened 2026-07-01
Add option to disable Audit Webhook Backend per datacenter

### LINKED ISSUE #14862: Add option to disable Audit Webhook Backend

Currently when creating a cluster, as soon as the "Audit logging" option is enabled, the "Audit Webhook Backend" option is shown to the user.
As we enforce the "Audit logging" from the seed, the "Audit Webhook Backend" option is always visible to users, but as they aren't able to create a secret in the KKP cluster, there is no possibility for them to configure this webhook correctly.
As an admin (or platform admin) I would like to have a configuration in the seed, similar to `enforceAuditLogging` which allows us to disable the Audit webhook for our users. 

We already tried to configure an empty webhook config through `enforcedAuditWebhookSettings` in the seed, but this just results in failures during cluster creation.

### Solution details

- Add a configuration in the seed to disable Audit Webhook. 
- Disable the option in UI when this seed configuration is set.

ahmadhamzh:
Adds a global `KubermaticSetting.spec.disabledAuditWebhookBackendDCs` (a list of datacenter names) that lets admins **hide and disable** the `Audit Webhook Backend` option for the listed datacenters.

Fixes https://github.com/kubermatic/kubermatic/issues/14862

**What type of PR is this?**
/kind feature

```release-note
Add a global setting to disable the Audit Webhook Backend option for specific datacenters.
```

**Documentation**:
```documentation
TBD
```

**Test issue**:

```test-issue
https://github.com/kubermatic/dashboard/issues/8257
```


modules/api/pkg/handler/common/cluster.go
```diff
@@ -573,6 +609,15 @@ func PatchEndpoint(
...
 	newInternalCluster.Spec.AuditLogging = patchedCluster.Spec.AuditLogging
+
+	globalSettings, err := settingsProvider.GetGlobalSettings(ctx)
+	if err != nil {
+		return nil, common.KubernetesErrorToHTTPError(err)
+	}
+
+	if err := validateAuditWebhookBackendAllowed(globalSettings.Spec.DisabledAuditWebhookBackendDCs, oldInternalCluster.Spec.Cloud.DatacenterName, oldInternalCluster.Spec.AuditLogging, newInternalCluster.Spec.AuditLogging); err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
ahmadhamzh: the datacenter is immutable.
KKP's validation webhook rejects any change to spec.cloud.dc, so a patch can never move a cluster into a different DC. The old and patched DC names are always identical, making the old name correct.

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -750,6 +753,12 @@ export class EditClusterComponent implements OnInit, OnDestroy {
     }
   }
 
+  private _handleAuditWebhookBackendVisibility(): void {
```
KhizerRehan, line 756: Confirmation:

Case-1

- I Created e.g Openstack Cluster
   - Audit Loggin Option was not enabled


Later I added entry of DC in `disabledAuditWebhookBackendDC` field
```
[
  "syseleven-dbl1"
]
```

and i was trying to edit cluster again which was already created before and NOW on edit i am not able to see is it expected to be like this?


Case-2:

Saved Cluster template with "Audit Logging" option enabled and `disabledAuditWebhookBackendDC` ddint' contained DC so 'Audit Webhook Backend' was enabled too but once template was created and DC was added `disabledAuditWebhookBackendDC`

- Edit of ClusterTemplate hide  'Audit Webhook Backend'  so I assume this is also expected due to this valid and it is hiding due to this `_handleAuditWebhookBackendVisibility` logic

modules/api/pkg/test/e2e/utils/apiclient/models/global_settings.go
```diff
@@ -35,6 +35,10 @@ type GlobalSettings struct {
 	// DisableChangelogPopup disables the changelog popup in KKP dashboard.
 	DisableChangelogPopup bool `json:"disableChangelogPopup,omitempty"`
 
+	// DisabledAuditWebhookBackendDCs is the list of datacenters for which the Audit Webhook Backend
+	// option is disabled in the dashboard.
+	DisabledAuditWebhookBackendDCs []string `json:"disabledAuditWebhookBackendDCs"`
+
```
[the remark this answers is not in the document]
ahmadhamzh: This is expected 
the generator never adds `omitempty` to array fields (see `StaticLabels`, same pattern), regardless of the server-side tag.
Matching it by hand would breaks `verify-api-client.sh`, so reverted.

modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts
```diff
@@ -750,6 +753,12 @@ export class EditClusterComponent implements OnInit, OnDestroy {
...
+    const datacenterName = this.datacenter?.metadata?.name;
+    this.isAuditWebhookBackendHidden =
+      !!datacenterName && !!this._settings?.disabledAuditWebhookBackendDCs?.includes(datacenterName);
+  }
```
KhizerRehan, line 760: Clarification:

Implementation done for [webhook visibility ](https://github.com/kubermatic/dashboard/pull/8161/changes#diff-39419deb96986de75a95c901fcfd3c1596b31dec835414cc8cf2f6c34fb90e08R1105). I see we are checking `enforcedAuditWebhookSettings` but incase of edit cluser 

- what if it is enforced but `hidden` or entry already exists in `disabledAuditWebhookBackendDCs` isn't should always shown incase of `enforced`? 

In Edit case when enforced: it is sending `null`  [here](https://github.com/kubermatic/dashboard/blob/main/modules/web/src/app/cluster/details/cluster/edit-cluster/component.ts#L565-L574)

or am I missing something?

modules/web/src/app/settings/admin/defaults/component.ts
```diff
@@ -253,6 +261,20 @@ export class DefaultsComponent implements OnInit, OnDestroy {
...
+    this.settings.disabledAuditWebhookBackendDCs = val;
+    this._disabledAuditWebhookBackendDCsDirty = true;
+  }
+
+  onDisabledAuditWebhookBackendDCsPanelToggle(isOpen: boolean): void {
+    if (!isOpen && this._disabledAuditWebhookBackendDCsDirty) {
+      this._disabledAuditWebhookBackendDCsDirty = false;
+      this.onSettingsChange();
+    }
+  }
```
KhizerRehan, line 266: Is there a specific reason to have a separate flag / dropdown state just to make a single PATCH call when the dropdown is explicitly closed?

- IMO, we can remove openedChange and the extra flag. WDYT? Whether the user selects 1 or 5 values, it can simply send n PATCH requests (atomic API calls).
- I don't think this endpoint performs any heavy computation, so multiple PATCH calls for multiple select/unselect actions should be fine.

From a UX perspective, I've often selected multiple values and expected them to be saved immediately. If I navigate away before explicitly closing the dropdown, those changes are lost, which feels a bit unintuitive.

WDYT? If you think the current approach is the right one, feel free to ignore this comment.

**Suggesion**
If there's no specific need, drop the dirty flag, the `onDisabledAuditWebhookBackendDCsPanelToggle` handler, and the `(openedChange)` binding:

```ts
onDisabledAuditWebhookBackendDCsSelect(val: string[]): void {
  this.settings.disabledAuditWebhookBackendDCs = val;
  this.onSettingsChange();
}
```

modules/web/src/app/wizard/step/cluster/template.html
```diff
@@ -483,7 +483,7 @@ <h4>IPv6</h4>
                  matTooltip="Audit Logging is enforced by your admin in the chosen datacenter."></i>
```
KhizerRehan, line 483: Adding few things

Thoughts on adding `i`? for consistency?

Image

modules/web/src/app/wizard/step/cluster/template.html
```diff
@@ -483,7 +483,7 @@ <h4>IPv6</h4>
...
               }
             </div>
-            @if (!!controlValue(Controls.AuditLogging)) {
+            @if (!!controlValue(Controls.AuditLogging) && !isAuditWebhookBackendHidden) {
             <div fxLayoutAlign=" center">
```
KhizerRehan, line 487: Image

We should do like this 

Image

wdyt? OR maybe some tooltip that can replaced that "Audit Webhook Backend" is disabled or hidden 

As is there admin setting but in wizard or edit dialog (we are hiding completely) which shows that  either there is no such thing exists (but Visually Greyed or Disable) may gave some indication something not enabled thoughts as shown above?

```
Hide the Audit Webhook Backend option in the cluster wizard and edit-cluster dialog for the selected datacenters. Users in these datacenters will not be able to configure an audit webhook backend.
```

Image

KhizerRehan (commented):
PTAL: Minor Feedback

modules/web/src/app/wizard/step/cluster/template.html
ahmadhamzh: I noticed that most of the i info icons just say "enable {feature} for this cluster", which doesn't really add any information.
So maybe in a separate PR we can either remove the unnecessary `i` icons or update the `tooltip` messages. I'd prefer to keep that out of this PR to stay scoped to the feature.

PR #8163, opened 2026-07-02
Discover OpenStack images per selected OS via new Admin Setting

### LINKED ISSUE #3024: Pre-fill Openstack image textbox with latest available image for the selected OS

This is a feature that we have in our fork and would like to bring upstream.

For the Openstack image field in the cluster creation wizard and the node deployment editor, the user has to look up and enter the desired image name manually. What we want to do instead is fill this field with the latest image matching the selected OS version.

image

For an existing node deployment we also notify the user when a newer image is available for the selected OS:

image

The way this works is that we provide a simple additional API endpoint (actually two, one with and one without credentials) for getting the list of all Openstack images, and then the dashboard looks at (string-valued) [image properties](https://docs.openstack.org/glance/latest/admin/useful-image-properties.html) to determine which OS an image belongs to, and the image creation time to see which one is the newest. If no image is found for the selected OS, we fall back to the images from the node datacenter spec (`dc.spec.openstack.images`). There's also a regexp that we match against the image name the user set (this is still a normal text field, so the user can still manually enter any image name) so that if they uploaded a custom image named something like "my image" and are using that, we just disable the whole update check functionality. We also disable it if Flatcar was selected and automatic updates enabled.

To bring this upstream, I'd propose we extend the UI configuration file a little so that you can specify the above values -- the image name regexp, and possibly the Openstack image property name and values to check to see what OS an image belongs to (or do we actually want these settings in the node datacenter spec instead?). There should also be a boolean flag to just disable the whole feature altogether.

KhizerRehan:
This PR adds OpenStack image discovery to the dashboard, listing Glance images filtered by the selected operating system in the machine deployment form when the new enableImageDiscovery admin setting is enabled.

Fixes #3024 

#### Admin Settings Flag: [PR Link](https://github.com/kubermatic/kubermatic/pull/16082)

Screenshot 2026-07-02 at 4 53 34 PM

**When (Admin Flag is False)**
 
FYI:
- Dropdown with 1 Value 
- Selects Default value as per defined in Seed Object / Datacenter <PROVIDER> images

Screenshot 2026-07-02 at 4 50 58 PM

**When (Admin Flag is `true`)**

Screenshot 2026-07-02 at 4 52 44 PM

**Flatcar**

Screenshot 2026-07-02 at 4 53 04 PM

**Rhel**

Screenshot 2026-07-02 at 4 53 16 PM

**What type of PR is this?**
/kind feature

```release-note
OpenStack: discover images for the selected OS via the new "enableImageDiscovery" admin setting (disabled by default)
```

**Documentation**:

```documentation
https://github.com/kubermatic/docs/pull/2235
```

**Test issue**:

```test-issue
TBD
```


KhizerRehan:
Hi @ahmadhamzh 

Method Invoked (Regarding Point Raised in Internal Sync) 

1. With credentials — GET /api/v2/projects/{project_id}/providers/openstack/images → listProjectOpenstackImages() 
2. Without credentials — GET /api/v2/projects/{project_id}/clusters/{cluster_id}/providers/openstack/images → listOpenstackImagesNoCredentials()

**Add**

Screenshot 2026-07-06 at 2 49 35 PM


**Edit**
Screenshot 2026-07-06 at 2 49 22 PM

You can review this PR and incase of any concerns let me know:

thx

modules/api/pkg/handler/v2/routes_v2.go
```diff
@@ -4547,6 +4555,31 @@ func (r Routing) listOpenstackAvailabilityZonesNoCredentials() http.Handler {
...
+//	  default: errorResponse
+//	  200: []OpenstackImage
+func (r Routing) listOpenstackImagesNoCredentials() http.Handler {
+	return httptransport.NewServer(
+		endpoint.Chain(
+			middleware.TokenVerifier(r.tokenVerifiers, r.userProvider),
+			middleware.UserSaver(r.userProvider),
+			middleware.SetClusterProvider(r.clusterProviderGetter, r.seedsGetter),
+			middleware.SetPrivilegedClusterProvider(r.clusterProviderGetter, r.seedsGetter),
+		)(provider.OpenstackImageWithClusterCredentialsEndpoint(r.projectProvider, r.privilegedProjectProvider, r.seedsGetter,
```
ahmadhamzh: i think maybe the name is confusing, shouldn't be `OpenstackImagesNoCredentialsEndpoint` ? WDYT?

modules/web/src/app/core/services/node-data/provider/openstack.ts
```diff
@@ -151,6 +156,70 @@ export class NodeDataOpenstackProvider {
...
+    switch (this._nodeDataService.mode) {
+      case NodeDataMode.Wizard:
+        // Emit the current cluster immediately (clusterChanges is an EventEmitter
+        // with no replay) so re-fetching on OS change does not wait for the next
+        // credential change, then keep reacting to later credential changes.
+        return merge(of(this._clusterSpecService.cluster), this._clusterSpecService.clusterChanges)
+          .pipe(
+            filter(
+              cluster =>
+                !!cluster?.spec?.cloud?.openstack && this._clusterSpecService.provider === NodeProvider.OPENSTACK
```
ahmadhamzh: i think this is enough `this._clusterSpecService.provider` already check for `cluster?.spec?.cloud?.openstack`

```suggestion
                this._clusterSpecService.provider === NodeProvider.OPENSTACK
```

modules/web/src/app/node-data/basic/provider/openstack/template.html
```diff
@@ -62,15 +62,18 @@
...
-           kmValueChangedIndicator>
-  </mat-form-field>
+  <km-combobox #imageCombobox
+               [required]="true"
+               [grouped]="false"
+               [selected]="selectedImage"
+               [options]="images"
+               [formControlName]="Controls.Image"
+               (changed)="onImageChange($event)"
+               [label]="imageLabel"
```
ahmadhamzh, line 73: in order to disable combobox component i think it's better to use `isDisabled` 
```suggestion
               [label]="imageLabel"
               [isDisabled]="form.get(Controls.Image).disabled"
```

ahmadhamzh (commented):
few suggestions

ahmadhamzh (approved):
/approve

buraksekili (approved):
/lgtm

PR #8167, opened 2026-07-06
switch dashboard api-e2e cluster-create test from Hetzner to AWS

buraksekili:

The `pre-dashboard-api-e2e` prow job is failing on every dashboard PR (8161, 8163, 8166, …) with the same error: `TestCreateUpdateHetznerCluster` creates a Hetzner cluster whose control plane becomes healthy, but the worker node never reaches `Ready` within the 15m timeout (`hetzner_test.go:145`, `NodeDeployment has not reached 1 ready replicas`). The failure is environmental to the shared Hetzner e2e setup, not a regression in any individual PR.

This switches the cluster-create e2e test to AWS so the job can pass, while keeping the Hetzner test code on disk for later re-enablement.

Changes:
- `hetzner_test.go`: retag `//go:build create` → `//go:build hetzner_e2e` to disable both Hetzner tests in CI (file + helpers preserved; mirrors the existing `kubevirt_test.go` exclusion convention).
- `aws_test.go` (new, tag `create`): `TestCreateUpdateAWSCluster` + `TestDeleteAWSClusterBeforeIsUp` targeting `dc=kubermatic`, `location=aws-eu-central-1a`, `credential=e2e-aws`, reusing the shared `createCluster`/`testCluster`/`cleanupProject` helpers.
- `utils/client.go`: add a preset-based `CreateAWSCluster(projectID, dc, name, credential, version, location, replicas)`; remove the prior unused inline-credential `CreateAWSCluster` (no callers) it conflicts with.
- `run-api-e2e.sh`: add an `e2e-aws` Preset heredoc reading `AWS_E2E_TESTS_KEY_ID`/`AWS_E2E_TESTS_SECRET` (e2e-hetzner kept so `credentials_test.go` still passes).
- `.prow/api.yaml`: attach `preset-aws-e2e-kkp: "true"` to the `pre-dashboard-api-e2e` job.

AWS credentials (`preset-aws-e2e-kkp` / secret `e2e-aws-kkp`) and the `aws-eu-central-1a` datacenter already exist, so no new secrets or seed changes are required.

Fixes #

**What type of PR is this?**
/kind failing-test

WIP — the local build/vet/gofmt/boilerplate checks pass, but this needs a real CI run to confirm the AWS cluster comes up in `aws-eu-central-1a`. Re-enabling Hetzner later is a one-line build-tag flip back to `create`.

The larger `client.go` diff is because the pre-existing `CreateAWSCluster` (12-arg inline-credential signature, no callers) had to be removed to avoid a duplicate-method compile error; the new method matches the Hetzner preset pattern.

```release-note
NONE
```

**Documentation**:
```documentation
NONE
```

```test-issue
NONE
```

### DIFF
--- a/modules/api/pkg/test/e2e/utils/client.go
+++ b/modules/api/pkg/test/e2e/utils/client.go
@@ -432,6 +432,69 @@ func (r *TestClient) ListCredentials(providerName, datacenter string) ([]string,
 	return names, nil
 }
 
+// CreateAWSCluster creates cluster for AWS provider. Credentials are injected
+// via the named preset, so the cloud spec is left empty.
+func (r *TestClient) CreateAWSCluster(projectID, dc, name, credential, version, location string, replicas int32) (*apiv1.Cluster, error) {
+	_, err := semverlib.NewVersion(version)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse version %s: %w", version, err)
+	}
+
+	clusterSpec := &models.CreateClusterSpec{}
+	clusterSpec.Cluster = &models.Cluster{
+		Type:       "kubernetes",
+		Name:       name,
+		Credential: credential,
+		Spec: &models.ClusterSpec{
+			Cloud: &models.CloudSpec{
+				DatacenterName: location,
+				Aws:            &models.AWSCloudSpec{},
+			},
+			Version: models.Semver(version),
+		},
+	}
+
+	if replicas > 0 {
+		instanceType := "t3.small"
+		volumeSize := int32(25)
+		volumeType := "gp2"
+
+		clusterSpec.NodeDeployment = &models.NodeDeployment{
+			Spec: &models.NodeDeploymentSpec{
+				Replicas: &replicas,
+				Template: &models.NodeSpec{
+					Cloud: &models.NodeCloudSpec{
+						Aws: &models.AWSNodeSpec{
+							InstanceType: &instanceType,
+							VolumeSize:   &volumeSize,
+							VolumeType:   &volumeType,
+						},
+					},
+					OperatingSystem: &models.OperatingSystemSpec{
+						Ubuntu: &models.UbuntuSpec{
+							DistUpgradeOnBoot: false,
+						},
+					},
+				},
+			},
+		}
+	}
+
+	r.test.Logf("Creating AWS cluster %q (%s, %d nodes)...", name, version, replicas)
+
+	params := &project.CreateClusterParams{ProjectID: projectID, DC: dc, Body: clusterSpec}
+	SetupParams(r.test, params, 1*time.Second, 3*time.Minute)
+
+	clusterResponse, err := r.client.Project.CreateCluster(params, r.bearerToken)
+	if err != nil {
+		return nil, err
+	}
+
+	r.test.Log("Cluster created successfully.")
+
+	return convertCluster(clusterResponse.Payload)
+}
+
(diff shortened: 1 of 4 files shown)


modules/api/pkg/test/e2e/utils/client.go
```diff
@@ -432,54 +432,42 @@ func (r *TestClient) ListCredentials(providerName, datacenter string) ([]string,
...
-		volumeType := "standard"
+		volumeType := "gp2"
 
 		clusterSpec.NodeDeployment = &models.NodeDeployment{
 			Spec: &models.NodeDeploymentSpec{
 				Replicas: &replicas,
 				Template: &models.NodeSpec{
 					Cloud: &models.NodeCloudSpec{
 						Aws: &models.AWSNodeSpec{
-							AvailabilityZone: availabilityZone,
```
KhizerRehan, line 479: Do we need to add `AvailabilityZone: "eu-central-1a" ? or will it picked by default. Although CI pipeline is passing just confirming!

modules/api/pkg/test/e2e/api/hetzner_test.go
```diff
@@ -1,4 +1,4 @@
-//go:build create
+//go:build hetzner_e2e
```
KhizerRehan, line 1: I was checking other spec e.g modules/api/pkg/test/e2e/api/kubevirt_test.go

I see we are doing same so it means Hetzner tests are being turned off in CI right now?

Is it something due to hetzner environment that is causing issue till that we have skipped or are we going to fix in future?

I was also checking we have tag like `//go:build e2e` like gcp wondering why can't we do `//go:build e2e` like it will still fail?

KhizerRehan (commented):
PTAL: Minor Clarificaition

modules/api/pkg/test/e2e/api/hetzner_test.go
buraksekili: yes, the Hetzner cluster-create tests are skipped in CI now. run-api-e2e.sh only runs the create, e2e, and logout tag sets, and hetzner_e2e isn't among them, same way kubevirt_test.go (`//go:build kubevirt`) is already skipped.

modules/api/pkg/test/e2e/utils/client.go
buraksekili: as per my understanding,  not needed as it gets defaulted downstream. the dashboard API just passes the AZ through (it does no AWS validation on this path). because the cluster is created with an initial NodeDeployment, the initial-machinedeployment-controller runs `CompleteAWSProviderSpec` on it, which defaults an empty region from the datacenter (`spec.aws.region: eu-central-1`) and an empty AZ to <region>a  ->> eu-central-1a. our aws-eu-central-1a datacenter in `hack/ci/testdata/seed.yaml` only sets region, so the node lands in eu-central-1a automatically. CI passing matches that.

but if that does not work, we can re-add it of course. do you have any concerns on this?

KhizerRehan: No! Just clarification.
thank you!

modules/api/pkg/test/e2e/api/hetzner_test.go
KhizerRehan: I was wondering if we've had a chance to investigate why the node isn't joining.

Could it be related to a deprecated Hetzner instance type? If that's the case, would it be something we could fix instead of moving the tests to AWS?

any thoughts or suggestios?

cc: @buraksekili

PR #8182, opened 2026-07-13
Raise hardcoded KubeVirt node defaults

### LINKED ISSUE #8179: Raise the hardcoded KubeVirt node defaults

When a user creates a KubeVirt machine deployment and selects the "Custom Configuration" option for the node size, the dashboard pre-fills the form with default values. These defaults are currently too small to run a useful workload.

The current defaults are hardcoded in the dashboard:

- Disk: 10G
- CPU: 2
- Memory: 2000 MB

This results in worker nodes that are too small for most real workloads. As a first step toward a proper solution, we want to raise these hardcoded defaults to sensible baseline values.

**details**

Raise the KubeVirt node defaults in the dashboard to the following values, chosen as practical minimums for a usable Kubernetes worker node:

- Disk: 25G
- CPU: 2
- Memory: 4G

The 25G disk size also matches the suggested minimum storage in the Ubuntu Server system requirements guide (the hard minimum is lower, but 25G is a reasonable default for a worker node that needs room for container images and ephemeral storage).

These defaults apply only when the user is on the "Custom Configuration" path. When an instance type is selected, the CPU and memory come from the instance type, and the dashboard already avoids overriding them.

The hardcoded values live in the dashboard repository at:

- [modules/web/src/app/shared/entity/node.ts](https://github.com/kubermatic/dashboard/blob/a6320f16ff331d0bf51bf28ae49a239198887be1/modules/web/src/app/shared/entity/node.ts#L407), the KubeVirt case sets `primaryDiskSize: '10'`.
- [modules/web/src/app/node-data/basic/provider/kubevirt/component.ts](https://github.com/kubermatic/dashboard/blob/a6320f16ff331d0bf51bf28ae49a239198887be1/modules/web/src/app/node-data/basic/provider/kubevirt/component.ts#L158-L159):
  - the form control for primary disk size, built with the literal `'10G'` https://github.com/kubermatic/dashboard/blob/a6320f16ff331d0bf51bf28ae49a239198887be1/modules/web/src/app/node-data/basic/provider/kubevirt/component.ts#L212
  - the same constants are re-applied when the user switches into custom-configuration mode https://github.com/kubermatic/dashboard/blob/a6320f16ff331d0bf51bf28ae49a239198887be1/modules/web/src/app/node-data/basic/provider/kubevirt/component.ts#L351-L353

All of these should be updated to the new baseline values.

### Notes

- The memory field today is an integer in MB.
- This change is dashboard only. It can ship as part of a patch release.
- This is the first half of the larger request. The second half is tracked separately and adds the ability for an admin to configure these defaults per datacenter, after which the values defined here become the fallback when no datacenter default is set.

buraksekili:

The KubeVirt worker-node defaults pre-filled in the dashboard's Custom Configuration path are too small for real workloads (disk 10G, CPU 2, memory 2000 MB). This raises them to sensible baselines:

- Disk: 10G -> 25G (matches the suggested minimum in the Ubuntu Server system requirements guide)
- Memory: 2000 -> 4000 MB (~4G, in the decimal-MB unit the field emits)
- CPU: 2 (unchanged)

Touched locations:
- `modules/web/src/app/shared/entity/node.ts` — default `KubeVirtNodeSpec.primaryDiskSize`: `'10'` -> `'25'`.
- `modules/web/src/app/node-data/basic/provider/kubevirt/component.ts` — `_defaultMemory`: `2000` -> `4000`; primary disk form control: `'10G'` -> `'25G'`. The constant is also consumed by the custom-mode re-apply and the existing-node load fallback, both of which pick up the new value by reference.

These defaults apply only on the Custom Configuration path. Instance-type selection continues to take CPU/memory from the instance type and is unaffected.

Fixes #8179

**What type of PR is this?**:
/kind feature

- Memory is an integer in MB; the build path appends a capital `M` (`${memory}M`), so the literal must stay a number. `4000M` is the decimal-megabyte interpretation of the issue's "4G" (~3.73 GiB). machine-controller treats memory as an opaque k8s quantity string (`resource.ParseQuantity`) and enforces no minimum, so 4000 is safe.
- This is the first half of #8179. The second half (admin-configurable per-datacenter defaults, with these values becoming the fallback) is tracked separately.
- No backend, API, or machine-controller changes. No existing unit test asserts these literals.

```release-note
The KubeVirt node defaults shown in the Custom Configuration path have been raised: primary disk 10G -> 25G, memory 2000 MB -> 4000 MB. CPU default is unchanged at 2.
```

**Documentation**:
NONE

### DIFF
--- a/modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
+++ b/modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
@@ -156,7 +156,7 @@ export class KubeVirtBasicNodeDataComponent
   selectedConfigurationMode = ConfigurationMode.CustomResources;
   private readonly _instanceTypeIDSeparator = ':';
   private readonly _defaultCPUs = 2;
-  private readonly _defaultMemory = 2000;
+  private readonly _defaultMemory = 4000;
   private readonly _initialData = _.cloneDeep(this._nodeDataService.nodeData.spec.cloud.kubevirt);
   private _quotaCalculationService: QuotaCalculationService;
   private _initialQuotaCalculationPayload: ResourceQuotaCalculationPayload;
@@ -209,7 +209,7 @@ export class KubeVirtBasicNodeDataComponent
       [Controls.Memory]: this._builder.control(this._defaultMemory, Validators.required),
       [Controls.PrimaryDiskOSImage]: this._builder.control('', Validators.required),
       [Controls.PrimaryDiskStorageClassName]: this._builder.control('', Validators.required),
-      [Controls.PrimaryDiskSize]: this._builder.control('10G', Validators.required),
+      [Controls.PrimaryDiskSize]: this._builder.control('25G', Validators.required),
       [Controls.NodeAffinityPreset]: this._builder.control(''),
       [Controls.NodeAffinityPresetKey]: this._builder.control('', Validators.required),
       [Controls.NodeAffinityPresetValues]: this._builder.control(''),
--- a/modules/web/src/app/shared/entity/node.ts
+++ b/modules/web/src/app/shared/entity/node.ts
@@ -404,7 +404,7 @@ export function getDefaultNodeProviderSpec(provider: string): object {
       } as AnexiaNodeSpec;
     case NodeProvider.KUBEVIRT:
       return {
-        primaryDiskSize: '10',
+        primaryDiskSize: '25',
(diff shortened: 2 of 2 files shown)


KhizerRehan:
/approve
/lgtm
/docs-none
/test-issue-none


@buraksekili incase this PR requires back porting please feel free to cherry-pick this PR once merged.
Thank you

PR #8052, opened 2026-05-04
Angular Upgrade v21.x

KhizerRehan:
**What this PR does / why we need it**:
 This PR upgrades the KKP Dashboard Angular frontend from v20.3.x to v21.2.x version.

**Which issue(s) this PR fixes**:

Fixes #7951 

/kind chore

**What type of PR is this?**

**Special notes for your reviewer**:
- Updated Angular packages to version 21.x in package.json.
- Replaced ngx-clipboard with Angular CDK's clipboard functionality in multiple components.

### Removed Package

- ngx-clipboard": "16.0.0" in favour of built support for copy to clipboard by [angular CDK ](https://v11.material.angular.dev/cdk/clipboard/api)

There are few packages that are no longer maintainer and therefore "PEER" depedency issue as supported version are not available however older verison of packages still works but adding overrides for few reasons:

Two packages in the project are overridden to force Angular 21 peer dependencies:

| Package | Version | Status | Override Reason |
|---------|---------|--------|-----------------|
| `@angular/flex-layout` | `15.0.0-beta.42` | Archived (2022) | Declares Angular 15 peers; last release pre-v21 |
| `@swimlane/ngx-charts` | `23.1.0` | Slow maintenance | No Angular 21 release available |

**What the override does:**
- Silences npm peer-dependency mismatch warnings

```bash
$ = version variable in npm overrides.                                              
                                                                                    
  $@angular/core means "use the same version as root-level @angular/core".            
                                                                                      
Example:                                                                                                                                                                  

  Root package.json:                                                                  
  "@angular/core": "21.2.9"                                                           

  Override:                                                                         
  "@angular/flex-layout": {                                 
    "@angular/core": "$@angular/core"
  }    
```

FYI: Even without `overrrides` of these packages these can work and bcz of this [line](https://github.com/kubermatic/dashboard/blob/main/modules/web/.npmrc#L3) otherwise you can probably see following error and currently those are SILENCED

```
npm error code ERESOLVE
npm error ERESOLVE could not resolve
npm error
npm error While resolving: @swimlane/ngx-charts@23.1.0
...
...
```

### Useful to Add Overrides property with $

  1. Auto-tracking: if you upgrade Angular to v22, overrides automatically use v22. No manual edit needed.
  2. Single source of truth: one place (@angular/core: 21.2.9 in root) controls all downstream versions.
  3. Alternative without $: "@angular/core": "21.2.9"  // hardcoded, stale when you upgrade then in future updated have to manually update.
[...]

PR #8189, opened 2026-07-17
Fix project selector dropdown not closing

### LINKED ISSUE #8192: Project selector dropdown in top navbar does not close
### What happened

Opening the project selector dropdown in the top navigation bar leaves it stuck open. Clicking the trigger again, clicking outside the panel, or picking a project does not dismiss the panel — it closes and instantly reopens in the same tick, so visually it never closes. Only a route change or page reload clears it.

**Root cause:** the project selector wraps its `mat-select` in a list-item that has an unconditional "open the dropdown" click handler; since the Angular/Material 21 upgrade the dropdown overlay and its backdrop are rendered *inline in the DOM inside that wrapper* (native Popover API) instead of in the body-level overlay container, so the dismiss click bubbles back into the wrapper's handler and reopens the panel immediately.

"Angular Upgrade v21.x"** (merged 2026-05-11), which moved `@angular/material` and `@angular/cdk` from 20.2.12 to 21.2.x. Material 21 opts `mat-select` into CDK's new inline-popover overlay mode; CDK then inserts the overlay host directly after the select trigger and prepends the backdrop inside it. In Material 20 the overlay lived in the global overlay container, outside the navbar DOM, so dismiss clicks never reached the wrapper.

### Expected behavior

The project selector dropdown closes when the user clicks the trigger a second time, clicks anywhere outside the panel, presses `Esc`, or selects a project.

### How to reproduce

1. Log in to the dashboard (any user with at least one project).
2. Click the project selector in the top navigation bar  the dropdown opens.
3. Click the project selector trigger again, or click anywhere outside the panel.
4. Observe the dropdown stays open (it closes and reopens within the same click, so it never appears dismissed).
5. Select a project from the list — the project switches, but the panel remains open.

### Environment

- UI Version: 2.30.0 (main, since #8052 merged 2026-05-11)
- API Version: N/A — frontend-only

### Current workaround

Navigate to another view or reload the page to clear the stuck panel. Project switching itself still works — the selection is applied, only the panel fails to dismiss.

### Related issues

- Introduced by: https://github.com/kubermatic/dashboard/pull/8052 (Angular Upgrade v21.x)

ahmadhamzh:
Fixes the project selector dropdown in the top navbar getting stuck open — the mat-list-item wrapper's unconditional open() call fought the panel's own close behavior.
Removes the wrapper so mat-select handles its own open/close natively, and moves layout/hover/arrow styling onto it directly.

Fixes: #8192 

**What type of PR is this?**
/kind bug

```release-note
Fixed the project selector dropdown in the top navigation bar not closing after being opened.
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/core/components/navigation/project/component.ts
+++ b/modules/web/src/app/core/components/navigation/project/component.ts
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import {Component, Input, OnDestroy, OnInit} from '@angular/core';
-import {MatSelect, MatSelectChange} from '@angular/material/select';
+import {Component, Input, OnDestroy, OnInit, ViewEncapsulation} from '@angular/core';
+import {MatSelectChange} from '@angular/material/select';
 import {ProjectService} from '@core/services/project';
 import {UserService} from '@core/services/user';
 import {Project} from '@shared/entity/project';
@@ -25,6 +25,7 @@ import {switchMap, takeUntil, tap} from 'rxjs/operators';
   selector: 'km-project-selector',
   templateUrl: './template.html',
   styleUrls: ['./style.scss'],
+  encapsulation: ViewEncapsulation.None,
   standalone: false,
 })
 export class ProjectSelectorComponent implements OnInit, OnDestroy {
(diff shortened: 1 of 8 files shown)


buraksekili:
@ahmadhamzh @KhizerRehan please address the cherrypick PRs. if cherrypick is not needed, feel free to close them

KhizerRehan:
@buraksekili This PR is not supposed to be cherrypick bcz this was recently introduced due to upgrade to [Angular V21 ](https://github.com/kubermatic/dashboard/pull/8052) 

This bug doesn't exist in past 3 releases. So, we don't need this PR to be backported.

thx

buraksekili:
superb, thanks @KhizerRehan

PR #8229, opened 2026-08-03
Add kubevirt node defaults values

### LINKED ISSUE #16139: Make KubeVirt node defaults configurable per datacenter

As a KKP admin
I would like to configure default CPU, memory, and disk size for KubeVirt worker nodes per datacenter
in order to set values that match my infrastructure and give my users sensible defaults when they create machine deployments.

Today the default node size for KubeVirt machine deployments is hardcoded in the dashboard. An admin has no way to control it.

Some operators run KubeVirt with the default instance types disabled (`disableDefaultInstanceTypes: true`), which forces their users onto the "Custom Configuration" path in the node form. For these environments, the hardcoded defaults are the only sizing the user sees, and there is no way for the admin to set values that match their infrastructure.

We want cluster admins to be able to define the default node size (disk, CPU, memory) for a given KubeVirt datacenter, in the Seed configuration. The dashboard should use these values when they are defined, and fall back to its own hardcoded defaults otherwise.

### Solution details

- A new field exists in the KubeVirt datacenter spec that lets an admin set default CPU, memory, and primary disk size.
- The dashboard reads these values from the datacenter spec and uses them to pre-fill the KubeVirt node form on the "Custom Configuration" path when they are present.
- When a value is not set by the admin, the dashboard falls back to the hardcoded defaults.
- The behavior of editing an existing machine deployment is unchanged. Saved values are kept; defaults only apply on creation.
- Selecting an instance type still overrides CPU and memory, as it does today.

This work spans two repositories.

Kubermatic (KKP):

- Add a block to `DatacenterSpecKubevirt` in `sdk/apis/kubermatic/v1/datacenter.go` to hold the default CPU, memory, and primary disk size.
- Regenerate the deepcopy code, the Seeds CRD, and the documentation.
- Optionally, apply the same defaults server-side in `CompleteKubevirtProviderSpec` (`pkg/machine/provider/kubevirt.go`), so that specs created through GitOps, cluster templates, or the API also receive the defaults instead of only the UI path.

Dashboard:

- Bump the KKP SDK dependency in `modules/api/go.mod`. The API datacenter type embeds the SDK `DatacenterSpecKubevirt` struct directly, so new fields on that struct are serialized into the `/dc` endpoint response automatically.
- Verify the datacenter handler returns the new fields to the frontend. The handler copies the whole KubeVirt spec struct by reference rather than field by field, so new fields are not dropped, but it is worth confirming.
- Add the new fields to the `KubeVirtDatacenterSpec` TypeScript entity, which is currently empty. The frontend entities under `modules/web/src/app/shared/entity/` are hand-maintained mirrors of the Go structs, not generated from them, so this step is manual and separate from the SDK bump.
- In the KubeVirt node component, read the datacenter defaults and use them where the hardcoded constants are used today.

ahmadhamzh:
Add support for KubeVirt DC node defaults (CPU, memory, primary disk size) in the node-data form, falling back to hardcoded defaults when a datacenter has none set.

Also fixes a bug where clearing the primary disk size field sent an invalid "null" value to the quota-calculation API.

Also updates e2e CI fixtures (Dex httpRoute.domain, envoy-gateway NodePort) needed by the newer KKP dependency this PR pulls in.

Fixes https://github.com/kubermatic/kubermatic/issues/16139

**What type of PR is this?**
/kind bug
/kind feature

```release-note
KubeVirt worker nodes now use the default CPU, memory, and primary disk size configured on the seed's datacenter (`seed.spec.kubevirt.nodeDefaults`) when creating or editing a node, falling back to the previous built-in defaults if none are set.
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
+++ b/modules/web/src/app/node-data/basic/provider/kubevirt/component.ts
@@ -27,10 +27,12 @@ import {MatDialog} from '@angular/material/dialog';
 import {ClusterSpecService} from '@app/core/services/cluster-spec';
 import {DynamicModule} from '@app/dynamic/module-registry';
 import {InstanceDetailsDialogComponent} from '@app/node-data/basic/provider/kubevirt/instance-details/component';
+import {DatacenterService} from '@core/services/datacenter';
 import {GlobalModule} from '@core/services/global/module';
 import {NodeDataService} from '@core/services/node-data/service';
 import {QuotaCalculationService} from '@dynamic/enterprise/quotas/services/quota-calculation';
 import {ComboboxControls, FilteredComboboxComponent} from '@shared/components/combobox/component';
+import {KubeVirtNodeDefaults} from '@shared/entity/datacenter';
 import {KubeVirtNodeAffinityPreset, KubeVirtNodeSpec, NodeCloudSpec, NodeSpec} from '@shared/entity/node';
 import {
   KubeVirtAffinityPreset,
@@ -56,8 +58,8 @@ import {NodeData} from '@shared/model/NodeSpecChange';
 import {BaseFormValidator} from '@shared/validators/base-form.validator';
 import {KUBERNETES_RESOURCE_NAME_PATTERN} from '@shared/validators/others';
 import _ from 'lodash';
-import {merge, Observable} from 'rxjs';
-import {filter, map, take, takeUntil} from 'rxjs/operators';
+import {merge, Observable, of} from 'rxjs';
+import {filter, map, switchMap, take, takeUntil} from 'rxjs/operators';
 
 enum Controls {
   InstanceType = 'instancetype',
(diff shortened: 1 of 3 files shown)


buraksekili (approved):
LGTM from my side - i'll leave the final "lgtm" to @KhizerRehan 

/approve

KhizerRehan:
/hold 

I think dex related and kind cluster changes should be in sep PR IMO and this PR should be only catering Kubevirt changes.

Thank you!

ahmadhamzh:
> I think dex related and kind cluster changes should be in sep PR IMO and this PR should be only catering Kubevirt changes.

@KhizerRehan https://github.com/kubermatic/dashboard/pull/8237

KhizerRehan:
Hi, @ahmadhamzh once this PR https://github.com/kubermatic/dashboard/pull/8237  is merged you can rebase this PR from main.

KhizerRehan:
HI, @ahmadhamzh I tested this PR and looks good to me.

Just FYI: I saw default values set in Seed / Kubevirt Datacenter

Did you added for testing purpose and needs to be removed for dev environment?

CRD: Seed: Shared

- nodeDefaults (object)

image

/approve
/lgtm

PR #8253, opened 2026-08-11
Add KubeVirt preset subnets field

ahmadhamzh:
Adds a subnets field to the KubeVirt preset (admin UI + API), and resolves Machine Deployment subnets from the preset before falling back to the seed or live credentials.

Fixes #https://github.com/kubermatic/kubermatic/issues/16217

**What type of PR is this?**
/kind feature

```release-note
KubeVirt presets can now define subnets, used first when picking Machine Deployment subnets.
```

**Documentation**:
```documentation
TBD
```

**Test issue**:
```test-issue
TBD
```

### DIFF
--- a/modules/api/pkg/handler/common/provider/kubevirt.go
+++ b/modules/api/pkg/handler/common/provider/kubevirt.go
@@ -43,6 +43,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/utils/ptr"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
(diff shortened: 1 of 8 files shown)


ahmadhamzh:
/retitle  WIP Add KubeVirt preset subnets field

modules/api/pkg/handler/common/provider/kubevirt.go
```diff
@@ -181,6 +182,19 @@ func KubeVirtSubnetsWithClusterCredentialsEndpoint(ctx context.Context, userInfo
 		return "", err
 	}
 
+	if presetName := cluster.Annotations[kubermaticv1.PresetNameAnnotation]; presetName != "" {
```
buraksekili, line 190: When the DC sets `MatchSubnetAndStorageLocation`, the seed path (below, around ~203) filters subnets by the storage class regions and zones, because infra storage is often zonal and a VM on a subnet in one zone cannot attach a volume in another.

The new preset path returns the admin's subnet and returns before that filter runs, so on a flag-enabled datacenter the UI can offer subnets that KKP core rejects at machine deployment creation. I guess, we can run them through the same matching/filtering mechanism. wdyt?

buraksekili (approved):
/lgtm
/approve

PR #8284, opened 2026-08-27
feat: Added support for specifying the values for Azure Security Profile

### LINKED ISSUE #16196: Supporting latest Azure Gen2 model images via UI (at least via json)

As a KKP Admin  
I would like to use new Azure gen2 images which need `securityType: TrustedLaunch`
to spawn new VMs.

Support for this configuration was added to Machine-controller via https://github.com/kubermatic/machine-controller/pull/2020

Currently, we do not support providing these parameters via KKP API.. so KKP user-clusters cannot make use of these VMs.

### Solution details

There can be two parts for solutionizing this.
1. Change to Kubermatic-API to support for providing these values via JSON content for cluster spec.
2. We can, optionally, also allow user to configure these values in dashboard.

### Use cases

Support latest features provided in machine-controller

dharapvj:
[This PR](https://github.com/kubermatic/machine-controller/pull/2020) on machine-controller added support for specifying TrustedLaunch security profile to Azure VMs.

This PR allows creation of Azure VMs with TrustedLaunch security profile for user-cluster worker nodes via KKP dashboard API. This paves way for API based GitOps for usercluster VMs to use TrustedLaunch security based Azure images.

Fixes https://github.com/kubermatic/kubermatic/issues/16196

**What type of PR is this?**

/kind feature
/kind chore

```release-note
Azure user-clusters - support for VMs with securityProfile TrustedLaunch added.
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
TBD
```


modules/api/pkg/api/v1/types.go
```diff
@@ -1781,6 +1781,23 @@ type AzureNodeSpec struct {
 	AssignAvailabilitySet bool `json:"assignAvailabilitySet"`
 	// EnableAcceleratedNetworking is used to check if an accelerating networking should be used for azure vms.
 	EnableAcceleratedNetworking *bool `json:"enableAcceleratedNetworking,omitempty"`
+	// SecurityProfile specifies the security settings, e.g. TrustedLaunch, for the VM.
+	// required: false
+	SecurityProfile *AzureSecurityProfile `json:"securityProfile,omitempty"`
```
KhizerRehan, line 1786: I see API change but it is not being touched from UI side 

Path: `modules/web/src/app/shared/entity/node.ts` can you confirm?

KhizerRehan: IMHO, incase we add above property do we need to uodate from UI wizard (Node Step) POV?

modules/api/pkg/validation/node.go
```diff
@@ -30,5 +31,31 @@ func ValidateCreateNodeSpec(c *kubermaticv1.Cluster, spec *apiv1.NodeSpec, dc *k
...
+	return nil
+}
+
+func validateAzureSecurityProfile(profile *apiv1.AzureSecurityProfile) error {
+	if profile == nil {
+		return nil
+	}
+
+	switch profile.SecurityType {
+	case "TrustedLaunch":
```
KhizerRehan, line 49: Is `ConfidentialVM` intentionally excluded from the allowed securityType values?

modules/api/pkg/validation/node.go
```diff
@@ -30,5 +31,31 @@ func ValidateCreateNodeSpec(c *kubermaticv1.Cluster, spec *apiv1.NodeSpec, dc *k
 		}
 	}
 
+	if c.Spec.Cloud.Azure != nil && spec.Cloud.Azure != nil {
+		if err := validateAzureSecurityProfile(spec.Cloud.Azure.SecurityProfile); err != nil {
```
KhizerRehan, line 35: ValidateCreateNodeSpec is only called in the OpenStack case?  is the Azure security-profile validation meant to run somewhere, or is it intentionally unreachable for now?

As per understanding
ValidateCreateNodeSpec 
   -> calls -> validateAzureSecurityProfile

- [resources/machine/machinedeployment.go#L236-L245  (Openstack Only)]
(https://github.com/kubermatic/dashboard/blob/ad124e93a90d30ae201474a0996f8b27776013de/modules/api/pkg/resources/machine/machinedeployment.go#L236-L245)
- resources/machine/machinedeployment.go#L218-L223  => no validation

PTAL: I don't have much context from other cloud providers POV

KhizerRehan (commented):
PTAL: Minor Feedback

thank you!

modules/api/pkg/validation/node.go
KhizerRehan: /cc: @buraksekili

modules/api/pkg/api/v1/types.go
dharapvj: As discussed, this PR is only addressing the 1st part of the solution which is adding the support for "securityProfile" in the Rest API. For 2nd part - which is UI - we should ideally create a separate PR.

From the original ticket
> There can be two parts for solutionizing this.
>    Change to Kubermatic-API to support for providing these values via JSON content for cluster spec.
>    We can, optionally, also allow user to configure these values in dashboard.

modules/api/pkg/validation/node.go
dharapvj: ConfidentialVM is not supported in machine-controller itself.. so for now, I guess, we can ignore it. We can add support for it later on if the customer requirement comes for this.

modules/api/pkg/validation/node.go
dharapvj: My opinion on this: 
All the validations of node options is done by machine controller webhook. I am not sure why only for openstack we have some extra validation in dashboard code. In general, all the other cloud should continue to happen in machine controller.. so we don't need to worry about this review comment from copilot.

buraksekili: I believe we need to wire the validation here. As Khizer mentioned, this validation does not run for Azure. Let's run it because even though the MC webhook needs to be the source of truth for validation (I agree with VJ), the initial cluster creation, dashboard creates "Cluster" object, not "Machine{Deployment}"; thus, machine-controller webhook won't validate the user's submission here, until seed-controller picks this Cluster object, reconciles and creates the initial MachineDeployment object. So, getting a webhook error from the MC webhook for the cluster creation is async; thus, it can't directly prevents creation of the clusters with malformed provider inputs.

That's why i am thinking of wiring this validation to protect initial Cluster creation for Azure provider, and further operations (such as updating MDs) can rely on MC webhooks since Dashboard directly modifies the MachineDeployments

buraksekili:
UI side can be implemented as a follow-up, TBD: @KhizerRehan

KhizerRehan:
/approve
/lgtm

Respective tickets will be created and UI part will be handled separately

PR #8295, opened 2026-09-07
Accelerator quota enhancement

ahmadhamzh:
some UX enhancements for the accelerator quota.

**Edit project quota dialog for a default quota:**

before:
image

after:
image

**Add project quota dialog:**

before:
image

after:
image

**project quota list view:**

before:
image

after:
image

Fixes #

**What type of PR is this?**
/kind design

```release-note
Accelerator quota can no longer be enabled on a default project quota.
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
NONE
```

### DIFF
--- a/modules/web/src/app/dynamic/enterprise/quotas/project-quota-dialog/component.spec.ts
+++ b/modules/web/src/app/dynamic/enterprise/quotas/project-quota-dialog/component.spec.ts
@@ -105,10 +105,12 @@ describe('ProjectQuotaDialogComponent accelerator accounting', () => {
   function quotaWith(
     accountingEnabled: boolean,
     phase?: AcceleratorAccountingPhase,
-    accelerators?: AcceleratorQuota[]
+    accelerators?: AcceleratorQuota[],
+    isDefault = false
   ): QuotaDetails {
     const quota = GetQuotasMock()[0];
     quota.acceleratorAccountingEnabled = accountingEnabled;
+    quota.isDefault = isDefault;
     if (accelerators) {
       quota.quota.accelerators = accelerators;
     }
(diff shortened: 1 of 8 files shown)


buraksekili (approved):
/lgtm

I'll leave web part to @KhizerRehan 's expertise

KhizerRehan:
Suggestion/Though

Since I was testing and I was thinking Do you think it can be a Good Enhancement to show which project has GPU Enabled. Like Someon who lands on Project list it doesn't show GPU is enabled only in this Project unless hover widget and see "Acceleraotrs"

image

Just thinking something like this we can add 🟢 and Hover tooltip as in indication? 

image

thoughts?

modules/web/src/app/dynamic/enterprise/quotas/project-quota-dialog/component.ts
```diff
@@ -129,9 +129,31 @@ export class ProjectQuotaDialogComponent implements OnInit, OnDestroy {
...
+
+  get acceleratorAccountingTooltip(): string {
+    if (!this.editQuota) {
+      return 'Accelerator quota can be enabled after the project quota has been created.';
+    }
+    if (this.isAcceleratorAccountingEnabled) {
+      return 'Accelerator quota is enabled and cannot be turned off again.';
+    }
+    if (this.isDefaultQuota) {
+      return 'Accelerator quota cannot be enabled on a default project quota.';
```
KhizerRehan: Comment [here](https://github.com/kubermatic/dashboard/pull/8295/changes#diff-fe89dafea0e4279963f2db78deda0a7070e6e83041019c8d5033707edc48ad50R576-R577)

can we add something like "setting has to be turned into a project-specific one first." which shows like how we can set then....bcz i was figuring if there was any setting to override default project quota but simple override and save can then allow to set it.

Eample

`
Accelerator quota cannot be enabled on a default project quota. Save a change to the quota values first to make it project-specific.
`

Just a suggestion let me wdyt?
thx

modules/web/src/app/dynamic/enterprise/quotas/project-quota-dialog/component.spec.ts
```diff
@@ -153,6 +186,8 @@ describe('ProjectQuotaDialogComponent accelerator accounting', () => {
     expect(component.canRemoveAccelerators).toBe(true);
   });
 
+  // canEditAccelerators also gates the accelerator naming hint below the limit rows, which is noise
+  // while the rows are disabled.
```
KhizerRehan: Do we need this sort of comment?

modules/web/src/app/dynamic/enterprise/quotas/project-quota-dialog/template.html
```diff
@@ -97,31 +97,28 @@
     <div fxLayout="column"
          fxLayoutGap="10px">
       <mat-card-header class="km-no-padding">
-        <mat-card-title>
-          Accelerators
-          <i class="km-icon-info km-pointer"
-             matTooltip="Accelerator must match the exact resource name reported by the provider (e.g. nvidia.com/GA100_A30). A limit of 0 explicitly denies that resource; a resource with no row is unconstrained."></i>
-        </mat-card-title>
+        <mat-card-title>Accelerators</mat-card-title>
```
KhizerRehan, line 100: IMHO,

Dev: Showed tootlip on "Accelerators" ℹ️  which adds value about 
.eg " A limit of 0 explicitly denies that resource ....0 with no constrained"

Image

Current PR:

Image


I think we can keep that tooltip or modify tooltip but can be part of "Accelerators" that adds some info NO?

ahmadhamzh: i remove it cause i thought that this tooltip doesn't belong to the title, so i add it for the accelerator limits rows 
image

ahmadhamzh:
> Just thinking something like this we can add 🟢 and Hover tooltip as in indication?

im not sure if just adding :green_circle: will indicate that this quota have an accelerator ? maybe we can check that more in the next enhancement, i think the feature will have some changes in the near future.

modules/web/src/app/dynamic/enterprise/quotas/project-quota-dialog/template.html
KhizerRehan: oh! I see

But I also before it is also added a valuable info too as far I understand `0` accelerator means NO constraint and I don't how setting will be add constraint but something i get to know from tooltip

and IMO, can we add Info icon to make it prominnet

image

wdyt?

PR #8311, opened 2026-09-17
Fix KubeVirt instance type memory units

### LINKED ISSUE #8276: [UX] Updating UoM for memory and storage intelligently
In the cluster creation wizard, for kubevirt provider, we list out VM sizes available. In this table the memory available is always expressed in MB and it is rather cumbersome to understand what GBs they were.

In a the cluster detail page, same VM visible as a node shows values correctly in GB. We should use same UI logic in the available template sizes table.

ahmadhamzh:
The API rescaled instance type memory to decimal SI, so the wizard listed `8590M` for an `8Gi` type and quota counted it as `8590Mi` — 4.86% more than the machine actually uses.
It is now returned as declared, displayed as 8 GB, and counted exactly.

Fixes #8276 

**What type of PR is this?**
/kind bug

```release-note
KubeVirt instance type memory is now returned and displayed in binary units,
```

**Documentation**:

```documentation
NONE
```

**Test issue**:

```test-issue
TBD
```


buraksekili (approved):
LGTM. I'll leave UI changes to @KhizerRehan

KhizerRehan:
/approve
/lgtm

@ahmadhamzh can you take care if this PR needs to be back ported. 
thx

